PR #3, opened 2017-06-28
Refactor connection to no longer be a global variable

tom-meyer:
We'd like comments from the team on this one. Things to discuss:

- Connection is constructed in main()
- Connection object is passed around all over the place
- We are slightly vulnerable with our DB teardown code right now... is there a better way?

### DIFF
--- a/backup/backup.go
+++ b/backup/backup.go
@@ -8,8 +8,7 @@ import (
 )
 
 var (
-	connection *utils.DBConn
-	logger     *utils.Logger
+	logger *utils.Logger
 )
 
 var ( // Command-line flags
@@ -39,7 +38,7 @@ func DoValidation() {
 }
 
 // This function handles setup that must be done after parsing flags.
-func DoSetup() {
+func DoSetup() *utils.DBConn {
 	if *quiet {
 		logger.SetVerbosity(utils.LOGERROR)
 	} else if *debug {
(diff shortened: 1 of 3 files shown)


jmcatamney:
Regarding `connection` being passed around a lot, it's done that way because at some point we plan to have multiple connections for parallelism and we wrote most functions with that in mind.  I'd prefer to have less passing around, maybe with some sort of global connection pool or the like, but either way making the connection non-global exacerbates the issue.

Regarding vulnerability in the teardown code, do you mean before or after the refactor, and what vulnerability do you see?

Personally, I'd like to see all teardown performed in the DoTeardown function to ensure that everything happens in one place, so having to split off the connection.Close() into a separate defer block is one downside of the refactor for me.

chrishajas:
I don't have very strong opinions on this either way, but I favor fewer globals. If we add parallel backups (and therefore multiple connections), we'll have to use a different organization for our connections anyway.

professor:
In general, less globals are nice. 

If a pair decided to do this refactoring, I would tend to want to trust that pair. 

Maybe DI could be another technique to pass a Connection to the needed structs. (I'm not saying to do this.)

chrishajas:
We've decided to hold off on refactoring the global connection until we decide how we want to handle parallel connections.

PR #8, opened 2017-11-17
Added support for multi-level external partitions

khuddlefish:
* Supports arbitrary depth as long as the name of the table doesn't
exceed the maximum for postgres

### DIFF
--- a/backup/wrappers.go
+++ b/backup/wrappers.go
@@ -90,9 +90,9 @@ func RetrieveAndProcessTables() ([]Relation, []Relation, map[uint32]TableDefinit
 	tableDefs := ConstructDefinitionsForTables(connection, tables)
 	extPartInfo := GetExternalPartitionInfo(connection)
 	for _, partInfo := range extPartInfo {
-		tableDef := tableDefs[partInfo.ParentOid]
+		tableDef := tableDefs[partInfo.ParentRelationOid]
 		tableDef.IsExternal = true
-		tableDefs[partInfo.ParentOid] = tableDef
+		tableDefs[partInfo.ParentRelationOid] = tableDef
 	}
 	metadataTables, dataTables := SplitTablesByPartitionType(tables, tableDefs, userPassedIncludeTables)
 	objectCounts["Tables"] = len(metadataTables)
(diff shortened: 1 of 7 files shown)


integration/predata_externals_queries_test.go
```diff
@@ -113,56 +113,133 @@ SEGMENT REJECT LIMIT 10 PERCENT
...
+			testutils.AssertQueryRuns(connection, `
+CREATE TABLE part_tbl (a int,b date,c text,d int)
+DISTRIBUTED BY (a)
+PARTITION BY RANGE (b)
+SUBPARTITION BY LIST (c)
+SUBPARTITION TEMPLATE
+(SUBPARTITION usa values ('usa'),
+SUBPARTITION apj values ('apj'),
+SUBPARTITION eur values ('eur'))
+(PARTITION Jan16 START (date '2016-01-01') INCLUSIVE ,
```
chrishajas: We probably can cut some of the number of partitions in this test to just a couple.

end_to_end/end_to_end_suite_test.go
```diff
@@ -193,5 +193,31 @@ var _ = Describe("backup end to end integration tests", func() {
...
+        START (1) END (13) EVERY (1),
+        DEFAULT SUBPARTITION other_months )
+           SUBPARTITION BY LIST (region)
+             SUBPARTITION TEMPLATE (
+               SUBPARTITION usa VALUES ('usa'),
+               SUBPARTITION europe VALUES ('europe'),
+               SUBPARTITION asia VALUES ('asia'),
+               DEFAULT SUBPARTITION other_regions )
+( START (2002) END (2012) EVERY (1),
+  DEFAULT PARTITION outlying_years );
```
chrishajas: If we have fewer partitions (both from the start and end year and the months), does the table creation time go down significantly? I'd prefer to have this in the create test rather than end to end, but if it still takes a long time I'm ok with keeping it here.

backup/wrappers.go
```diff
@@ -474,7 +474,7 @@ func BackupData(tables []Relation, tableDefs map[uint32]TableDefinition) {
 		if numExtParts > 1 {
 			s = "s"
 		}
-		logger.Warn("Skipped data backup of %d partition table%s.", numExtParts, s)
+		logger.Warn("Skipped data backup of %d partition table%s containing external partitions.", numExtParts, s)
```
jmcatamney: We might want to make this "containing one or more external partitions" but that's a nitpick.

khuddlefish: I can change that. I was wrestling a bit with syntax that made sense for singular or plural, but that would work fine.

khuddlefish:
Merged

PR #14, opened 2017-12-08
Implement connection pooling in DBConn

jmcatamney:
While the sqlx.DB struct (and indirectly the sql.DB struct) maintains its own
connection pool, there is no guarantee of session-level consistency between
queries, and we require that level of control for e.g. setting certain GUCs on
each connection before restoring data.  Also, while there is a sql.Conn struct
that represents a single session, there is no sqlx.Conn equivalent we could use.

Thus, this commit adds a connection pool to DBConn in the form of an array of
sqlx.DBs, all set to have exactly one database connection each, such that callers
can create NumConns goroutines and assign each an index from 0 to NumConns to
guarantee that each goroutine gets its own connection and that the connection
will exhibit single-session behavior.

The Exec, Select, and Get functions are set up to default to the first connection
(index 0) so the DBConn will still exhibit session-like behavior if no connection
is specified, and other functions that want to execute in serial should pass in
a 0 wherever a connection number is needed.


restore/parallel_test.go
```diff
@@ -52,13 +52,13 @@ var _ = Describe("restore/parallel tests", func() {
 				mock.ExpectExec(commentStr).WillReturnResult(sqlmock.NewResult(0, 0))
 				mock.ExpectExec(createStr).WillReturnResult(sqlmock.NewResult(1, 0))
 				mock.ExpectExec(gucStr).WillReturnResult(sqlmock.NewResult(0, 1))
-				restore.ExecuteStatements(statements, "", false, shouldExecute)
+				restore.ExecuteStatements(statements, "", false, shouldExecute, false)
```
chrishajas: Should these parallel tests have the `executeInParallel` flag set to true?

khuddlefish: related to that, are we doing anything to validate that these are actually running in parallel?

jmcatamney: Yes, they should be true; copy-paste error.

No, we're not specifically testing that they _are_ running in parallel, only that they _can_ thanks to the  MatchExpectationsInOrder(false).  I'll add some sleeps to guarantee non-serial ordering.

PR #15, opened 2017-12-08
Speed up end_to_end tests

chrishajas:
Instead of dropping and recreating the database on every scenario, drop and
recreate the schema. This reduces the total test time from ~70s to ~40s.
Each individual scenario now takes ~2-3s.

### DIFF
--- a/end_to_end/end_to_end_suite_test.go
+++ b/end_to_end/end_to_end_suite_test.go
@@ -65,6 +65,10 @@ var _ = Describe("backup end to end integration tests", func() {
 		if err != nil {
 			Fail(fmt.Sprintf("%v", err))
 		}
+		err = exec.Command("createdb", "restoredb").Run()
+		if err != nil {
+			Fail(fmt.Sprintf("%v", err))
+		}
 		backupConn = utils.NewDBConn("testdb")
 		backupConn.Connect()
 		utils.ExecuteSQLFile(backupConn, "test_tables.sql")
@@ -83,13 +87,12 @@ var _ = Describe("backup end to end integration tests", func() {
 			if restoreConn != nil {
 				restoreConn.Close()
 			}
-			exec.Command("dropdb", "restoredb").Run()
-			err := exec.Command("createdb", "restoredb").Run()
+			restoreConn = utils.NewDBConn("restoredb")
+			restoreConn.Connect()
+			err := exec.Command("psql", "-d", "restoredb", "-c", "DROP SCHEMA IF EXISTS schema2 CASCADE; DROP SCHEMA public CASCADE; CREATE SCHEMA public;").Run()
 			if err != nil {
 				Fail(fmt.Sprintf("%v", err))
 			}
-			restoreConn = utils.NewDBConn("restoredb")
-			restoreConn.Connect()
 		})
 		It("runs gpbackup and gprestore without redirecting restore to another db", func() {
 			timestamp := gpbackup(gpbackupPath)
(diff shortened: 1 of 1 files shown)


end_to_end/end_to_end_suite_test.go
```diff
@@ -83,13 +87,12 @@ var _ = Describe("backup end to end integration tests", func() {
 			if restoreConn != nil {
 				restoreConn.Close()
 			}
-			exec.Command("dropdb", "restoredb").Run()
-			err := exec.Command("createdb", "restoredb").Run()
+			restoreConn = utils.NewDBConn("restoredb")
```
khuddlefish: If we don't drop and restore the db every time, the DBConn creation should probably go in BeforeSuite

end_to_end/end_to_end_suite_test.go
```diff
@@ -197,8 +200,7 @@ var _ = Describe("backup end to end integration tests", func() {
 		})
 		It("runs gpbackup and gprestore on database with all objects", func() {
 			backupConn.Close()
```
khuddlefish: Don't need to close the connection if we're not dropping the database

end_to_end/end_to_end_suite_test.go
```diff
@@ -197,8 +200,7 @@ var _ = Describe("backup end to end integration tests", func() {
...
-			exec.Command("dropdb", "testdb").Run()
-			err := exec.Command("createdb", "testdb").Run()
+			err := exec.Command("psql", "-d", "testdb", "-c", "DROP SCHEMA IF EXISTS schema2 CASCADE; DROP SCHEMA public CASCADE; CREATE SCHEMA public;").Run()
```
khuddlefish: We could probably use testutils.AssertQueryRuns for these statements. Would be a bit more consistent with the rest of our tests.

end_to_end/end_to_end_suite_test.go
khuddlefish: Can also take out creating the new backupConn at the end of this test

end_to_end/end_to_end_suite_test.go
```diff
@@ -65,6 +65,10 @@ var _ = Describe("backup end to end integration tests", func() {
 		if err != nil {
 			Fail(fmt.Sprintf("%v", err))
 		}
+		err = exec.Command("createdb", "restoredb").Run()
+		if err != nil {
+			Fail(fmt.Sprintf("%v", err))
+		}
 		backupConn = utils.NewDBConn("testdb")
 		backupConn.Connect()
 		utils.ExecuteSQLFile(backupConn, "test_tables.sql")
```
khuddlefish: We could also change this to be like 
``` 
connStr := []string{
"-d", "testdb",
"-f", "all_objects.sql",
"-q",
}
exec.Command("psql", connStr...)
```
below and then we can remove the `ExecuteSQLFile` function.

chrishajas: The ExecuteSQLFile function also sets `ON_ERROR_STOP=1`, which prevents the swallowing of errors. I think instead we should use this function when executing sql files so we don't accidentally commit bad syntax.

end_to_end/end_to_end_suite_test.go
chrishajas: Fixed, thanks!

end_to_end/end_to_end_suite_test.go
chrishajas: Fixed, thanks!

end_to_end/end_to_end_suite_test.go
chrishajas: Fixed, thanks!

khuddlefish:
Pushed to master!

PR #17, opened 2017-12-08
Move timestamp and lock file generation earlier

chrishajas:
Previously, we were generating the timestamp and the associated lock
file after we had validated the filter lists. Since this validation
requires querying the database, it can take long enough that the
timestamp generated could be over a second later than when the 
user started the backup. Now, we generate the timestamp and 
lock file before this validation closer to the beginning of the backup.

### DIFF
--- a/backup/backup.go
+++ b/backup/backup.go
@@ -56,6 +56,8 @@ func DoFlagValidation() {
 // This function handles setup that must be done after parsing flags.
 func DoSetup() {
 	SetLoggerVerbosity()
+	timestamp := utils.CurrentTimestamp()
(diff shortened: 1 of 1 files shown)


chrishajas:
@khuddlefish Looking at this closer, the issue is with a timestamp being generated at a later point than was started due to the time it took for the validation query (even if the user added a sleep for a second), not that two backups could be started simultaneously as I previously thought. Since flag validation takes basically no time, generating the timestamp and lock file after basic flag validation shouldn't be an issue. It's basically whether we want to tell the user that their syntax is wrong or that another backup is running first.

PR #19, opened 2017-12-14
Treat backups/restores with no tables as metadata backups/restores

chrishajas:
Previously, backups and restores would fail if no tables were present
due to the logic around segment TOC files. Instead, don't attempt to
backup/restore data if no tables exist.

### DIFF
--- a/backup/data.go
+++ b/backup/data.go
@@ -112,3 +112,15 @@ func printDataBackupWarnings(numExtTables int) {
 		logger.Warn("See %s for a complete list of skipped tables.", logger.GetLogFilePath())
 	}
 }
+
+func CheckTablesContainData(tables []Relation, tableDefs map[uint32]TableDefinition) {
+	if !backupReport.MetadataOnly {
+		for _, table := range tables {
+			if !tableDefs[table.Oid].IsExternal {
+				return
+			}
+		}
+		logger.Warn("No tables in backup set contain data. Performing metadata-only backup instead.")
+		backupReport.MetadataOnly = true
+	}
+}
--- a/backup/data_test.go
+++ b/backup/data_test.go
@@ -80,4 +80,32 @@ var _ = Describe("backup/data tests", func() {
 			backup.CopyTableOut(connection, testTable, filename)
 		})
 	})
+	Describe("CheckDBContainsData", func() {
+		config := utils.BackupConfig{}
+		testTable := []backup.Relation{backup.BasicRelation("public", "testtable")}
+
+		BeforeEach(func() {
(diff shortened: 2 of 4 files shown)


jmcatamney:
In this scenario, I think we should treat the backup as a metadata-only backup in all ways (set the type to Metadata Only in the config and so forth) with a Warning/Info message informing the user of that, since a "full" backup with no tables and no data doesn't make much sense and there's no sense looking for nonexistent data during the restore.

Thoughts?

chrishajas:
Updated PR to automatically change backup to metadata backup.

backup/data.go
```diff
@@ -112,3 +112,15 @@ func printDataBackupWarnings(numExtTables int) {
 		logger.Warn("See %s for a complete list of skipped tables.", logger.GetLogFilePath())
 	}
 }
+
+func CheckDBContainsData(tables []Relation, tableDefs map[uint32]TableDefinition) {
```
jmcatamney: I'd change this to "CheckTablesContainData", since we're specifically checking certain tables and "CheckDB" isn't accurate in the table filter case.

backup/data.go
```diff
@@ -112,3 +112,15 @@ func printDataBackupWarnings(numExtTables int) {
...
+	if !backupReport.MetadataOnly {
+		for _, table := range tables {
+			if !tableDefs[table.Oid].IsExternal {
+				return
+			}
+		}
+		logger.Warn("No data found in backup. Performing metadata-only backup instead.")
```
jmcatamney: "No data found in backup" might scare people.  How about "No tables in backup set contain data" instead?

backup/data_test.go
```diff
@@ -89,22 +89,22 @@ var _ = Describe("backup/data tests", func() {
...
 			Expect(backup.GetReport().BackupConfig.MetadataOnly).To(BeTrue())
 		})
 		It("does not change backup type if metadata-only backup", func() {
 			config.MetadataOnly = true
 			backup.SetReport(&utils.Report{BackupConfig: config})
-			backup.CheckDBContainsData([]backup.Relation{}, map[uint32]backup.TableDefinition{})
+			backup.CheckTablesContainData([]backup.Relation{}, map[uint32]backup.TableDefinition{})
 			Expect(backup.GetReport().BackupConfig.MetadataOnly).To(BeTrue())
 		})
 		It("does not change backup type tables present in database", func() {
```
khuddlefish: Not to be picky but I think you're missing an "if" :)

chrishajas: Thanks, have to rebase anyway due to conflicts.

PR #21, opened 2017-12-16
Combine global, predata and postdata into a single metadata file

khuddlefish:
Now that we have the toc and the ability to selectively restore any
objects, we can combine these sections into a single file. Behind the
scenes, we still maintain separate lists of the three for easier
isolation of the various sections on the restore side.

### DIFF
--- a/utils/cluster.go
+++ b/utils/cluster.go
@@ -404,9 +404,7 @@ func (cluster *Cluster) GetTableBackupFilePathForCopyCommand(tableOid uint32, si
 
 var metadataFilenameMap = map[string]string{
 	"config":            "config.yaml",
-	"global":            "global.sql",
-	"predata":           "predata.sql",
-	"postdata":          "postdata.sql",
+	"metadata":          "metadata.sql",
 	"statistics":        "statistics.sql",
 	"table of contents": "toc.yaml",
 	"report":            "report",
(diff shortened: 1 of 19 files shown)


khuddlefish:
I changed the logging a bit already so it will only give the name of the metadata file once and then just state when it is writing/restoring the other sections without giving the file names.

Do we also want to put comments in the metadata file to separate the sections for user readability?

utils/cluster.go
```diff
@@ -456,9 +446,7 @@ func (cluster *Cluster) VerifyMetadataFilePaths(dataOnly bool, withStats bool, t
 	filetypes := []string{"config", "table of contents"}
 	if !dataOnly {
 		if !tableFiltered {
```
jmcatamney: This check for tableFiltered should be taken out; we always want to check for the metadata file if it's not a data-only backup.

chrishajas:
Do we even want/need to list the location of the metadata file? I can see putting it in verbose logs, but in most cases that shouldn't be a concern to users.

As for comments in the metadata file, I think the metadata files will typically be large enough where users are grepping around for certain statements instead of reading the file, so I don't have a strong opinion either way.

jmcatamney:
Having the metadata filename in the INFO output is useful for people who want to tail the file or something without having to find the name in gpAdmingLogs, but not so necessary that we have to keep it there and not move it to verbose output.  I don't have strong feelings either way.

I agree with Chris, users should only be using the metadata file without gprestore to either run the whole thing at once or to look for and grab specific statements, and in both cases boundary comments don't really add any value.

khuddlefish:
Metadata file path moved to verbose output, tableFilter check removed in cluster.go and not adding comments to separate sections in the metadata file. Anything else?

PR #22, opened 2017-12-18
Remove logging and progress bars to stdout when -quiet flag used

chrishajas:
Also, don't print progress bars when -verbose flag used or to log file

### DIFF
--- a/restore/restore.go
+++ b/restore/restore.go
@@ -59,12 +59,12 @@ func DoValidation() {
 	if !utils.IsValidTimestamp(*timestamp) {
 		logger.Fatal(errors.Errorf("Timestamp %s is invalid.  Timestamps must be in the format YYYYMMDDHHMMSS.", *timestamp), "")
 	}
-	logger.Info("Restore Key = %s", *timestamp)
 }
 
 // This function handles setup that must be done after parsing flags.
 func DoSetup() {
 	SetLoggerVerbosity()
+	logger.Info("Restore Key = %s", *timestamp)
 
 	InitializeConnection("postgres")
 	DoPostgresValidation()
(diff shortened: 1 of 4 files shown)


restore/restore.go
```diff
@@ -151,7 +151,7 @@ func restoreData(gucStatements []utils.StatementWithType) {
 
 	filteredMasterDataEntries := globalTOC.GetDataEntriesMatching(includeSchemas, includeTables)
 	totalTables := len(filteredMasterDataEntries)
-	dataProgressBar := utils.NewProgressBar(totalTables, "Tables restored: ", logger.GetVerbosity() == utils.LOGINFO)
+	dataProgressBar := utils.NewProgressBar(totalTables, "Tables restored: ", true)
```
khuddlefish: We introduced the showProgressBar argument to allow for some progress bars to show outside of INFO logging mode and others to hide. Now that they are all passing "true" I think we can remove that property.

utils/log.go
```diff
@@ -214,6 +214,6 @@ func NewProgressBar(count int, prefix string, showProgressBar bool) *pb.Progress
 	progressBar.ShowTimeLeft = false
 	progressBar.SetMaxWidth(100)
 	progressBar.SetRefreshRate(time.Millisecond * 200)
-	progressBar.NotPrint = !(showProgressBar && count > 0)
+	progressBar.NotPrint = !(showProgressBar && count > 0 && logger.GetVerbosity() == LOGINFO)
```
khuddlefish: I think this should be || not &&?

chrishajas: We want to print the progress bar when tables are present and the verbosity is LOGINFO. The NotPrint property = !(print conditions statement).

jmcatamney: Nope, it should be &&, it would be || if they weren't in parentheses; `!(count && info)` = `!count || !info`.

khuddlefish: Ok sounds good. I wasn't totally sure either way :P

restore/restore.go
khuddlefish: Actually, I think I want to keep this for the verbose solution I'm working on so nevermind. @chrishajas Can you push this soon so I can rebase on top for what I'm working on?

PR #23, opened 2017-12-19
Log in increments of 10% for progress bar in verbose mode

khuddlefish:
The previous progress bar doesn't work well when redirected to a file so
replace it with logging statements every 10% of progress when in verbose
mode (we guess users will often direct to a file when in this mode).

This is not used for tracking the progress of backing up or restoring data
for tables because we have a different mode of logging in that case.

### DIFF
--- a/backup/data.go
+++ b/backup/data.go
@@ -68,7 +68,7 @@ func BackupDataForAllTables(tables []Relation, tableDefs map[uint32]TableDefinit
 		}
 	}
 	totalRegTables := len(tables) - totalExtTables
-	dataProgressBar := utils.NewProgressBar(totalRegTables, "Tables backed up: ", logger.GetVerbosity() == utils.LOGINFO)
+	dataProgressBar := utils.NewProgressBar(totalRegTables, "Tables backed up: ", utils.PB_INFO)
 	dataProgressBar.Start()
 	backupFile := ""
 	if *singleDataFile {
--- a/backup/data.go
+++ b/backup/data.go
@@ -68,7 +68,7 @@ func BackupDataForAllTables(tables []Relation, tableDefs map[uint32]TableDefinit
 		}
 	}
 	totalRegTables := len(tables) - totalExtTables
-	dataProgressBar := utils.NewProgressBar(totalRegTables, "Tables backed up: ", true)
+	dataProgressBar := utils.NewProgressBar(totalRegTables, "Tables backed up: ", logger.GetVerbosity() == utils.LOGINFO)
 	dataProgressBar.Start()
 	backupFile := ""
 	if *singleDataFile {
--- a/integration/parallel_queries_test.go
+++ b/integration/parallel_queries_test.go
@@ -84,14 +84,14 @@ var _ = Describe("backup, utils, and restore integration tests related to parall
 			It("can execute all statements in the list serially", func() {
 				shouldExecute := utils.NewEmptyIncludeSet()
 				expectedOrderArray := []string{"1", "2", "3", "4"}
-				restore.ExecuteStatements(statements, "", 0, shouldExecute, false)
+				restore.ExecuteStatements(statements, "", utils.PB_NONE, shouldExecute, false)
 				resultOrderArray := utils.SelectStringSlice(tempConn, orderQuery)
 				Expect(resultOrderArray).To(Equal(expectedOrderArray))
 			})
 			It("can execute all statements in the list that are not of the specified object type serially", func() {
 				shouldExecute := utils.NewExcludeSet([]string{"DATABASE"})
 				expectedOrderArray := []string{"1", "3"}
-				restore.ExecuteStatements(statements, "", 0, shouldExecute, false)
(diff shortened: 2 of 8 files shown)


khuddlefish:
Sample output
```
20171218:17:28:25 gpbackup:khuddleston:karens_mac:089325-[DEBUG]:-Locks acquired:  10% (780/7798)
20171218:17:28:25 gpbackup:khuddleston:karens_mac:089325-[DEBUG]:-Locks acquired:  20% (1560/7798)
20171218:17:28:25 gpbackup:khuddleston:karens_mac:089325-[DEBUG]:-Locks acquired:  30% (2340/7798)
20171218:17:28:26 gpbackup:khuddleston:karens_mac:089325-[DEBUG]:-Locks acquired:  40% (3120/7798)
20171218:17:28:26 gpbackup:khuddleston:karens_mac:089325-[DEBUG]:-Locks acquired:  50% (3899/7798)
20171218:17:28:26 gpbackup:khuddleston:karens_mac:089325-[DEBUG]:-Locks acquired:  60% (4679/7798)
20171218:17:28:26 gpbackup:khuddleston:karens_mac:089325-[DEBUG]:-Locks acquired:  70% (5459/7798)
20171218:17:28:26 gpbackup:khuddleston:karens_mac:089325-[DEBUG]:-Locks acquired:  80% (6239/7798)
20171218:17:28:27 gpbackup:khuddleston:karens_mac:089325-[DEBUG]:-Locks acquired:  90% (7019/7798)
20171218:17:28:27 gpbackup:khuddleston:karens_mac:089325-[DEBUG]:-Locks acquired:  100% (7798/7798)
```

utils/log.go
```diff
@@ -209,11 +209,63 @@ func formatStackTrace(err error) string {
 /*
  * Progress bar functions
  */
-func NewProgressBar(count int, prefix string, showProgressBar bool) *pb.ProgressBar {
+func NewProgressBar(count int, prefix string, showInVerbose bool) ProgressBar {
+	if showInVerbose && logger.GetVerbosity() >= LOGVERBOSE {
+		return NewVerboseProgressBar(count, prefix)
```
chrishajas: Do we want to print the progress increments to the log if the user does not use verbose mode? For everything else, we print all verbose statements to the log whether the user uses verbose, normal, or quiet logging. In this case, we'll only print this to stdout or the log if the user uses verbose logging.

utils/log.go
```diff
@@ -209,11 +209,63 @@ func formatStackTrace(err error) string {
...
+}
+
+/*
+ * If progress bar reaches a percentage that is a multiple of 10, log a message to stdout
+ * We increment nextPercentToPrint so the same percentage will not be printed multiple times
+ */
+func (vpb *VerboseProgressBar) checkPercent() {
+	currPercent := int(float64(vpb.current) / float64(vpb.total) * 100)
+	//closestMult is the nearest percentage <= currPercent that is a multiple of 10
+	closestMult := currPercent / 10 * 10
```
chrishajas: We probably can replace 10 with a constant.

utils/log.go
```diff
@@ -209,11 +209,63 @@ func formatStackTrace(err error) string {
...
+		vpb.checkPercent()
+	}
+	return vpb.current
+}
+
+/*
+ * If progress bar reaches a percentage that is a multiple of 10, log a message to stdout
+ * We increment nextPercentToPrint so the same percentage will not be printed multiple times
+ */
+func (vpb *VerboseProgressBar) checkPercent() {
```
chrishajas: We should add unit tests to verify/prevent regressions in this logic.

khuddlefish: Oh haha yup. Definitely intended to do that. Thanks for the catch

backup/data.go
```diff
@@ -68,7 +68,7 @@ func BackupDataForAllTables(tables []Relation, tableDefs map[uint32]TableDefinit
 		}
 	}
 	totalRegTables := len(tables) - totalExtTables
-	dataProgressBar := utils.NewProgressBar(totalRegTables, "Tables backed up: ", true)
+	dataProgressBar := utils.NewProgressBar(totalRegTables, "Tables backed up: ", logger.GetVerbosity() == utils.LOGINFO)
```
khuddlefish: Note to me: now that this is 'showInVerbose' not 'showProgressBar' this should probably simply be true/false not conditional upon the actual logging level.

integration/parallel_queries_test.go
```diff
@@ -84,14 +84,14 @@ var _ = Describe("backup, utils, and restore integration tests related to parall
 			It("can execute all statements in the list serially", func() {
 				shouldExecute := utils.NewEmptyIncludeSet()
 				expectedOrderArray := []string{"1", "2", "3", "4"}
-				restore.ExecuteStatements(statements, "", false, shouldExecute, false)
+				restore.ExecuteStatements(statements, "", 0, shouldExecute, false)
```
jmcatamney: These four unit tests should probably take in PB_NONE here instead of 0 so the behavior is clearer.

utils/log.go
khuddlefish: Tests added in new commit

PR #24, opened 2017-12-20
Add and clarify information in report file

jmcatamney:
This commit adds the start time, end time, and duration of a backup to the
report file, and also breaks out the "backup type" string into separate fields
for the various parameters to make it easier to understand.

### DIFF
--- a/backup/backup.go
+++ b/backup/backup.go
@@ -246,16 +246,18 @@ func DoTeardown() {
 		}
 		reportFilename := globalCluster.GetReportFilePath()
 		configFilename := globalCluster.GetConfigFilePath()
-		backupReport.WriteReportFile(reportFilename, globalCluster.Timestamp, objectCounts, errMsg)
-		backupReport.WriteConfigFile(configFilename)
-		utils.EmailReport(globalCluster)
-		// We sleep for 1 second to ensure multiple backups do not start within the same second.
-		time.Sleep(1000 * time.Millisecond)
+
+		time.Sleep(time.Second) // We sleep for 1 second to ensure multiple backups do not start within the same second.
 		timestampLockFile := fmt.Sprintf("/tmp/%s.lck", globalCluster.Timestamp)
 		err := os.Remove(timestampLockFile)
 		if err != nil {
 			logger.Warn("Failed to remove lock file %s.", timestampLockFile)
 		}
+
+		endTime := time.Now()
+		backupReport.WriteConfigFile(configFilename)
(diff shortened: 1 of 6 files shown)


utils/report.go
```diff
@@ -114,23 +124,59 @@ gpbackup Version: %s
...
+	startTime, _ := time.ParseInLocation("20060102150405", timestamp, time.Local)
+	duration := endTime.Sub(startTime).String()
+	startTimestamp := startTime.Format("2006-01-02 15:04:05")
+	endTimestamp := endTime.Format("2006-01-02 15:04:05")
+	return startTimestamp, endTimestamp, reformatDuration(duration)
+}
+
+// Turns "1h2m3.456s" into "1:02:03"
+func reformatDuration(duration string) string {
+	re := regexp.MustCompile("(?:([0-9]+)h)?(?:([0-9]+)m)?([0-9]+)(?:\\.[0-9]+)?s")
```
chrishajas: Instead of parsing the duration as a string with regex, would something like https://stackoverflow.com/questions/47341278/how-to-format-a-duration-in-golang be easier?

utils/report.go
```diff
@@ -49,41 +52,48 @@ func ParseErrorMessage(errStr string) (string, int) {
...
 	}
-	filesStr := ""
-	if singleDataFile {
+	filesStr := "One File Per Table"
+	if ddlOnly {
+		filesStr = "No Data Files"
+	} else if singleDataFile {
 		report.SingleDataFile = true
-		filesStr = " With One Data File Per Segment"
+		filesStr = "One File Per Segment"
```
khuddlefish: It might be more clear to say "Single Data File" so it matches the argument that created it?

backup/backup.go
```diff
@@ -246,11 +246,13 @@ func DoTeardown() {
 		}
 		reportFilename := globalCluster.GetReportFilePath()
 		configFilename := globalCluster.GetConfigFilePath()
-		backupReport.WriteReportFile(reportFilename, globalCluster.Timestamp, objectCounts, errMsg)
+		endTime := time.Now()
```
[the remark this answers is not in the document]
jmcatamney: I'll move it.  Note that there's still no way to guarantee the report file and log file end times match, though, because the act of emailing the report file can take more time between the report file being written and the success message being printed.

utils/report.go
khuddlefish: I don't have strong feeling about it

jmcatamney:
@chrishajas @khuddlefish Made the requested changes, let me know if you have any other suggestions.

PR #25, opened 2017-12-20
Genericize functions in cluster.go

jmcatamney:
Previously, many functions in cluster.go used to execute commands on the segments
were mostly copy-pasted from one another.  This commit adds a new function,
ExecuteCommandAndCheckErrors, that puts all the logic to generate and execute
commands and report errors from the command in one place and leaves the existing
functions as wrappers.

Also, the new function returns command output instead of just reporting errors,
so functions can check return values to e.g. report exact file counts.

### DIFF
--- a/utils/cluster.go
+++ b/utils/cluster.go
@@ -6,6 +6,7 @@ package utils
  */
 
 import (
+	"bytes"
 	"fmt"
 	"os/exec"
 	"path"
@@ -49,7 +50,7 @@ func SetCompressionParameters(compress bool, compression Compression) {
 
 type Executor interface {
 	ExecuteLocalCommand(commandStr string) error
-	ExecuteClusterCommand(commandMap map[int][]string) map[int]error
+	ExecuteClusterCommand(commandMap map[int][]string) *RemoteOutput
 }
 
 // This type only exists to allow us to mock Execute[...]Command functions for testing
@@ -69,6 +70,13 @@ type SegConfig struct {
 	ContentID int
 	Hostname  string
 	DataDir   string
+}
+
+type RemoteOutput struct {
+	NumErrors int
+	Stdouts   map[int]string
+	Stderrs   map[int]string
+	Errors    map[int]error
 }
 
 /*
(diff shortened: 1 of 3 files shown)


utils/cluster.go
```diff
@@ -120,220 +128,205 @@ func (cluster *Cluster) GenerateSSHCommandMapForSegments(generateCommand func(in
...
+	for contentID := range remoteOutput.Errors {
+		logger.Verbose("%s on segment %d on host %s", messageFunc(contentID), contentID, cluster.GetHostForContent(contentID))
 	}
-	cluster.LogFatalError("Backup files missing", numErrors)
+	cluster.LogFatalError(finalErrMsg, remoteOutput.NumErrors)
 }
 
 func (cluster *Cluster) VerifyBackupDirectoriesExistOnAllHosts() {
-	commandMap := cluster.GenerateSSHCommandMapForCluster(func(contentID int) string {
+	cluster.ExecuteCommandAndCheckErrors("Verifying backup directories exist", "Backup directories missing or inaccessible", func(contentID int) string {
```
khuddlefish: Does the new function allow us to get more information about this command as well?

utils/cluster.go
```diff
@@ -120,220 +128,205 @@ func (cluster *Cluster) GenerateSSHCommandMapForSegments(generateCommand func(in
...
-	if fileCount != 1 {
-		s = "s"
+	remoteOutput := cluster.ExecuteClusterCommand(commandMap)
+	if remoteOutput.NumErrors == 0 {
+		return
 	}
-	for contentID := range errMap {
-		logger.Verbose("Expected to see %d backup file%s on segment %d, but some were missing.", fileCount, s, contentID)
+	for contentID := range remoteOutput.Errors {
+		logger.Verbose("%s on segment %d on host %s", messageFunc(contentID), contentID, cluster.GetHostForContent(contentID))
```
chrishajas: Does stderr, output, or err give us some relevant information that we want to log in all cases? I've noticed when we see these errors, they're not super helpful to us (many times they're a symptom of the root cause). If the stdout/sterr provides any more information, we should log that, even if it's only in verbose/debug.

utils/cluster.go
chrishajas: If it doesn't exist, it could mean they used an incorrect timestamp or forgot to use/used an incorrect backupdir. If we're only running `test -d`, I'm not sure what would make it "inaccessible"

utils/cluster.go
```diff
@@ -120,220 +128,205 @@ func (cluster *Cluster) GenerateSSHCommandMapForSegments(generateCommand func(in
...
+			return fmt.Sprintf("Unable to copy file %s", cluster.GetSegmentTOCFilePath(cluster.SegDirMap[contentID], fmt.Sprintf("%d", contentID)))
+		}, func(contentID int) string {
+			tocFile := cluster.GetSegmentTOCFilePath(cluster.SegDirMap[contentID], fmt.Sprintf("%d", contentID))
+			tocFilename := fmt.Sprintf("gpbackup_%d_%s_toc.yaml", contentID, cluster.Timestamp)
+			return fmt.Sprintf("cp -f %s/%s %s", cluster.GetDirForContent(contentID), tocFilename, tocFile)
+		},
+		true)
+}
+
+func (cluster *Cluster) VerifyBackupFileCountOnSegments(fileCount int) {
```
chrishajas: In this case we're expecting the command to pass, but parsing the stdout. Could we instead still run ExecuteCommandAndCheckErrors, but have that function return the stdout that we can parse after on a case-by-case basis? This

PR #29, opened 2018-01-12
Add support for backing up functions with WINDOW attribute

chrishajas:

### DIFF
--- a/backup/predata_functions_test.go
+++ b/backup/predata_functions_test.go
@@ -225,12 +225,20 @@ $_$`)
 					Expect(buffer.Contents()).To(Equal([]byte{}))
 				})
 			})
+			Context("Prints 'WINDOW' if IsWindow is set", func() {
+				It("print if window is returned true", func() {
+					funcDef.IsWindow = true
+					backup.PrintFunctionModifiers(backupfile, funcDef)
+					testutils.ExpectRegexp(buffer, "WINDOW")
+				})
+			})
 			It("prints config statements if any are set", func() {
 				funcDef.Config = "SET client_min_messages TO error"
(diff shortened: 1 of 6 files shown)


backup/predata_functions_test.go
```diff
@@ -225,12 +225,20 @@ $_$`)
 					Expect(buffer.Contents()).To(Equal([]byte{}))
 				})
 			})
+			Context("Prints 'WINDOW' if IsWindow is set", func() {
```
khuddlefish: This message is more the type that we usually have describing a specific It case, not a Context block.

backup/queries_functions.go
```diff
@@ -58,10 +59,14 @@ func GetFunctions(connection *utils.DBConn) []Function {
 		}
 		return functions
 	}
-	return GetFunctions5(connection)
+	return GetFunctionsMaster(connection)
```
jmcatamney: The name should probably be changed to make it clear that this function is used for both 5.X and 6.X.

khuddlefish: We actually have a similar pattern in some of the query files where we specify a different query for 4 vs. 5 or later. We decided to call them "version4query" and "masterQuery" so this would be consistent with that.

backup/predata_functions_test.go
chrishajas: Agreed, I'll change this.

backup/queries_functions.go
chrishajas: Left this function name as as `GetFunctionsMaster`, but changed `GetFunctions` to `GetFunctionsAllVersions`

PR #36, opened 2018-01-26
Add interrupt handler to gpbackup and gprestore

jmcatamney:
To ensure that the cluster is not left in a bad state if someone aborts a backup
or restore with Ctrl-C, this commit moves cleanup-related functions that were
previously called with defer to a DoCleanup function and sets up a signal handler
goroutine to call DoCleanup and exit if an interrupt is received.  DoTeardown
and several other functions are modified to account for a possible abort in the
middle of the function.

### DIFF
--- a/backup/backup.go
+++ b/backup/backup.go
@@ -4,6 +4,7 @@ import (
 	"flag"
 	"fmt"
 	"os"
+	"sync"
 	"time"
 
 	"github.com/greenplum-db/gpbackup/utils"
@@ -37,8 +38,11 @@ func initializeFlags() {
 
 // This function handles setup that can be done before parsing flags.
 func DoInit() {
+	CleanupGroup = &sync.WaitGroup{}
+	CleanupGroup.Add(1)
 	SetLogger(utils.InitializeLogging("gpbackup", ""))
 	initializeFlags()
+	InitializeSignalHandler()
 }
 
 func DoFlagValidation() {
@@ -230,12 +234,22 @@ func DoTeardown() {
 	errStr := ""
 	if err := recover(); err != nil {
 		errStr = fmt.Sprintf("%v", err)
-		fmt.Println(err)
 	}
-	errMsg, exitCode := utils.ParseErrorMessage(errStr)
-	if connection != nil {
-		connection.Close()
+	if wasTerminated {
+		/*
+		 * Don't print an error or create a report file if the backup was canceled,
+		 * as the signal handler will take care of cleanup and return codes.  Just
+		 * wait until the signal handler's DoCleanup completes so the main goroutine
+		 * doesn't exit while cleanup is still in progress.
+		 */
+		CleanupGroup.Wait()
+		return
 	}
+	if errStr != "" {
+		fmt.Println(errStr)
+	}
+	errMsg := utils.ParseErrorMessage(errStr)
+	errorCode := utils.GetErrorCode()
 
 	/*
 	 * Only create a report file if we fail after the cluster is initialized
@@ -244,7 +258,7 @@ func DoTeardown() {
 	if globalCluster.Timestamp != "" {
 		_, statErr := os.Stat(globalCluster.GetDirForContent(-1))
 		if statErr != nil { // Even if this isn't os.IsNotExist, don't try to write a report file in case of further errors
-			os.Exit(exitCode)
+			os.Exit(errorCode)
 		}
 		reportFilename := globalCluster.GetReportFilePath()
 		configFilename := globalCluster.GetConfigFilePath()
@@ -262,8 +276,35 @@ func DoTeardown() {
 		utils.EmailReport(globalCluster)
 	}
 
-	if exitCode == 0 {
+	DoCleanup()
+
+	if errorCode == 0 {
 		logger.Info("Backup completed successfully")
 	}
-	os.Exit(exitCode)
+	os.Exit(errorCode)
+}
+
+func DoCleanup() {
+	defer func() {
+		if err := recover(); err != nil {
+			logger.Warn("Encountered error during cleanup: %v", err)
+		}
+		logger.Verbose("Cleanup complete")
+		CleanupGroup.Done()
+	}()
+	logger.Verbose("Beginning cleanup")
+	if *singleDataFile {
+		CleanUpSegmentPipesOnAllHosts(globalCluster)
+		CleanUpSegmentTailProcesses(globalCluster)
+		if wasTerminated { // These should all end on their own in a successful backup
+			utils.TerminateHangingCopySessions(connection, globalCluster, "gpbackup")
+		}
+	}
+	if globalCluster.Timestamp != "" {
+		timestampLockFile := fmt.Sprintf("/tmp/%s.lck", globalCluster.Timestamp)
+		os.Remove(timestampLockFile) // Don't check error, as DoTeardown may have removed it already
+	}
+	if connection != nil {
+		connection.Close()
+	}
 }
--- a/end_to_end/end_to_end_suite_test.go
+++ b/end_to_end/end_to_end_suite_test.go
@@ -3,12 +3,14 @@ package end_to_end_test
 import (
 	"fmt"
 	"io/ioutil"
+	"math/rand"
 	"os"
 	"os/exec"
 	"path/filepath"
 	"regexp"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/greenplum-db/gpbackup/testutils"
 	"github.com/greenplum-db/gpbackup/utils"
@@ -86,11 +88,11 @@ var _ = Describe("backup end to end integration tests", func() {
 		exec.Command("dropdb", "testdb").Run()
 		err = exec.Command("createdb", "testdb").Run()
 		if err != nil {
-			Fail(fmt.Sprintf("%v", err))
+			Fail(fmt.Sprintf("Could not create testdb: %v", err))
 		}
 		err = exec.Command("createdb", "restoredb").Run()
 		if err != nil {
-			Fail(fmt.Sprintf("%v", err))
+			Fail(fmt.Sprintf("Could not create restoredb: %v", err))
 		}
 		backupConn = utils.NewDBConn("testdb")
 		backupConn.Connect(1)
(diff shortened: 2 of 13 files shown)


utils/db.go
```diff
@@ -278,3 +278,15 @@ func SelectStringSlice(connection *DBConn, query string, whichConn ...int) []str
 	}
 	return retval
 }
+
+// TODO: Uniquely identify COPY commands in the multiple data file case to allow terminating sessions
```
chrishajas: Does this TODO still apply? It seems like we're identifying the COPY by its PID.

restore/remote.go
```diff
@@ -72,7 +72,7 @@ func CleanUpHelperFilesOnAllHosts(cluster utils.Cluster) {
 	remoteOutput := cluster.GenerateAndExecuteCommand("Removing oid list and helper script files from segment data directories", func(contentID int) string {
 		oidFile := cluster.GetSegmentHelperFilePath(contentID, "oid")
 		scriptFile := cluster.GetSegmentHelperFilePath(contentID, "script")
-		return fmt.Sprintf("rm %s && rm %s", oidFile, scriptFile)
+		return fmt.Sprintf("rm -f %s && rm -f %s", oidFile, scriptFile)
```
chrishajas: Is there a reason for `-f` here? Is it possible for these files to not exist/be cleaned up previously?

restore/wrappers.go
```diff
@@ -30,11 +32,30 @@ func SetLoggerVerbosity() {
...
+SET gp_default_storage_options='';
+`
+	for i := 0; i < connection.NumConns; i++ {
+		connection.MustExec(setupQuery, i)
+	}
+}
+
+func InitializeSignalHandler() {
+	signalChan := make(chan os.Signal, 1)
+	signal.Notify(signalChan, os.Interrupt)
```
chrishajas: Do we handle a `kill` signal here in addition to interrupt?

restore/remote.go
```diff
@@ -141,10 +141,36 @@ func VerifyHelperVersionOnSegments(cluster utils.Cluster, version string) {
...
+	cluster.CheckClusterError(remoteOutput, "Unable to clean up segment data pipes", func(contentID int) string {
+		return "Unable to clean up segment data pipe"
+	})
+}
+
 func CleanUpSegmentTOCs(cluster utils.Cluster) {
 	remoteOutput := cluster.GenerateAndExecuteCommand("Removing segment table of contents files from segment data directories", func(contentID int) string {
 		tocFile := cluster.GetSegmentTOCFilePathWithPID(cluster.SegDirMap[contentID], fmt.Sprintf("%d", contentID))
-		return fmt.Sprintf("rm %s", tocFile)
+		return fmt.Sprintf("rm -f %s", tocFile)
```
chrishajas: Is the `-f` necessary? In wht case would the TOC file not exist?

restore/remote.go
```diff
@@ -141,10 +141,36 @@ func VerifyHelperVersionOnSegments(cluster utils.Cluster, version string) {
...
+		 * the time DoCleanup is called.
+		 */
+		return fmt.Sprintf("PIDS=`ps ux | grep \"%s\" | grep \"%s\" | grep -v grep | awk '{print $2}'`; if [[ ! -z \"$PIDS\" ]]; then kill -9 $PIDS; fi", procPattern)
+	})
+	cluster.CheckClusterError(remoteOutput, "Unable to clean up restore agent processes", func(contentID int) string {
+		return "Unable to clean up restore agent process"
+	})
+}
+
+func CleanUpSegmentPipesOnAllHosts(cluster utils.Cluster) {
```
chrishajas: I see this is used on the backup side, but is this function used anywhere for restore?

khuddlefish: It does not appear to be and shouldn't be. It won't actually do the right thing if called on the restore side.

restore/restore.go
```diff
@@ -225,12 +237,41 @@ func DoTeardown() {
...
+	defer func() {
+		if err := recover(); err != nil {
+			logger.Warn("Encountered error during cleanup: %v", err)
+		}
+		logger.Verbose("Cleanup complete")
+		CleanupGroup.Done()
+	}()
+	logger.Verbose("Beginning cleanup")
+	if backupConfig.SingleDataFile {
+		CleanUpSegmentHelperProcesses(globalCluster)
```
chrishajas: In a case with no error (eg: normal gprestore exit), the helper processes should have already exited and not required cleanup right?

restore/remote.go
```diff
@@ -141,10 +141,36 @@ func VerifyHelperVersionOnSegments(cluster utils.Cluster, version string) {
...
+	remoteOutput := cluster.GenerateAndExecuteCommand("Cleaning up segment data pipes", func(contentID int) string {
+		pipePath := cluster.GetSegmentPipeFilePath(contentID)
```
khuddlefish: If this was used on the restore side, this should be GetSegmentPipeFilePathWithPID. It might actually make sense for us to move those functions to backup/restore respectively. Maybe not part of this commit though.

end_to_end/end_to_end_suite_test.go
```diff
@@ -318,5 +330,54 @@ var _ = Describe("backup end to end integration tests", func() {
...
+				rng := rand.New(rand.NewSource(time.Now().UnixNano()))
+				time.Sleep(time.Duration(rng.Intn(1000)+500) * time.Millisecond)
+				cmd.Process.Signal(os.Interrupt)
+			}()
+			output, _ := cmd.CombinedOutput()
+			stdout := string(output)
+
+			Expect(stdout).To(ContainSubstring("Received an interrupt, aborting backup process"))
+			Expect(stdout).To(ContainSubstring("Cleanup complete"))
+			Expect(stdout).To(Not(ContainSubstring("CRITICAL")))
```
chrishajas: Would it be worth doing a `ps ux` and asserting no processes still exist anywhere, or is that a bit overkill?

backup/backup.go
```diff
@@ -262,8 +274,33 @@ func DoTeardown() {
...
+			logger.Warn("Encountered error during cleanup: %v", err)
+		}
+		logger.Verbose("Cleanup complete")
+		CleanupGroup.Done()
+	}()
+	logger.Verbose("Beginning cleanup")
+	if *singleDataFile {
+		CleanUpSegmentPipesOnAllHosts(globalCluster)
+		CleanUpSegmentTailProcesses(globalCluster)
+		utils.TerminateHangingCopySessions(connection, globalCluster, "gpbackup")
```
chrishajas: We also don't necessarily have to run this if we exit normally.

backup/backup.go
```diff
@@ -230,12 +234,20 @@ func DoTeardown() {
...
+		/*
+		 * Don't print an error or create a report file if the backup was canceled,
+		 * as the signal handler will take care of cleanup and return codes.  Just
+		 * wait until the signal handler's DoCleanup completes so the main goroutine
+		 * doesn't exit while cleanup is still in progress.
+		 */
+		CleanupGroup.Wait()
+		return
 	}
+	fmt.Println(errStr)
```
chrishajas: When errStr is empty, it will still print an empty line. Is this desired?

utils/db.go
jmcatamney: We identify the COPY by its PID in the single file case only, not in the multiple-file case (c.f. GetTableBackupFilePathForCopyCommand).

restore/remote.go
jmcatamney: Yes, in two cases.  If a Ctrl-C is received during DoTeardown both the normal DoCleanup and the signal handler DoCleanup will try to remove them, and if restore errors out early (e.g. trying to --include-schema a bad schema) it'll still call DoCleanup even if no files were created yet.

restore/wrappers.go
jmcatamney: Just interrupt for now, as that's all that was in the story and we may leave to allow a `kill -9` for "No, seriously, die now and don't try to clean up" cases.

restore/remote.go
jmcatamney: If the cleanup happens before the files are created, as mentioned in a previous comment, we don't want a bunch of extraneous error messages at the end.

restore/remote.go
jmcatamney: Yep, just copy-pasted it into restore/remote.go by mistake when I was moving some things around.  I'll delete it.

chrishajas: I think there's already an identical function in backup/remote.go actually, so I think this function can just be removed.

restore/restore.go
jmcatamney: Correct, in a successful restore everything should end on its own.  I can add an `if wasTerminated` check here.

end_to_end/end_to_end_suite_test.go
jmcatamney: Unless we want to get really paranoid and assert there are no `tail`s, no `gpbackup_helper`s, no pipes, and so forth, it's probably overkill.

backup/backup.go
jmcatamney: Nope, good catch.

restore/remote.go
jmcatamney: Yeah, the filename functions were in utils originally since backup and restore both use the same files, so if they're different we can move them to remote.go later.

restore/restore.go
```diff
@@ -266,10 +268,12 @@ func DoCleanup() {
 	}()
 	logger.Verbose("Beginning cleanup")
 	if backupConfig.SingleDataFile {
-		CleanUpSegmentHelperProcesses(globalCluster)
 		CleanUpSegmentTOCs(globalCluster)
-		CleanUpHelperFilesOnAllHosts(globalCluster)
-		utils.TerminateHangingCopySessions(connection, globalCluster, "gprestore")
+		if wasTerminated { // These should all end on their own in a successful restore
+			CleanUpSegmentHelperProcesses(globalCluster)
+			CleanUpHelperFilesOnAllHosts(globalCluster)
```
chrishajas: We might always need to remove the oid and helper script files (outside the wasTerminated). I don't think they're removed somewhere else?

jmcatamney: Fixed.

PR #37, opened 2018-01-27
Add support for clustering based on indexes.

khuddlefish:
Author: Vignesh Nandakumar <vignesh.n.iyer@gmail.com>


khuddlefish:
We use the QuerySimpleDefinition for Indexes, Rules and Triggers, but indexes are the only ones with TableSpaces and now IsClustered. Do we wanted to split Indexes out into their own struct?

jmcatamney:
@khuddlefish QuerySimpleDefinition was meant to be just a wrapper for the pg_get_[foo]definition functions, so if we're doing more with indexes now I think a split would be good.

khuddlefish:
Changes applied and merged

PR #38, opened 2018-01-30
Add signal handler and cleanup to gpbackup_helper

khuddlefish:
This prevents COPY from hanging indefinitely if an agent is killed. We also
create an error file if the agent is killed, so that the restore will not
erroneously report success if the agent is killed while restoring the last
table.

### DIFF
--- a/helper/helper.go
+++ b/helper/helper.go
@@ -10,12 +10,13 @@ import (
 	"sort"
 	"strconv"
 	"strings"
+	"sync"
 	"syscall"
 
 	"github.com/greenplum-db/gpbackup/utils"
 )
 
-var (
+var ( // Shared globals
 	content      *int
 	dataFile     *string
 	logger       *utils.Logger
@@ -28,12 +29,24 @@ var (
 	version      string
 )
 
+var ( // Globals for restore only
+	CleanupGroup  *sync.WaitGroup
+	currentPipe   string
+	lastPipe      string
+	nextPipe      string
+	wasTerminated bool
+	writer        *bufio.Writer
+	writeHandle   *os.File
+)
+
 /*
  * Shared functions
  */
 
 func DoHelper() {
+	defer DoTeardown()
 	InitializeGlobals()
+	utils.InitializeSignalHandler(DoCleanup, fmt.Sprintf("restore agent on segment %d", *content), &wasTerminated)
 	if *restoreAgent {
 		doRestoreAgent()
 	} else {
@@ -42,6 +55,8 @@ func DoHelper() {
 }
 
 func InitializeGlobals() {
+	CleanupGroup = &sync.WaitGroup{}
+	CleanupGroup.Add(1)
 	content = flag.Int("content", -2, "Content ID of the corresponding segment")
 	dataFile = flag.String("data-file", "", "Absolute path to the data file")
 	logger = utils.InitializeLogging("gpbackup_helper", "")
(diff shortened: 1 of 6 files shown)


restore/remote.go
```diff
@@ -70,9 +70,10 @@ func WriteOidListToSegments(cluster utils.Cluster, filteredEntries []utils.Maste
 
 func CleanUpHelperFilesOnAllHosts(cluster utils.Cluster) {
 	remoteOutput := cluster.GenerateAndExecuteCommand("Removing oid list and helper script files from segment data directories", func(contentID int) string {
+		errorFile := fmt.Sprintf("%s_error", cluster.GetSegmentPipeFilePathWithPID(contentID))
 		oidFile := cluster.GetSegmentHelperFilePath(contentID, "oid")
 		scriptFile := cluster.GetSegmentHelperFilePath(contentID, "script")
-		return fmt.Sprintf("rm -f %s && rm -f %s", oidFile, scriptFile)
+		return fmt.Sprintf("rm -f %s; rm -f %s; rm -f %s", errorFile, oidFile, scriptFile)
```
chrishajas: This will only fail if it fails to remove the scriptFile for some reason. Any particular reason we want to remove the `&&`?

helper/helper.go
```diff
@@ -186,28 +197,78 @@ func getPipeReader() *bufio.Reader {
...
+	if writer != nil {
+		err := writer.Flush()
+		utils.CheckError(err)
+		writer = nil
+	}
+	if writeHandle != nil {
+		err := writeHandle.Close()
+		utils.CheckError(err)
+		writeHandle = nil
+	}
```
chrishajas: I don't expect writer/writeHandle to ever be nil in this case, during normal operation, but if it were, we'd just be ignoring the error. Before, when we assumed it wasn't nil, it would seg fault (which is better than failing silently).

restore/restore.go
```diff
@@ -270,8 +271,8 @@ func DoCleanup() {
 	if backupConfig.SingleDataFile {
 		CleanUpSegmentTOCs(globalCluster)
 		CleanUpHelperFilesOnAllHosts(globalCluster)
+		CleanUpSegmentHelperProcesses(globalCluster)
 		if wasTerminated { // These should all end on their own in a successful restore
-			CleanUpSegmentHelperProcesses(globalCluster)
```
chrishajas: What are the cases where we would actually need to clean up the segment helper processes? The only case I can think of is where we find an error on a segment using the error file. We might be able to use that fact and not make this call in the majority of cases.

khuddlefish: We are only checking for the error file if you reach the end of the restore table loop (i.e. an agent died only on the last table). If you kill an agent before the last table, copy will error out and not reach that check. It's possible if we wanted to log the agent error instead of the COPY error we could catch that internally and then check for the agent error file. That would help us avoid the confusion about it failing on the wrong table.

restore/remote.go
khuddlefish: We have a slight preference for the semicolon so all three commands will try to run regardless of whether one fails. We wouldn't want to miss cleaning up a file if one of them errored out.

helper/helper.go
khuddlefish: In the termination case, Cleanup might fire twice so setting to nil here and checking at the top prevents the second run from trying to close the file again. We wouldn't expect it to be nil in otherwise normal operation.

restore/remote.go
chrishajas: We should change the error messages below to reflect that. If the first fails, and the subsequent passes, it will not report an error at all. In this case, the error is really only checking that we cleaned up the script file.

khuddlefish:
Merged

PR #40, opened 2018-01-30
Add support for backing up extensions

chrishajas:
An extra clause to exclude objects that are part of an extension was added to each query. The total metadata backup time did not show any noticeable increase due to the addition of this clause. This clause is also included in queries on GPDB4 for simplicity, but extensions are only backed up/restored in GPDB5 or later.

### DIFF
--- a/backup/predata_functions.go
+++ b/backup/predata_functions.go
@@ -167,6 +167,15 @@ func PrintCreateCastStatements(metadataFile *utils.FileWithByteCount, toc *utils
 	}
 }
 
+func PrintCreateExtensionStatements(metadataFile *utils.FileWithByteCount, toc *utils.TOC, extensionDefs []Extension, extensionMetadata MetadataMap) {
+	for _, extensionDef := range extensionDefs {
+		start := metadataFile.ByteCount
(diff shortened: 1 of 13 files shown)


backup/predata_functions.go
```diff
@@ -167,6 +167,15 @@ func PrintCreateCastStatements(metadataFile *utils.FileWithByteCount, toc *utils
 	}
 }
 
+func PrintCreateExtensionStatements(metadataFile *utils.FileWithByteCount, toc *utils.TOC, extensionDefs []Extension, extensionMetadata MetadataMap) {
+	for _, extensionDef := range extensionDefs {
+		start := metadataFile.ByteCount
+		metadataFile.MustPrintf("\n\nSET search_path=%s,pg_catalog;\nCREATE EXTENSION IF NOT EXISTS %s WITH SCHEMA %s;\nSET search_path=pg_catalog;", extensionDef.Schema, extensionDef.Name, extensionDef.Schema)
```
khuddlefish: do extension defs always have a schema or could it be empty?

chrishajas: They always have a schema (it's a not NULL column on pg_extension).

PR #42, opened 2018-01-30
Add support for backing up base type storage options

chrishajas:


backup/queries_relations.go
```diff
@@ -354,7 +354,7 @@ func GetPartitionTemplates(connection *utils.DBConn) map[uint32]string {
 }
 
 func GetStorageOptions(connection *utils.DBConn) map[uint32]string {
```
khuddlefish: Is it confusing that we are calling both table storage options and base type storage options "StorageOptions"?

chrishajas: They do represent the same thing. For columns I think we call them encoding, but they all are "storage options", just for different objects.

khuddlefish: Sure, but I meant maybe we should qualify it with "TypeStorageOptions" and "TableStorageOptions" or something. They are essentially the same thing but on different objects.

chrishajas: Just the global function or the struct fields too?

khuddlefish: The struct fields are probably ok since they're qualified by being a part of the type object. Mostly the function.

PR #45, opened 2018-02-08
Split shared packages into gp-common-go-libs repo

chrishajas:
This splits shared functionality into the testhelper, structmatcher, operating, gplog, dbconn, and cluster packages in https://github.com/greenplum-db/gp-common-go-libs.


testutils/functions_test.go
```diff
@@ -1,12 +1,19 @@
...
 import (
+	"testing"
+
 	"github.com/greenplum-db/gpbackup/utils"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 	"github.com/onsi/gomega/gbytes"
 )
 
+func TestTestUtils(t *testing.T) {
```
khuddlefish: We should try to be consistent with where we register test packages

utils/cluster.go
```diff
@@ -7,244 +7,86 @@ package utils
...
-	ExecuteLocalCommand(commandStr string) (string, error)
-	ExecuteClusterCommand(commandMap map[int][]string) *RemoteOutput
-}
-
-// This type only exists to allow us to mock Execute[...]Command functions for testing
-type GPDBExecutor struct{}
-
-type Cluster struct {
-	ContentIDs             []int
+type FilePathInfo struct {
```
khuddlefish: We should change the name of this file

utils/log.go
```diff
@@ -5,221 +5,12 @@ package utils
...
-	"os"
 	"time"
 
-	"github.com/pkg/errors"
+	"github.com/greenplum-db/gp-common-go-libs/gplog"
 	pb "gopkg.in/cheggaaa/pb.v1"
 )
 
-var (
-	logger *Logger
```
khuddlefish: Rename this file?

PR #46, opened 2018-02-09
Use template0 database as template in CREATE DATABASE statements

chrishajas:

### DIFF
--- a/backup/metadata_globals.go
+++ b/backup/metadata_globals.go
@@ -26,7 +26,7 @@ SET client_encoding = '%s';
 func PrintCreateDatabaseStatement(metadataFile *utils.FileWithByteCount, toc *utils.TOC, db Database, dbMetadata MetadataMap) {
 	dbname := db.Name
 	start := metadataFile.ByteCount
-	metadataFile.MustPrintf("\n\nCREATE DATABASE %s", dbname)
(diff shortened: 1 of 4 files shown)


asimrp:
Is using `template0` as template absolutely unavoidable?  The reason is we don't have a way to update age of `template0` as of now (there will be a solution soon but it's not there as of now).  And that opens a possibility of wrap around or transaction ID stop / warn limits to be hit right after creating a new database.

chrishajas:
@asimrp We assume an empty database for a successful restore. All objects that are are created by a user are backed up. If a user creates an object in template1, and creates a new database based on it, those objects from template1 will still be backed up. When they restore it, if we created a database based on template1, they would see errors as the object already exists.

This is also how pg_dump/gpcrondump behave. If template1 was guaranteed to be empty, we could use that, but I assume some/many users have objects in template1 specific to their needs.

khuddlefish:
@asimrp When we create the database ourselves because the user passed the `-createdb` flag, we want to be sure to start from a completely empty database, so template0 seems to be the best option. If the user cannot use template0 due to the age, or other reasons, the workaround is for them to create a database themselves using whatever template works for them and then running gprestore without the `-createdb` flag. That way they still have a viable option even if template0 doesn't work for them. Does that sound like it would be alright to you?

PR #48, opened 2018-02-12
Do not attempt to restore session user or current user

khuddlefish:
These roles have to exist when you are running gprestore and an error will
occurr if you attempt to restore them as part of the global metadata. We
now get these user values from the database and skip over the roles if
they are in the global metadata.

### DIFF
--- a/utils/toc.go
+++ b/utils/toc.go
@@ -82,6 +82,8 @@ func (toc *SegmentTOC) WriteToFile(filename string) {
 }
 
 type StatementWithType struct {
+	Schema          string
+	Name            string
 	ObjectType      string
 	ReferenceObject string
 	Statement       string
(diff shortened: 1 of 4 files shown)


utils/toc.go
```diff
@@ -164,6 +167,19 @@ func SubstituteRedirectDatabaseInStatements(statements []StatementWithType, oldN
 	return statements
 }
 
+func RemoveActiveRoles(connection *dbconn.DBConn, statements []StatementWithType) []StatementWithType {
+	sessionUser := dbconn.MustSelectString(connection, "SELECT quote_ident(session_user) AS string")
```
chrishajas: We might be able to use connection.User instead of these queries for the user.

PR #56, opened 2018-02-16
Clean up extra logging if user interrupts backup or restore

chrishajas:
This commit adds guards to prevent misleading logging. Additionally, it
cleans up some extra progress bars that are printed if the user exits
the utility.

### DIFF
--- a/backup/backup.go
+++ b/backup/backup.go
@@ -132,10 +132,17 @@ func backupGlobal(metadataFile *utils.FileWithByteCount) {
 		BackupRoles(metadataFile)
 		BackupRoleGrants(metadataFile)
 	}
-	gplog.Info("Global database metadata backup complete")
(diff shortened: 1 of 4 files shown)


backup/backup.go
```diff
@@ -214,10 +220,17 @@ func backupData(tables []Relation, tableDefs map[uint32]TableDefinition) {
 	if *singleDataFile {
 		MoveSegmentTOCsAndMakeReadOnly()
 	}
-	gplog.Info("Data backup complete")
+	if wasTerminated {
+		gplog.Info("Data backup incomplete")
```
jmcatamney: You might want to add a block like this for the other stages as well, so whichever stage is running when a Ctrl-C happens gets a "[stage] [backup|restore] incomplete" message.

PR #57, opened 2018-02-16
Add filters to restore to match backup filters.

jmcatamney:
This PR adds --exclude-schema, --include-table, --exclude-table, and --exclude-table-file flags to gprestore, so that gpbackup and gprestore have the same set of filtering options.


end_to_end/end_to_end_suite_test.go
```diff
@@ -182,6 +189,24 @@ var _ = Describe("backup end to end integration tests", func() {
...
 		})
+		It("runs gpbackup and gprestore with exclude-table restore flag", func() {
+			timestamp := gpbackup(gpbackupPath)
+			gprestore(gprestorePath, timestamp, "-redirect-db", "restoredb", "-exclude-table", "schema2.foo2", "-exclude-table", "schema2.returns")
+
+			assertTablesCreated(restoreConn, 16)
+			assertDataRestored(restoreConn, map[string]int{"schema2.foo3": 100, "public.foo": 40000, "public.holds": 50000, "public.sales": 13})
+
+			os.Remove("/tmp/exclude-tables.txt")
+		})
```
chrishajas: Do we want a test for exclude-table-file?

restore/restore.go
```diff
@@ -22,8 +22,12 @@ func initializeFlags() {
 	backupDir = flag.String("backup-dir", "", "The absolute path of the directory in which the backup files to be restored are located")
 	createDB = flag.Bool("create-db", false, "Create the database before metadata restore")
 	debug = flag.Bool("debug", false, "Print verbose and debug log messages")
+	flag.Var(&excludeSchemas, "exclude-schema", "Restore all schemas except the specified schema(s). --exclude-schema can be specified multiple times.")
+	flag.Var(&excludeTables, "exclude-table", "Restore all tables except the specified table(s). --exclude-table can be specified multiple times.")
```
chrishajas: This is actually restore all metadata except the specified table. We should change this help message on the backup side too.

end_to_end/end_to_end_suite_test.go
jmcatamney: I figured a representative sample of tests was fine, since the command line and file flags only differ by an InitializeFilterLists call.

restore/restore.go
jmcatamney: Will do.

PR #62, opened 2018-03-21
Upload packaged gpbackup tar to s3 when tagged.

chrishajas:
Currently, this requires a human to manually tag a gpbackup version. When a new version is tagged, it will upload that binary to S3.

### DIFF
--- a/ci/pipeline.yml
+++ b/ci/pipeline.yml
@@ -18,6 +18,13 @@ resources:
     uri: https://github.com/greenplum-db/gpbackup
     branch: master
 
+- name: gpbackup_tagged_src
+  type: git
+  source:
+    branch: master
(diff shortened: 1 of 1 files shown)


chrishajas:
@kmacoskey 

As part of this, we would be committing a change to a manifest that lives in a repo somewhere. In the proposed manifest, there is a "artifact_version" parameter. Is this the S3 version? If so, is there a way to get this sha prior to uploading it to S3? Would you rather us upload different file names depending on the version (eg: foo/bar-1.2.3.tar.gz)? Another approach is using a github release resource (https://github.com/concourse/github-release-resource)

chrishajas:
Merged in commit a596c775b157d897274440084fb67a0bc0f011df

PR #64, opened 2018-03-29
gpbackup can dump the right schema for the replicated table

pengzhout:
We added a new distribution policy named "DISTRIBUTED REPLICATED", so
gpbackup need to recognize it and dump the schema correctly


pengzhout:
Hi @chrishajas @khuddlefish, can you help reviewing this?

jmcatamney:
@pengzhout Thanks for making a PR for the new feature!  Two comments:

1) Does the new distribution policy exist in 4.3, 5, and master (or will it be ported soon)?  If not, there will need to be different queries for versions that do and do not include it; see e.g. GetOperators in queries_operators.go for an example pattern of how to do that.

2) This PR should include a test for the new distribution policy in `integration/predata_relations_queries_test.go` to validate the new query.  You can use the existing GetDistributionPolicies tests as a template to make one, or I'm sure @khuddlefish or @chrishajas can help you with that.

pengzhout:
@jmcatamney sorry for the late response, thanks for the detailed guide, this feature only exists in master, I pushed a new commit to add version check and test case, please take a look.

jmcatamney:
@pengzhout The version check and the test look good to me.  Could you squash the two commits together with just the first commit's message before we merge the PR?

backup/queries_relations.go
```diff
@@ -328,28 +328,61 @@ func SelectAsOidToStringMap(connection *dbconn.DBConn, query string) map[uint32]
 
 func GetDistributionPolicies(connection *dbconn.DBConn, tables []Relation) map[uint32]string {
```
khuddlefish: tables is not being used anymore. We can take care of removing this before we merge.

pengzhout:
@jmcatamney many thanks.

PR #66, opened 2018-04-04
Ensure passing tests before release job in pipeline

kmacoskey:
This helps with understanding the pipeline when looking at it with the idea that tests are first and they block the final action which in this case is release. A tag in the source will still trigger the release job, but it will now require every test before it to be green as well.

@khuddlefish @chrishajas

### DIFF
--- a/ci/pipeline.yml
+++ b/ci/pipeline.yml
@@ -501,11 +501,7 @@ jobs:
     - get: gpbackup
       passed:
        - units
-       - scale-master
-       - scale-43-stable
-       - scale-5x-stable
        - integrations-GPDB5
-       - integrations-master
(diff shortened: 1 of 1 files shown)


chrishajas:
This would be more for visual purposes correct? A tag would not necessarily trigger any of the tests, but if any are red it would prevent us from uploading to S3. I would be ok with requiring everything except the scale and integrations-master tests, as the scale tests are triggered manually when we make performance changes to queries and integrations-master can go red often with catalog changes.

kmacoskey:
Correct, it does help with visibility and forces green tests before release. I wasn't sure which set of tests made sense so I defaulted to adding them all. I will trust your better judgement on which tests to block on.

PR #67, opened 2018-04-18
Use gpbackup_helper as an agent for backup with --single-data-file flag

chrishajas:
Previously, when performing a backup with --single-data-file, we used a long-running tail process as a pseudo agent. While this worked fine when we were simply writing to a file on the local system, it did not allow for error handling when using plugins and started to become cumbersome to add additional functionality.

Now, we create a backup agent the same way we do for restore. If an error occurs with the plugin, it will be reported in the gpbackup_helper log. 

Additionally, these commits combine many of the remote calls that were similar between backup and restore.


backup/backup.go
```diff
@@ -362,12 +380,11 @@ func DoCleanup() {
 	gplog.Verbose("Beginning cleanup")
 	if globalFPInfo.Timestamp != "" {
 		if *singleDataFile {
-			CleanUpSegmentPipesOnAllHosts()
-			CleanUpSegmentTailProcesses()
+			utils.CleanUpHelperFilesOnAllHosts(globalCluster, globalFPInfo)
 			if wasTerminated {
 				// It is possible for the COPY command to become orphaned if an agent process is killed
```
khuddlefish: technically this comment should go one line lower.

helper/helper.go
```diff
@@ -80,84 +87,135 @@ func InitializeGlobals() {
...
-	oid = &newoid
-}
+	oidList := getOidListFromFile()
 
-/*
- * Backup helper functions
- */
+	currentPipe = fmt.Sprintf("%s_%d", *pipeFile, oidList[0])
+	log(fmt.Sprintf("Opening pipe for oid %d", oidList[0]))
+	reader, readHandle := getBackupPipeReader(currentPipe)
```
khuddlefish: Did you want some kind of comment explaining why these needed to be in this order?

helper/helper.go
```diff
@@ -80,84 +87,135 @@ func InitializeGlobals() {
...
+	var finalWriter io.Writer
+	var gzipWriter *gzip.Writer
+	bufIoWriter := bufio.NewWriter(writeHandle)
+	finalWriter = bufIoWriter
+	if compressLevel > 0 {
+		gzipWriter, err = gzip.NewWriterLevel(bufIoWriter, compressLevel)
+		gplog.FatalOnError(err)
+		finalWriter = gzipWriter
+	}
+	return finalWriter, gzipWriter, bufIoWriter, writeHandle, writeCmd
```
khuddlefish: Do you think it is clear enough why we have these different writers?

helper/helper.go
```diff
@@ -80,84 +87,135 @@ func InitializeGlobals() {
...
-	reader := getPipeReader()
+	log(fmt.Sprintf("Opening pipe for oid %d", oidList[0]))
+	writer, writeHandle = getRestorePipeWriter(currentPipe)
+	reader := getRestorePipeReader()
 	for i, oid := range oidList {
 		log(fmt.Sprintf("Restoring table with oid %d", oid))
 		if i < len(oidList)-1 {
 			nextPipe = fmt.Sprintf("%s_%d", *pipeFile, oidList[i+1])
-			createNextPipe()
+			createPipe(nextPipe)
```
khuddlefish: Good change :)

helper/helper.go
```diff
@@ -287,8 +367,8 @@ func DoCleanup() {
 		handle := utils.MustOpenFileForWriting(fmt.Sprintf("%s_error", *pipeFile))
 		handle.Close()
 	}
-	if *restoreAgent {
-		flushAndCloseWriter()
+	if *restoreAgent || *backupAgent {
```
khuddlefish: These should actually be the only two cases now, so we should take this conditional out.

helper/helper.go
```diff
@@ -287,8 +367,8 @@ func DoCleanup() {
...
+		flushAndCloseRestoreWriter()
```
khuddlefish: Do we need to add something similar to clean up the backup writers?

integration/helper_test.go
```diff
@@ -0,0 +1,270 @@
...
+	})
+	AfterEach(func() {
+
+	})
+	It("runs backup gpbackup_helper without compression", func() {
+		f, _ := os.Create(oidFile)
+		f.WriteString("1\n2\n3\n")
+		helperCmd := gpbackupHelper(gpbackupHelperPath, "--backup-agent", "--compression-level", "0", "--data-file", testDir+dataFilename)
+		for i := 1; i <= 3; i++ {
+			output, err := exec.Command("bash", "-c", fmt.Sprintf("echo here is some data > %s_%d", pipeFile, i)).CombinedOutput()
```
khuddlefish: I guess it's not necessary, but I like wrapping the thing we are echoing in quotes if possible.

PR #73, opened 2018-04-30
Don't check error for mkfifo command

khuddlefish:
Our tests were failing frequently for "file exists" errors with this
command. This should not be a reason for the helper to error out. If the
command errors out and is unable to create the pipe, we should see the
error show up somewhere else, so this check is not very useful.


chrishajas:
Do we know why we're getting that the file already exists? Is it due to test pollution, the way we're setting up our tests, or is it an assertion that we can't guarantee due to a race condition in the logic? I agree that this change shouldn't change the logic, but I don't know under what conditions a file would already exist.

chrishajas:
Fixed in 1779f05e2bfa3f0a53e2f874bb22e80c4191b7b6

PR #80, opened 2018-05-24
Don't place cross compiled binaries into $GOPATH/bin

adam8157:
It helps nothing to place not executable binaries in $PATH I suppose.


khuddlefish:
@adam8157 Has your team experienced pain around this somehow? I do agree they don't really need to go into our executable path, but I'm not sure I like naming them something different. The main way we use these options currently is when we build Linux binaries from our Mac machine to give to our PM Oak for story validation. It would be slightly inconvenient for him to have to call *_linux binaries when he runs them.

adam8157:
Not a big deal, it just looked weird to me when we build Linux binaries from macOS to give PM a try 😆 

Updated to put them in project directory, feel free to take or reject.

PR #81, opened 2018-05-24
Move BoostFS setup to ddboost_plugin_tests job

chrishajas:
Also, use a faster instance type for 5X scale test and use 2 node cluster
instead of single node.

### DIFF
--- a/ci/pipeline.yml
+++ b/ci/pipeline.yml
@@ -117,7 +117,7 @@ resources:
       bucket: gpdb5-pipeline-dynamic-terraform
       bucket_path: clusters-google/
 
-- name: boostfs_terraform
+- name: ddboost_terraform
   type: terraform
   source:
     env:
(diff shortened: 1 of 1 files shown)


ci/pipeline.yml
```diff
@@ -549,7 +580,6 @@ jobs:
       terraform_source: ccp_src/google-nvme-block-device/
       vars:
         instance_type: n1-highmem-2
-        number_of_nodes: 1
         ccp_reap_minutes: 720
```
khuddlefish: Do we still need this here? Do we need to add it to other jobs?

chrishajas: Nope, I meant to remove that. All of the scale tests are now also doing data, so it's good to do a true multi-node test.

khuddlefish: My comment was actually meant to be about about ccp_reap_minutes

chrishajas: That I'm not sure of. I don't know the default reap time, but master tests are definitely taking a longer time (~8 hours) than 5, so I'd probably err on keeping it in for now.

PR #84, opened 2018-05-29
Skip database size query for filtered backups

khuddlefish:
Also moved call to initialize backup params string to be earlier. We
were trying to log the params before they were set.


khuddlefish:
The function I modified InitializeBackupReport has a bit of logic inside it. We don't have any tests for it right now. Do we want to add some? We would probably need to modify the function so we were passing in several variables.

PR #85, opened 2018-05-30
Use `make build` instead of `make build_linux` in packaging job

chrishajas:
`make build_linux` places the binary in the current directory instead of
~/go/bin/, which we reference during the packaging job.

### DIFF
--- a/ci/pipeline.yml
+++ b/ci/pipeline.yml
@@ -717,7 +717,7 @@ jobs:
           pushd $GOPATH/src/github.com/greenplum-db/gpbackup
           make depend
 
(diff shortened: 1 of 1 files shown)


soumyadeep2007:
We're ok with this approach. We could also just pull the binaries from the gpbackup directory directly now. We reference that path several times so making a GPBACKUP_DIR variable and using that would be equally easy. We'll leave it up to you.

Also I think gpbackup_s3_plugin didn't change the way it is installing so we don't actually need to change that, but maybe we want to make it consistent with how we're doing gpbackup.

PR #86, opened 2018-05-30
Drop plpgsql extension before tests

chrishajas:
With the 9.1 merge, the plpgsql extension is loaded into the database by
default. This interferes with tests expecting only extensions created
by the test to exist.


chrishajas:
@khuddlefish Yes. As of postgres 9.1, plpgsql is loaded through an extension (which in tern loads the language) instead of loading the language directly.

PR #87, opened 2018-05-30
Disable comment on resource queue tests temporarily

chrishajas:
This is currently not working in master. These tests will be enabled
once it is fixed.


chrishajas:
https://github.com/greenplum-db/gpdb/commit/ace7a3e9298f241a7c0aa2ebe5c5d2a4df399ea6 has been merged in the GPDB master branch, so this commit is no longer necessary.

PR #89, opened 2018-06-04
Add support for backing up Collations

chrishajas:

### DIFF
--- a/backup/predata_shared.go
+++ b/backup/predata_shared.go
@@ -488,3 +488,13 @@ func PrintCreateDependentTypeAndFunctionAndTablesStatements(metadataFile *utils.
 		}
 	}
 }
+
+func PrintCreateCollationStatements(metadataFile *utils.FileWithByteCount, toc *utils.TOC, collations []Collation, collationMetadata MetadataMap) {
+	for _, collation := range collations {
+		collationFQN := utils.MakeFQN(collation.Schema, collation.Name)
+		start := metadataFile.ByteCount
+		metadataFile.MustPrintf("\nCREATE COLLATION %s (LC_COLLATE = '%s', LC_CTYPE = '%s');", collationFQN, collation.Collate, collation.Ctype)
+		PrintObjectMetadata(metadataFile, collationMetadata[collation.Oid], collation.Name, "COLLATION")
+		toc.AddPredataEntry(collation.Schema, collation.Name, "COLLATION", "", start, metadataFile)
+	}
+}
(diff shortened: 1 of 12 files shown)


integration/predata_shared_queries_test.go
```diff
@@ -822,6 +837,26 @@ LANGUAGE SQL`)
 				resultMetadata := resultMetadataMap[oid]
 				structmatcher.ExpectStructsToMatch(&configurationMetadata, &resultMetadata)
 			})
+			It("returns a slice of default metadata for a collation in a specific schema", func() {
+				testutils.SkipIfBefore6(connection)
+				testhelper.AssertQueryRuns(connection, `CREATE COLLATION some_coll (lc_collate = 'POSIX', lc_ctype = 'POSIX');`)
+				defer testhelper.AssertQueryRuns(connection, "DROP COLLATION some_coll")
+				testhelper.AssertQueryRuns(connection, "CREATE SCHEMA testschema")
```
soumyadeep2007: Quick Q: Why do these "AssertQueryRuns" have to be defers? Can these not be asserted right after the call to the function under test?

chrishajas: The `defer` keyword ensures that the cleanup queries execute whether the test passes or fails. That way, if one test fails, other tests can still run without pollution. Go does not have a `finally`-esqe construct.

backup/predata_shared.go
```diff
@@ -488,3 +488,13 @@ func PrintCreateDependentTypeAndFunctionAndTablesStatements(metadataFile *utils.
 		}
 	}
 }
+
+func PrintCreateCollationStatements(metadataFile *utils.FileWithByteCount, toc *utils.TOC, collations []Collation, collationMetadata MetadataMap) {
```
khuddlefish: We should move this to predata_types.go

integration/predata_shared_queries_test.go
```diff
@@ -563,6 +563,21 @@ LANGUAGE SQL`)
 				resultMetadata := resultMetadataMap[oid]
 				structmatcher.ExpectStructsToMatchExcluding(&expectedMetadata, &resultMetadata, "Oid")
 			})
+			It("returns a slice of default metadata for a collation", func() {
```
khuddlefish: Move these to predata_types_queries_test.go

integration/predata_shared_queries_test.go
soumyadeep2007: Ohh! My bad, got confused thinking that the assert was related to what we were testing in the function.

backup/predata_shared.go
chrishajas: Whoops, meant to do that.

PR #90, opened 2018-06-05
Fully qualify error tables in CREATE EXTERNAL TABLE statements

chrishajas:
Previously, the error table was not fully qualified, causing gprestore to
attempt to put the table in pg_catalog. Now, we fully qualify the error
table. Additionally, we now properly quote the error table.

### DIFF
--- a/backup/queries_externals.go
+++ b/backup/queries_externals.go
@@ -26,7 +26,7 @@ SELECT
 	coalesce(rejectlimit, 0) AS rejectlimit,
 	coalesce(rejectlimittype, '') AS rejectlimittype,
 	coalesce(quote_ident(c.relname),'') AS errtablename,
-	coalesce((SELECT quote_ident(nspname) FROM pg_namespace WHERE oid = relnamespace), '') AS errtableschema,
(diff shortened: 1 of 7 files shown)


backup/queries_externals.go
```diff
@@ -25,10 +25,11 @@ SELECT
 	coalesce(command, '') AS command,
 	coalesce(rejectlimit, 0) AS rejectlimit,
 	coalesce(rejectlimittype, '') AS rejectlimittype,
-	coalesce((SELECT relname FROM pg_class WHERE oid = fmterrtbl), '') AS errtable,
+	coalesce(quote_ident(c.relname),'') AS errtablename,
+	coalesce((SELECT quote_ident(nspname) FROM pg_namespace WHERE oid = relnamespace), '') AS errtableschema,
```
khuddlefish: Would we be able to use some letters in this query to indicate which table the columns are coming from?
```
coalesce((SELECT quote_ident(n.nspname) FROM pg_namespace n WHERE e.oid = n.relnamespace), '') AS errtableschema,
```
for example? I'm not actually sure which table we are pulling oid from in this case.

chrishajas: Yeah, it's ambiguous. Updated.

PR #92, opened 2018-06-07
Make updates for cluster.Cluster changes

jmcatamney:
This commit updates source and test code to account for the removal of SegDirMap
from Cluster and NewCluster now returning a *Cluster instead of a Cluster.  The
changes are generally limited to initialization of FilePathInfo and Cluster
objects and the arguments of functions that take Clusters as arguments.

Authored-by: Jamie McAtamney <jmcatamney@pivotal.io>

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -19,7 +19,7 @@
     "structmatcher",
     "testhelper"
   ]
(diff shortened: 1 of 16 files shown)


khuddlefish:
Merged

PR #94, opened 2018-06-13
Support Resource Group MemoryAuditor and CPUSET

dujialun:
- gpbackup can backup resgroup with memoryauditor and cpuset
  property
- Be sure it can restore correctly on default resource groups, adjust some
  parameters to 1 first, which the total sum can't exceed 100.

### DIFF
--- a/backup/metadata_globals.go
+++ b/backup/metadata_globals.go
@@ -92,6 +92,33 @@ func PrintCreateResourceQueueStatements(metadataFile *utils.FileWithByteCount, t
 	}
 }
 
+type resGroupPrepareStruct struct {
+	oid     uint32
+	name    string
+	setting string
+}
+
+func PrintPrepareResourceGroupStatements(metadataFile *utils.FileWithByteCount, toc *utils.TOC, resGroupMetadata MetadataMap) {
+	/*
+	 * total cpu_rate_limit and memory_limit should less than 100, so clean
+	 * them before we seting new memory_limit and cpu_rate_limit.
+	 */
+	defSettings := []resGroupPrepareStruct{
+		{6438, "admin_group", "SET CPU_RATE_LIMIT 1"},
+		{6438, "admin_group", "SET MEMORY_LIMIT 1"},
+		{6437, "default_group", "SET CPU_RATE_LIMIT 1"},
+		{6437, "default_group", "SET MEMORY_LIMIT 1"},
+	}
+	for _, prepare := range defSettings {
+		start := uint64(0)
+
+		start = metadataFile.ByteCount
+		metadataFile.MustPrintf("\n\nALTER RESOURCE GROUP %s %s;", prepare.name, prepare.setting)
+		PrintObjectMetadata(metadataFile, resGroupMetadata[prepare.oid], prepare.name, "RESOURCE GROUP")
+		toc.AddGlobalEntry("", prepare.name, "RESOURCE GROUP", start, metadataFile)
+	}
+}
+
 type resGroupStruct struct {
 	setting string
 	value   int
@@ -103,7 +130,6 @@ func PrintCreateResourceGroupStatements(metadataFile *utils.FileWithByteCount, t
 
 		if resGroup.Name == "default_group" || resGroup.Name == "admin_group" {
 			resGroupList := []resGroupStruct{
-				{"CPU_RATE_LIMIT", resGroup.CPURateLimit},
 				{"MEMORY_LIMIT", resGroup.MemoryLimit},
 				{"MEMORY_SHARED_QUOTA", resGroup.MemorySharedQuota},
 				{"MEMORY_SPILL_RATIO", resGroup.MemorySpillRatio},
@@ -115,10 +141,44 @@ func PrintCreateResourceGroupStatements(metadataFile *utils.FileWithByteCount, t
 				PrintObjectMetadata(metadataFile, resGroupMetadata[resGroup.Oid], resGroup.Name, "RESOURCE GROUP")
 				toc.AddGlobalEntry("", resGroup.Name, "RESOURCE GROUP", start, metadataFile)
 			}
+
+			/* special handling for cpu properties */
+			start = metadataFile.ByteCount
+			if resGroup.CPURateLimit >= 0 {
+				/* cpu rate mode */
+				metadataFile.MustPrintf("\n\nALTER RESOURCE GROUP %s SET CPU_RATE_LIMIT %d;", resGroup.Name, resGroup.CPURateLimit)
+			} else {
+				/* cpuset mode */
+				metadataFile.MustPrintf("\n\nALTER RESOURCE GROUP %s SET CPUSET '%s';", resGroup.Name, resGroup.Cpuset)
+			}
+			PrintObjectMetadata(metadataFile, resGroupMetadata[resGroup.Oid], resGroup.Name, "RESOURCE GROUP")
+			toc.AddGlobalEntry("", resGroup.Name, "RESOURCE GROUP", start, metadataFile)
 		} else {
 			start = metadataFile.ByteCount
 			attributes := []string{}
-			attributes = append(attributes, fmt.Sprintf("CPU_RATE_LIMIT=%d", resGroup.CPURateLimit))
+
+			/* special handling for cpu properties */
+			if resGroup.CPURateLimit >= 0 {
+				/* cpu rate mode */
+				attributes = append(attributes, fmt.Sprintf("CPU_RATE_LIMIT=%d", resGroup.CPURateLimit))
+			} else {
+				/* cpuset mode */
+				attributes = append(attributes, fmt.Sprintf("CPUSET='%s'", resGroup.Cpuset))
+			}
+
+			/*
+			 * Possible values of memory_auditor:
+			 * - "1": cgroup;
+			 * - "0": vmtracker;
+			 * - "": not set, e.g. created on an older version which does not
+			 *   support memory_auditor yet, consider it as vmtracker;
+			 */
+			if resGroup.MemoryAuditor == 1 {
+				attributes = append(attributes, fmt.Sprintf("MEMORY_AUDITOR=cgroup"))
+			} else {
+				attributes = append(attributes, fmt.Sprintf("MEMORY_AUDITOR=vmtracker"))
+			}
+
 			attributes = append(attributes, fmt.Sprintf("MEMORY_LIMIT=%d", resGroup.MemoryLimit))
 			attributes = append(attributes, fmt.Sprintf("MEMORY_SHARED_QUOTA=%d", resGroup.MemorySharedQuota))
 			attributes = append(attributes, fmt.Sprintf("MEMORY_SPILL_RATIO=%d", resGroup.MemorySpillRatio))
--- a/backup/metadata_globals.go
+++ b/backup/metadata_globals.go
@@ -92,29 +92,27 @@ func PrintCreateResourceQueueStatements(metadataFile *utils.FileWithByteCount, t
 	}
 }
 
-type resGroupPrepareStruct struct {
-	oid     uint32
-	name    string
-	setting string
-}
-
-func PrintPrepareResourceGroupStatements(metadataFile *utils.FileWithByteCount, toc *utils.TOC, resGroupMetadata MetadataMap) {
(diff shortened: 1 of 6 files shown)


backup/metadata_globals.go
```diff
@@ -92,6 +92,33 @@ func PrintCreateResourceQueueStatements(metadataFile *utils.FileWithByteCount, t
...
+}
+
+func PrintPrepareResourceGroupStatements(metadataFile *utils.FileWithByteCount, toc *utils.TOC, resGroupMetadata MetadataMap) {
+	/*
+	 * total cpu_rate_limit and memory_limit should less than 100, so clean
+	 * them before we seting new memory_limit and cpu_rate_limit.
+	 */
+	defSettings := []resGroupPrepareStruct{
+		{6438, "admin_group", "SET CPU_RATE_LIMIT 1"},
+		{6438, "admin_group", "SET MEMORY_LIMIT 1"},
```
soumyadeep2007: It would help readability if 6437 and 6438 are declared as constants before their use

backup/metadata_globals.go
```diff
@@ -92,6 +92,33 @@ func PrintCreateResourceQueueStatements(metadataFile *utils.FileWithByteCount, t
...
 
+type resGroupPrepareStruct struct {
+	oid     uint32
+	name    string
+	setting string
+}
+
+func PrintPrepareResourceGroupStatements(metadataFile *utils.FileWithByteCount, toc *utils.TOC, resGroupMetadata MetadataMap) {
+	/*
+	 * total cpu_rate_limit and memory_limit should less than 100, so clean
```
soumyadeep2007: Where is the cleanup being done?

soumyadeep2007: If we need to add this, can we put the set statements inside `PrintCreateResourceGroupStatements`. We also don't need to run PrintObjectMetadata in here.

backup/metadata_globals.go
```diff
@@ -115,10 +141,44 @@ func PrintCreateResourceGroupStatements(metadataFile *utils.FileWithByteCount, t
 				PrintObjectMetadata(metadataFile, resGroupMetadata[resGroup.Oid], resGroup.Name, "RESOURCE GROUP")
 				toc.AddGlobalEntry("", resGroup.Name, "RESOURCE GROUP", start, metadataFile)
 			}
+
+			/* special handling for cpu properties */
+			start = metadataFile.ByteCount
+			if resGroup.CPURateLimit >= 0 {
+				/* cpu rate mode */
+				metadataFile.MustPrintf("\n\nALTER RESOURCE GROUP %s SET CPU_RATE_LIMIT %d;", resGroup.Name, resGroup.CPURateLimit)
+			} else {
```
soumyadeep2007: Is there a default value of cpuset? If the cpuset is equal to its default values, we should not print the ALTER statement, since if a user needed to restore back to a previous GPDB5 version before this was introduced, restore would fail.

backup/metadata_globals.go
```diff
@@ -115,10 +141,44 @@ func PrintCreateResourceGroupStatements(metadataFile *utils.FileWithByteCount, t
...
 		} else {
 			start = metadataFile.ByteCount
 			attributes := []string{}
-			attributes = append(attributes, fmt.Sprintf("CPU_RATE_LIMIT=%d", resGroup.CPURateLimit))
+
+			/* special handling for cpu properties */
+			if resGroup.CPURateLimit >= 0 {
+				/* cpu rate mode */
+				attributes = append(attributes, fmt.Sprintf("CPU_RATE_LIMIT=%d", resGroup.CPURateLimit))
+			} else {
```
soumyadeep2007: Same concern as above, does a default value of cpuset exist that we can check against?

backup/metadata_globals.go
dujialun: The sum of cpu_rate_limit & memory_limit will not exceed 100, suppose a sql generated by gpbackup is:

```sql
ALTER RESOURCE GROUP admin_group SET CPU_RATE_LIMIT 80;
ALTER RESOURCE GROUP default_group SET CPU_RATE_LIMIT 10;
```

And suppose the resgroup has default status as below when user wants to restore:

- admin_group CPU_RATE_LIMIT=10;
- default_group CPU_RATE_LIMIT=30;

Then the restore will cause an error first sql as the total cpu_rate_limit exceeds 100.

That's why we need some preparation to alter default groups' cpu_rate_limit & memory_limit to low enough values.

backup/metadata_globals.go
```diff
@@ -92,6 +92,33 @@ func PrintCreateResourceQueueStatements(metadataFile *utils.FileWithByteCount, t
...
+	setting string
+}
+
+func PrintPrepareResourceGroupStatements(metadataFile *utils.FileWithByteCount, toc *utils.TOC, resGroupMetadata MetadataMap) {
+	/*
+	 * total cpu_rate_limit and memory_limit should less than 100, so clean
+	 * them before we seting new memory_limit and cpu_rate_limit.
+	 */
+	defSettings := []resGroupPrepareStruct{
+		{6438, "admin_group", "SET CPU_RATE_LIMIT 1"},
```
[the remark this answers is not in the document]
dujialun: 6437 and 6438 are OIDs for default_group and admin_group, they are fixed in catalog, will never change at runtime, will never change across different gpdb versions. Both the OID and the resgroup name (like 6438 and "admin_group") are hard coded constants.

backup/metadata_globals.go
```diff
@@ -115,10 +141,44 @@ func PrintCreateResourceGroupStatements(metadataFile *utils.FileWithByteCount, t
 				PrintObjectMetadata(metadataFile, resGroupMetadata[resGroup.Oid], resGroup.Name, "RESOURCE GROUP")
 				toc.AddGlobalEntry("", resGroup.Name, "RESOURCE GROUP", start, metadataFile)
 			}
+
+			/* special handling for cpu properties */
```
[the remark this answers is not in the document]
dujialun: make sense, will change

backup/metadata_globals.go
dujialun: a resgroup can have either cpu_rate_limit or cpuset, but not both. When cpu_rate_limit is specified its value must be >= 0, otherwise its value is -1 and cpuset is in use. ALTER a resgroup to CPU_RATE_LIMIT mode will reset the cpuset property, ALTER a resgroup to CPUSET will reset the cpu_rate_limit property.

cpuset is not supported until gpdb 5.9, so if a cpuset resgroup is dumped via gpbackup then it's not possible to restore it on gpdb 5.0, e.g..

backup/queries_globals.go
```diff
@@ -127,28 +127,36 @@ type ResourceGroup struct {
...
+	Cpuset            string
 }
 
 func GetResourceGroups(connection *dbconn.DBConn) []ResourceGroup {
 	query := `
 SELECT g.oid,
 	quote_ident(g.rsgname) AS name,
 	t1.proposed AS concurrency,
-	t2.proposed AS cpuratelimit,
+	t2.value    AS cpuratelimit,
```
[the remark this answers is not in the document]
dujialun: proposed is only meaningful for memory related properties, like memory_limit and concurrency, for other properties proposed==value, and it's recommended to use value for them.

dujialun:
@soumyadeep2007 Integration test has been added.

backup/metadata_globals.go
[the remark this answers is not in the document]
dujialun: @soumyadeep2007 When we refractor the code, we realize that cpuset is string, not int, so we can't reuse the struct. Not all the parameters are integers now.

backup/metadata_globals.go
soumyadeep2007: I see. Let's rename the function to PrintResetResourceGroupStatements. We don't need the PrintObjectMetadata call in this function (and therefore we don't need the oid in the struct).

soumyadeep2007: We could make resGroupPrepareStruct anonymous since it is not used elsewhere.

pivotal-ning-yu: Thanks for the suggestion. If we do not need the metadata call then I think the struct is neither needed, we could simply print 4 alter commands in this function.

backup/metadata_globals.go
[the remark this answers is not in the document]
pivotal-ning-yu: make sense, we'll make the change.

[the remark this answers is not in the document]
dujialun: @soumyadeep2007 `resGroup` is also a variable name in function `PrintCreateResourceGroupStatements()`, to prevent conflict we didn't rename the struct `resGroupStruct` to `resGroup` in this PR.

backup/metadata_globals.go
soumyadeep2007: We could but I think its cleaner and shorter with the anonymous struct.

backup/metadata_globals.go
```diff
@@ -92,6 +92,31 @@ func PrintCreateResourceQueueStatements(metadataFile *utils.FileWithByteCount, t
 	}
 }
 
+func PrintResetResourceGroupStatements(metadataFile *utils.FileWithByteCount, toc *utils.TOC, resGroupMetadata MetadataMap) {
+	/*
+	 * total cpu_rate_limit and memory_limit should less than 100, so clean
+	 * them before we seting new memory_limit and cpu_rate_limit.
+	 */
+	defSettings := []struct {
+		oid     uint32
```
khuddlefish: It looks like this oid isn't actually being used.

backup/metadata_globals.go
```diff
@@ -115,10 +139,44 @@ func PrintCreateResourceGroupStatements(metadataFile *utils.FileWithByteCount, t
...
+			}
+
+			/*
+			 * Possible values of memory_auditor:
+			 * - "1": cgroup;
+			 * - "0": vmtracker;
+			 * - "": not set, e.g. created on an older version which does not
+			 *   support memory_auditor yet, consider it as vmtracker;
+			 */
+			if resGroup.MemoryAuditor == 1 {
```
khuddlefish: Do we need to worry about MEMORY_AUDITOR for the other case above?

backup/metadata_globals.go
```diff
@@ -92,6 +92,31 @@ func PrintCreateResourceQueueStatements(metadataFile *utils.FileWithByteCount, t
 	}
 }
 
+func PrintResetResourceGroupStatements(metadataFile *utils.FileWithByteCount, toc *utils.TOC, resGroupMetadata MetadataMap) {
```
khuddlefish: I don't think we should need this function. We should be setting both of these properties to valid values later so we shouldn't need to reset them to 1 first.

chrishajas: There's a comment thread above where we had the same concern, but we actually do need this due to restrictions on how they're set for the default resource groups.

khuddlefish: I see. Ok, we can leave it then

backup/metadata_globals.go
dujialun: @khuddlefish I can't catch you, can you explain what you worry about?

khuddlefish: I don't believe we are doing anything to set MEMORY_AUDITIOR for admin_group or default_group. Can this property be set on those groups as well?

backup/metadata_globals.go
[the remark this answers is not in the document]
khuddlefish: @dujialun I think there are some unresolved change requests in this section. Otherwise, this is looking pretty good :)

backup/metadata_globals.go
dujialun: @khuddlefish `MEMORY_AUDITOR` has some limitation in use, it can not be altered, so default resgroups `admin_group` and `default_group` always has NULL or 0 for this property. For newly created resgroups unless we explicitly set it to `cgroup` (internally 1) it always has the value NULL or 0.

backup/metadata_globals.go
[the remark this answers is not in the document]
dujialun: A commit was pushed to make the `resGroup` struct anonymous. We didn't change the value type to string because even by doing so we still have to make special handling on cpu properties and memory_auditor.

Looks like there is no other unresolved change requests, could you please review again and check whether it needs further improvements?

soumyadeep2007:
Looking good! Thanks a lot!

dujialun:
@soumyadeep2007 Please squash and merge, we don't have the privilege.

PR #95, opened 2018-06-22
Properly handle objects created by extensions

chrishajas:
Objects that are created by extensions should not be dumped explicitly since they are created by a script when CREATE EXTENSION is invoked (this statement is dumped). While we had most objects covered, there were a few objects that we missed (functions, types, constraints, postdata).

Additionally, we drop the hyperloglog extension so it does not pollute our tests.

### DIFF
--- a/backup/queries_relations.go
+++ b/backup/queries_relations.go
@@ -76,8 +76,8 @@ JOIN pg_namespace n
 WHERE %s
 %s
 AND relkind = 'r'
-AND c.oid NOT IN (select objid from pg_depend where deptype = 'e')
-ORDER BY c.oid;`, relationAndSchemaFilterClause(), childPartitionFilter)
+AND %s
+ORDER BY c.oid;`, relationAndSchemaFilterClause(), childPartitionFilter, ExtensionFilterClause("c"))
 
 	results := make([]Relation, 0)
 	err := connection.Select(&results, query)
@@ -151,8 +151,8 @@ AND (
 	%s
 )
 AND relkind = 'r'
-AND c.oid NOT IN (select objid from pg_depend where deptype = 'e')
-ORDER BY c.oid;`, SchemaFilterClause("n"), oidStr, oidStr, oidStr, childPartitionFilter)
+AND %s
+ORDER BY c.oid;`, SchemaFilterClause("n"), oidStr, oidStr, oidStr, childPartitionFilter, ExtensionFilterClause("c"))
 
 	results := make([]Relation, 0)
 	err := connection.Select(&results, query)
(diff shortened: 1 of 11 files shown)


backup/queries_relations.go
```diff
@@ -634,8 +634,8 @@ WHERE d.classid = 'pg_rewrite'::regclass::oid
 	AND v1.oid != v2.oid
 	AND v1.relkind = 'v'
 	AND %s
-	AND v1.oid NOT IN (select objid from pg_depend where deptype = 'e')
-ORDER BY v2.oid, referencedobject;`, SchemaFilterClause("n"))
+	AND %s
+ORDER BY v2.oid, referencedobject;`, SchemaFilterClause("n"), ExtensionFilterClause("v1"))
```
soumyadeep2007: Do we need one for v2 as well?

backup/queries_shared.go
```diff
@@ -41,8 +41,8 @@ SELECT
 	quote_ident(nspname) AS name
 FROM pg_namespace n
 WHERE %s
-AND oid NOT IN (select objid from pg_depend where deptype = 'e')
-ORDER BY name;`, SchemaFilterClause("n"))
+AND %s
+ORDER BY name;`, SchemaFilterClause("n"), ExtensionFilterClause(""))
```
soumyadeep2007: It might be good to use n.oid and n.nspname in the select part for consistency. Then we could use "n" in the extension clause.

integration/integration_suite_test.go
```diff
@@ -60,6 +60,7 @@ var _ = BeforeSuite(func() {
 	if connection.Version.AtLeast("6") {
 		// Drop plpgsql extension to not interfere in extension tests
 		testhelper.AssertQueryRuns(connection, "DROP EXTENSION plpgsql CASCADE")
+		testhelper.AssertQueryRuns(connection, "DROP EXTENSION hyperloglog_counter")
```
soumyadeep2007: We should be careful with this. Just dropping the extension before would have fixed all of our other tests and we might not have found the extension bug. Maybe we want to do the extension dropping stuff only for the extension tests so that we are not indirectly affecting other tests.

backup/queries_relations.go
chrishajas: Actually, I'm not even sure if we need it at all here since the list of views will have already been filtered through this.

backup/queries_shared.go
chrishajas: Yep, I agree.

integration/integration_suite_test.go
chrishajas: It would have failed on end-to-end since the bug was during backup of an empty database (and it would likely fail on restore too since the object would end up being created twice). I'd rather be explicit about what items are in the database during these tests, so if we want to explicitly test that objects aren't part of an extension, we can add a test to every object type and assert that it isn't picked up by the query. The hyperloglog extension has some coverage, but it's by no means exhaustive. The other option would be to create a "test extension" that contains every type of object, but we'd have to make sure we update that as we add additional objects. Currently, we have little/no coverage of objects belonging to an extension. I think having a test extension would be the easiest/we wouldn't have to add specific new tests for every object, but it would be testing that no extra objects are added to the dump as a side effect of checking that the length of results == 1. What do you think?

khuddlefish: I think we should be running our tests on a database environment as similar to our customer's environments as we can get. It's not that I want to have some other random objects in the database, I mostly want to have the same objects in the database that we can actually expect to be there. I feel uneasy about dropping objects that are in the database by default because those objects will be in our customer's databases and we are testing on a different setup. The only tests either of those extensions should affect are the actual extension tests since all other object queries should be filtering out objects that are dependent on extensions. I would rather keep the objects in the database and make adjustments for the extension tests themselves than change the database for all other tests.

chrishajas: @khuddlefish Changed. Good to push?

PR #96, opened 2018-06-25
Fix backup of external tables with --single-data-file

chrishajas:
During the gpbackup --single-data-file refactor to use an agent
approach, the oid list created for the master included external tables,
while the oid list on the segments did not. This caused the backup to
fail as the pipe names did not match (the segment created a pipe with
the name of a regular table, whereas the master executed a COPY TO
command to a pipe with an external table). This only occurred when
--single-data-file was used.

### DIFF
--- a/end_to_end/end_to_end_suite_test.go
+++ b/end_to_end/end_to_end_suite_test.go
@@ -351,6 +351,22 @@ var _ = Describe("backup end to end integration tests", func() {
 
 				os.RemoveAll(backupdir)
 			})
+			It("runs gpbackup and gprestore on database with all objects", func() {
+				testhelper.AssertQueryRuns(backupConn, "DROP SCHEMA IF EXISTS schema2 CASCADE; DROP SCHEMA public CASCADE; CREATE SCHEMA public; DROP PROCEDURAL LANGUAGE IF EXISTS plpythonu;")
+				defer testutils.ExecuteSQLFile(backupConn, "test_tables_data.sql")
+				defer testutils.ExecuteSQLFile(backupConn, "test_tables_ddl.sql")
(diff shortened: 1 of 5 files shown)


end_to_end/end_to_end_suite_test.go
```diff
@@ -456,22 +472,6 @@ var _ = Describe("backup end to end integration tests", func() {
 
 			os.RemoveAll(backupdir)
 		})
-		It("runs gpbackup and gprestore on database with all objects", func() {
```
khuddlefish: Do we not have one of these that runs without single-data-file now?

chrishajas: Correct. We still have tests that run without `--single-data-file`, but the one containing all objects is now in `--single-data-file`. The metadata portion is identical, and really that's the bulk of what's being covered in this test. However, the single-data-file has some corner cases that will be covered better here. The other option is to do this test in both places, but it's ~20 seconds.

end_to_end/end_to_end_suite_test.go
```diff
@@ -351,6 +351,22 @@ var _ = Describe("backup end to end integration tests", func() {
...
+				testutils.ExecuteSQLFile(backupConn, "gpdb4_objects.sql")
+				if backupConn.Version.AtLeast("5") {
+					testutils.ExecuteSQLFile(backupConn, "gpdb5_objects.sql")
+				}
+				if backupConn.Version.AtLeast("6") {
+					testutils.ExecuteSQLFile(backupConn, "gpdb6_objects.sql")
+				}
+				timestamp := gpbackup(gpbackupPath, "--leaf-partition-data", "--single-data-file")
+				gprestore(gprestorePath, timestamp, "--redirect-db", "restoredb")
+
```
soumyadeep2007: There is no assert here?

khuddlefish: The gpbackup and gprestore commands will fail the test if there is some problem. It is an implied assert that they run successfully.

PR #97, opened 2018-07-02
Restore from restore plans for both backup types

soumyadeep2007:
Iterate through timestamps found in restore plan of specified backup
and restore table data corresponding to each timestamp.
Also, ensure that if a restore is done on a legacy backup (having a
missing restore plan), a default restore plan is supplied. Here we
have assumed that such a legacy backup can only be a full backup.
Thus, the default restore plan would always be a single entry with
a list of all tables backed up in the full backup.

Co-authored-by: Soumyadeep Chakraborty <sochakraborty@pivotal.io>

### DIFF
--- a/end_to_end/end_to_end_suite_test.go
+++ b/end_to_end/end_to_end_suite_test.go
@@ -145,12 +145,24 @@ var _ = Describe("backup end to end integration tests", func() {
 	})
 
 	Describe("end to end gpbackup and gprestore tests", func() {
-		publicSchemaTupleCounts := map[string]int{"public.foo": 40000, "public.holds": 50000, "public.sales": 13}
-		schema2TupleCounts := map[string]int{"schema2.returns": 6, "schema2.foo2": 0, "schema2.foo3": 100}
+		var publicSchemaTupleCounts, schema2TupleCounts map[string]int
+
 		BeforeEach(func() {
 			testhelper.AssertQueryRuns(restoreConn, "DROP SCHEMA IF EXISTS schema2 CASCADE; DROP SCHEMA public CASCADE; CREATE SCHEMA public;")
+			publicSchemaTupleCounts = map[string]int{
+				"public.foo":   40000,
+				"public.holds": 50000,
+				"public.sales": 13,
+			}
+			schema2TupleCounts = map[string]int{
+				"schema2.returns": 6,
+				"schema2.foo2":    0,
+				"schema2.foo3":    100,
+				"schema2.ao1":     1000,
+				"schema2.ao2":     1000,
+			}
 		})
-		Context("Backup include filtering", func() {
+		Describe("Backup include filtering", func() {
 			It("runs gpbackup and gprestore with include-schema backup flag and compression level", func() {
 				timestamp := gpbackup(gpbackupPath, "--include-schema", "public", "--compression-level", "2")
 				gprestore(gprestorePath, timestamp, "--redirect-db", "restoredb")
@@ -180,13 +192,13 @@ var _ = Describe("backup end to end integration tests", func() {
 			})
 
 		})
-		Context("Restore include filtering", func() {
+		Describe("Restore include filtering", func() {
 			It("runs gpbackup and gprestore with include-schema restore flag", func() {
 				backupdir := "/tmp/include_schema"
 				timestamp := gpbackup(gpbackupPath, "--backup-dir", backupdir)
 				gprestore(gprestorePath, timestamp, "--redirect-db", "restoredb", "--backup-dir", backupdir, "--include-schema", "schema2")
 
-				assertRelationsCreated(restoreConn, 15)
+				assertRelationsCreated(restoreConn, 17)
 				assertDataRestored(restoreConn, schema2TupleCounts)
 
 				os.RemoveAll(backupdir)
@@ -214,19 +226,19 @@ var _ = Describe("backup end to end integration tests", func() {
 				os.Remove("/tmp/include-tables.txt")
 			})
 		})
-		Context("Backup exclude filtering", func() {
+		Describe("Backup exclude filtering", func() {
 			It("runs gpbackup and gprestore with exclude-schema backup flag", func() {
 				timestamp := gpbackup(gpbackupPath, "--exclude-schema", "public")
 				gprestore(gprestorePath, timestamp, "--redirect-db", "restoredb")
 
-				assertRelationsCreated(restoreConn, 15)
+				assertRelationsCreated(restoreConn, 17)
 				assertDataRestored(restoreConn, schema2TupleCounts)
 			})
 			It("runs gpbackup and gprestore with exclude-table backup flag", func() {
 				timestamp := gpbackup(gpbackupPath, "--exclude-table", "schema2.foo2", "--exclude-table", "schema2.returns", "--exclude-table", "public.myseq1", "--exclude-table", "public.myview1")
 				gprestore(gprestorePath, timestamp, "--redirect-db", "restoredb")
 
-				assertRelationsCreated(restoreConn, 18)
+				assertRelationsCreated(restoreConn, 20)
 				assertDataRestored(restoreConn, map[string]int{"schema2.foo3": 100, "public.foo": 40000, "public.holds": 50000, "public.sales": 13})
 
 				os.Remove("/tmp/exclude-tables.txt")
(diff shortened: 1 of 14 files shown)


restore/restore.go
```diff
@@ -186,29 +193,39 @@ func restorePredata(metadataFilename string) {
...
-func restoreData(gucStatements []utils.StatementWithType) {
+func restoreData(fpInfoList []utils.FilePathInfo, gucStatements []utils.StatementWithType) {
 	if wasTerminated {
 		return
 	}
-	gplog.Info("Restoring data")
-	filteredMasterDataEntries := globalTOC.GetDataEntriesMatching(*includeSchemas, *excludeSchemas, *includeRelations, *excludeRelations)
+	latestRestorePlan := backupConfig.RestorePlan
+	for i, fpInfo := range fpInfoList {
+		gplog.Info("Restoring data from backup with timestamp: %s	", fpInfo.Timestamp)
```
jmcatamney: Any particular reason for the extra whitespace here?

restore/restore.go
```diff
@@ -186,29 +193,39 @@ func restorePredata(metadataFilename string) {
...
 }
 
-func restoreData(gucStatements []utils.StatementWithType) {
+func restoreData(fpInfoList []utils.FilePathInfo, gucStatements []utils.StatementWithType) {
 	if wasTerminated {
 		return
 	}
-	gplog.Info("Restoring data")
-	filteredMasterDataEntries := globalTOC.GetDataEntriesMatching(*includeSchemas, *excludeSchemas, *includeRelations, *excludeRelations)
+	latestRestorePlan := backupConfig.RestorePlan
```
jmcatamney: This might make more sense as just `restorePlan`, since it's not comparing multiple restore plans.

restore/restore.go
soumyadeep2007: Typo! Thanks!

restore/restore.go
soumyadeep2007: We have multiple restore plans (one for each backup). At first glance it may not be apparent that the global config points to the latest config and by extension, the latest restore plan. That is why it is explicitly called out.

end_to_end/end_to_end_suite_test.go
```diff
@@ -145,12 +145,24 @@ var _ = Describe("backup end to end integration tests", func() {
...
+			schema2TupleCounts = map[string]int{
+				"schema2.returns": 6,
+				"schema2.foo2":    0,
+				"schema2.foo3":    100,
+				"schema2.ao1":     50000,
+				"schema2.ao2":     50000,
+			}
 		})
-		Context("Backup include filtering", func() {
+		Describe("Backup include filtering", func() {
```
khuddlefish: What's the difference between Describe and Context for this purpose?

end_to_end/end_to_end_suite_test.go
```diff
@@ -368,6 +380,32 @@ var _ = Describe("backup end to end integration tests", func() {
 
 			})
 		})
+		Describe("Incremental", func() {
+			It("restores from a full backup", func() {
```
khuddlefish: This should already be encompassed in some other case. These tests take long enough that I would rather not add an unnecessary repeat case.

restore/restore.go
khuddlefish: I'm good with latestRestorePlan. It is a bit more clear

end_to_end/end_to_end_suite_test.go
soumyadeep2007: Context is usually used to provide documentation about a particular scenario whereas Describe describes high level functionality that we want to test. In a unit test, Describe is typically a function, whereas here in end to end, it is a piece of higher level functionality.

That's why I replaced the context blocks.

end_to_end/end_to_end_suite_test.go
soumyadeep2007: Yeah, I will take that out.

utils/report.go
```diff
@@ -42,6 +42,23 @@ type RestorePlanEntry struct {
 	TableFQNs []string
 }
 
+func EnsureDefaultRestorePlan(fpInfo *FilePathInfo, backupConfig *BackupConfig) {
```
chrishajas: We should add a test for this function.

restore/wrappers.go
```diff
@@ -62,6 +62,8 @@ func InitializeBackupConfig() {
 	utils.InitializeCompressionParameters(backupConfig.Compressed, 0)
 	utils.EnsureBackupVersionCompatibility(backupConfig.BackupVersion, version)
 	utils.EnsureDatabaseVersionCompatibility(backupConfig.DatabaseVersion, connectionPool.Version)
+
+	utils.EnsureDefaultRestorePlan(&globalFPInfo, backupConfig)
```
chrishajas: The above Ensure functions will exit the backup if the backup or database versions are incompatible, which differs from the purpose of `EnsureDefaultRestorePlan`. Let's name this function something more descriptive, such as `SetRestorePlanForLegacyBackup`.

utils/toc_test.go
```diff
@@ -218,44 +218,62 @@ var _ = Describe("utils/toc tests", func() {
 			Expect(statements).To(Equal([]utils.StatementWithType{}))
 		})
 	})
-	Context("GetDataEntriesMatching", func() {
-		It("returns matching entry on include schema", func() {
-			includeSchemas := []string{"schema1"}
+	Describe("GetDataEntriesMatching", func() {
```
chrishajas: Let's add tests for the restore plan (if the table matches the filter, but isn't in the restore plan for example, which would test a filtered restore case of incremental).

end_to_end/test_tables_ddl.sql
```diff
@@ -98,22 +98,30 @@ CREATE TABLE returns (
...
           PARTITION dec17 START ('2017-12-01'::date) END ('2018-01-01'::date) WITH (tablename='returns_1_prt_dec17', appendonly=false )
           );
 
+CREATE TABLE ao1 (
+    i integer
+) WITH (appendonly=true);
+
+CREATE TABLE ao2 (
+    i integer
+) WITH (appendonly=true);
```
chrishajas: Let's make one of these column oriented too. There's some logic in GPDB core (and possibly COPY) that differs between them.

end_to_end/test_tables_data.sql
```diff
@@ -50,6 +50,10 @@ SET search_path = schema2, pg_catalog;
 
 INSERT INTO foo3 SELECT generate_series(201,300);
 
+INSERT INTO ao1 SELECT generate_series(1,50000);
+
+INSERT INTO ao2 SELECT generate_series(1,50000);
```
chrishajas: We already have a couple tables with 50k rows, so we can reduce this to like 1000 rows to make the tests a bit faster.

end_to_end/end_to_end_suite_test.go
```diff
@@ -368,6 +380,23 @@ var _ = Describe("backup end to end integration tests", func() {
...
 			})
 		})
+		Describe("Incremental", func() {
+			It("restores from an incremental backup", func() {
+				fullBackupTimestamp := gpbackup(gpbackupPath)
+
+				testhelper.AssertQueryRuns(backupConn, "INSERT into schema2.ao1 values(50001)")
+				defer testhelper.AssertQueryRuns(backupConn, "DELETE from schema2.ao1 where i=50001")
+
+				incrementalTimestamp := gpbackup(gpbackupPath, "--incremental", fullBackupTimestamp)
```
chrishajas: I think it would be good to have one more incremental in here, to ensure we handle incrementing on an incremental backup.

utils/report.go
soumyadeep2007: Cool, will add!

restore/wrappers.go
soumyadeep2007: Makes sense.

utils/toc_test.go
soumyadeep2007: The tests do test that. We have schema3.table3 which is not in the restore plan. For example in test : "returns matching entry on exclude table" schema3.table3 does match the filter but it is not in the restore plan.

end_to_end/test_tables_ddl.sql
soumyadeep2007: Sure!

end_to_end/end_to_end_suite_test.go
soumyadeep2007: Sure!

end_to_end/test_tables_data.sql
soumyadeep2007: Sure!

soumyadeep2007:
@chrishajas Have a look. I added the tests, corrected the issue with filtering (double the no of tuples getting populated) and also refactored SetRestorePlanForLegacyBackup.

utils/toc.go
```diff
@@ -176,6 +176,7 @@ func (toc *TOC) GetDataEntriesMatching(includeSchemas []string, excludeSchemas [
 	}
 
 	restorePlanTableSet := NewIncludeSet(restorePlanTableFQNs)
+	restorePlanTableSet.AlwaysMatchesFilter = false
```
chrishajas: Also would be good to have a unit test that catches this.

soumyadeep2007: Done.

chrishajas: Just want to confirm that this was added. Once it's been added, looks good to merge it to the incremental branch!

soumyadeep2007: Yup done!

PR #98, opened 2018-07-05
Plugin revamp

soumyadeep2007:
1) Additional hooks for plugin lifecycle methods:
Now plugin lifecycle methods (setup and cleanup) are invoked once each
in the following scopes:
- Master
- Segment Host
- Segment
Also, refactored lifecycle methods : plugin.go

2) Fix: Cleanup plugin now executes for restore
cleanup_plugin_for_restore function was not executing as the
pluginConfig global was nil. It was not getting set as it was being
shadowed inside RecoverMetadataFilesUsingPlugin.

3) Logging plugin method invocations in example_plugin.sh for ease of verification.

### DIFF
--- a/plugins/README.md
+++ b/plugins/README.md
@@ -72,17 +72,23 @@ Our utility calls all commands listed below. Errors will occur if any of them ar
 
 These arguments are passed to the plugin by gpbackup/gprestore.
 
-<a name="config_path">**config_path:**</a> Absolute path to the config yaml file
+[config_path](#config_path): Absolute path to the config yaml file
 
-<a name="local_backup_directory">**local_backup_directory:**</a> The path to the directory where gpbackup would place backup files on the master host if not using a plugin. Our plugins reference this path to recreate a similar directory structure on the destination system. gprestore will read files from this location so the plugin will need to create the directory during setup if it does not already exist.
+[local_backup_directory](#local_backup_directory): The path to the directory where gpbackup would place backup files on the master host if not using a plugin. Our plugins reference this path to recreate a similar directory structure on the destination system. gprestore will read files from this location so the plugin will need to create the directory during setup if it does not already exist.
 
-<a name="filepath">**filepath:**</a> The local path to a file written by gpbackup and/or read by gprestore.
+[scope](#scope): The scope at which this plugin's setup/cleanup hook is invoked. Values for this parameter are: "master", "segment_host" and "segment". Each such hook is invoked at each of these scopes. For eg. If we have a cluster with a master on 1 master host and 2 segment hosts each with 4 segments, each of these hooks will be executed in the following manner: There will be 1 invocation
+of each method with the parameter "master", offering a chance to perform some setup/cleanup to be done *once* per cluster. Creation/Deletion of a remote directory is a perfect candidate here. Furthermore, there will be 1 invocation for each of these commands for each of the segment hosts, offering a chance to establish/teardown connectivity to a remote storage provider such as S3 for instance. Finally, there will be 1 invocation for each of these commands for each of the segments.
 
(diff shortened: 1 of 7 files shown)


plugins/README.md
```diff
@@ -96,6 +99,8 @@ Called at the start of the backup process on the master and each segment host.
 
 [local_backup_directory](#local_backup_directory)
 
+[scope](#scope)
+
```
chrishajas: Update the examples with the scope.

plugins/README.md
```diff
@@ -76,6 +76,9 @@ These arguments are passed to the plugin by gpbackup/gprestore.
 
 <a name="local_backup_directory">**local_backup_directory:**</a> The path to the directory where gpbackup would place backup files on the master host if not using a plugin. Our plugins reference this path to recreate a similar directory structure on the destination system. gprestore will read files from this location so the plugin will need to create the directory during setup if it does not already exist.
 
+<a name="scope">**scope**</a> The scope at which this plugin lifecycle method is invoked. Values for this parameter are:
+"master", "segment_host" and "segment".
```
chrishajas: Let's give a bit more description about the difference between master, segment_host, and segment (i.e. make clear the difference between segment_host and segment)

plugins/README.md
soumyadeep2007: Done!

plugins/README.md
soumyadeep2007: Done!

plugins/README.md
chrishajas: I think we need to somehow get across that that each of these setup/cleanup functions are called once with master, segment, and segment_host. Might also be good to provide examples of what a developer might put in each (for example, host would be checking connectivity, master would be creating a remote directory or anything that needs to be done only once)

chrishajas:
Looks good! We'll wait to merge this until we've completed the associated changes for the s3 and ddboost plugin.

utils/plugin.go
```diff
@@ -81,42 +89,83 @@ func (plugin *PluginConfig) CheckPluginExistsOnAllHosts(c *cluster.Cluster) {
...
+	// Execute command once for each segment
+	scope = Segment
+	commandFunc = plugin.buildLifecycleCommandFunc(command, configPath, backupDir, scope)
+	verboseErrorMsg, errorMsgFunc = plugin.buildLifecycleErrorMsgAndFunc(command, scope)
+	verboseCommandSegMsg := fmt.Sprintf(verboseCommandMsg, "segments")
+	remoteOutput = c.GenerateAndExecuteCommand(verboseCommandSegMsg, commandFunc, cluster.ON_SEGMENTS)
+	c.CheckClusterError(remoteOutput, verboseErrorMsg, errorMsgFunc, noFatal)
+}
+
+func (plugin *PluginConfig) buildLifecycleCommandFunc(command, configPath, backupDir string,
```
khuddlefish: Small thing, but I think we usually have used the word "construct" rather than "build" for these sorts of functions. Could be good to change it for consistency across the repo.

khuddlefish:
@soumyadeep2007 @chrishajas The only feedback I really have on this is basically nitpicking names, everything else looks good :) I'm still not loving the term "lifecycle commands". Most examples of things that use that term actually allow you access to several different parts of the lifecycle. I don't really think a single command before and a single command after qualifies as proper lifecycle control. We had mention of pre/post backup/restore. Maybe we could do something with "pre" and "post". Also, I somewhat prefer the term "hooks" rather than "commands" since it communicates that there is a separately running backup/restore process, but we give them hooks where they can inject their own functionality into some specific moments during that process.

We could also see what Lisa comes up with since she'll probably be the one adding more information about this in the docs.

soumyadeep2007:
Doing away with that terminology. We could just say "setup/cleanup hooks". This would just totally avoid calling the same thing with different names, which we should avoid doing at all costs. I shouldn't have named them with another term in the first place.

PR #99, opened 2018-07-06
Fix dependency ordering of protocols, tables, and functions.

chrishajas:
Previously, protocols were dumped after tables and functions. However,
protocols can be used in tables (and protocols are dependent on
functions). Therefore, we need to add them to the dependency logic.

### DIFF
--- a/backup/backup.go
+++ b/backup/backup.go
@@ -191,7 +191,7 @@ func backupPredata(metadataFile *utils.FileWithByteCount, tables []Relation, tab
 	BackupCreateSequences(metadataFile, sequences, relationMetadata)
 
 	constraints, conMetadata := RetrieveConstraints()
-	protocols, protoMetadata := RetrieveProtocols(funcInfoMap)
+	protocols, protoMetadata := RetrieveAndProcessProtocols(funcInfoMap)
 
 	BackupFunctionsAndTypesAndTablesAndProtocols(metadataFile, otherFuncs, types, tables, protocols, functionMetadata, typeMetadata, relationMetadata, protoMetadata, tableDefs, constraints)
 	PrintAlterSequenceStatements(metadataFile, globalTOC, sequences, sequenceOwnerColumns)
--- a/backup/backup.go
+++ b/backup/backup.go
@@ -193,7 +193,7 @@ func backupPredata(metadataFile *utils.FileWithByteCount, tables []Relation, tab
 	constraints, conMetadata := RetrieveConstraints()
 	protocols, protoMetadata := RetrieveAndProcessProtocols(funcInfoMap)
 
-	BackupFunctionsAndTypesAndTablesAndProtocols(metadataFile, otherFuncs, types, tables, protocols, functionMetadata, typeMetadata, relationMetadata, protoMetadata, tableDefs, constraints)
+	BackupDependentObjects(metadataFile, otherFuncs, types, tables, protocols, functionMetadata, typeMetadata, relationMetadata, protoMetadata, tableDefs, constraints)
 	PrintAlterSequenceStatements(metadataFile, globalTOC, sequences, sequenceOwnerColumns)
 
 	if len(*includeSchemas) == 0 {
--- a/backup/backup.go
+++ b/backup/backup.go
@@ -191,12 +191,12 @@ func backupPredata(metadataFile *utils.FileWithByteCount, tables []Relation, tab
 	BackupCreateSequences(metadataFile, sequences, relationMetadata)
 
 	constraints, conMetadata := RetrieveConstraints()
+	protocols, protoMetadata := RetrieveProtocols(funcInfoMap)
 
-	BackupFunctionsAndTypesAndTables(metadataFile, otherFuncs, types, tables, functionMetadata, typeMetadata, relationMetadata, tableDefs, constraints)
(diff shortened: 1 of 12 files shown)


backup/backup.go
```diff
@@ -191,12 +191,12 @@ func backupPredata(metadataFile *utils.FileWithByteCount, tables []Relation, tab
 	BackupCreateSequences(metadataFile, sequences, relationMetadata)
 
 	constraints, conMetadata := RetrieveConstraints()
+	protocols, protoMetadata := RetrieveProtocols(funcInfoMap)
 
-	BackupFunctionsAndTypesAndTables(metadataFile, otherFuncs, types, tables, functionMetadata, typeMetadata, relationMetadata, tableDefs, constraints)
+	BackupFunctionsAndTypesAndTablesAndProtocols(metadataFile, otherFuncs, types, tables, protocols, functionMetadata, typeMetadata, relationMetadata, protoMetadata, tableDefs, constraints)
```
jmcatamney: The names of this function and the other `FunctionsAndTypesAnd...` functions are getting a bit unwieldy, and we're passing in slices of different objects in the same order so the names aren't adding much information.

Would it make sense to simplify them to something like `BackupDependentObjects`, `SortObjectsInDependencyOrder`, etc. and add comments in `dependencies.go` explaining which types need dependency sorting and why?

chrishajas: Agreed, I'll change it.

backup/dependencies.go
```diff
@@ -21,11 +21,14 @@ func SortFunctionsAndTypesAndTablesInDependencyOrder(functions []Function, types
...
 	}
+	for _, protocol := range protocols {
+		objects = append(objects, protocol)
+	}
 	sorted := TopologicalSort(objects)
 	return sorted
 }
 
-func ConstructFunctionAndTypeAndTableMetadataMap(functions MetadataMap, types MetadataMap, tables MetadataMap) MetadataMap {
+func ConstructFunctionAndTypeAndTableMetadataMap(functions MetadataMap, types MetadataMap, tables MetadataMap, protocols MetadataMap) MetadataMap {
```
khuddlefish: We missed adding "AndProtocol" to this function (though I think we'll be changing all of these names anyways)

backup/backup.go
khuddlefish: @chrishajas It looks like this didn't get changed yet? +1 on simplifying all the silly long names.

backup/dependencies_test.go
```diff
@@ -26,12 +26,15 @@ var _ = Describe("backup/dependencies tests", func() {
 		view1     backup.View
 		view2     backup.View
 		view3     backup.View
```
khuddlefish: I was considering taking the view objects out of these tests (switching them for tables). It's a bit confusing having them there since we don't actually sort views along with the other objects for dependency ordering. I think we used them initially because they also use the relation struct.

Probably doesn't need to be part of this story, but could be good to do.

chrishajas: Yeah, this area could use some TLC. Gonna make a chore for it.

integration/predata_externals_create_test.go
```diff
@@ -97,21 +97,31 @@ var _ = Describe("backup integration create statement tests", func() {
...
+				2: "public.read_from_s3",
+			},
 		}
-		protocolReadOnly := backup.ExternalProtocol{Oid: 1, Name: "s3_read", Owner: "testrole", Trusted: true, ReadFunction: 2, WriteFunction: 0, Validator: 0}
-		protocolWriteOnly := backup.ExternalProtocol{Oid: 1, Name: "s3_write", Owner: "testrole", Trusted: false, ReadFunction: 0, WriteFunction: 1, Validator: 0}
-		protocolReadWrite := backup.ExternalProtocol{Oid: 1, Name: "s3_read_write", Owner: "testrole", Trusted: false, ReadFunction: 2, WriteFunction: 1, Validator: 0}
-		emptyMetadataMap := backup.MetadataMap{}
+		protocolWriteOnly := backup.ExternalProtocol{Oid: 1, Name: "s3_write", Owner: "testrole", Trusted: false, ReadFunction: 0, WriteFunction: 1, Validator: 0,
+			FuncMap: map[uint32]string{
+				1: "public.write_to_s3",
```
soumyadeep2007: Some extra whitespace here

backup/backup.go
```diff
@@ -191,12 +191,12 @@ func backupPredata(metadataFile *utils.FileWithByteCount, tables []Relation, tab
 	BackupCreateSequences(metadataFile, sequences, relationMetadata)
 
 	constraints, conMetadata := RetrieveConstraints()
+	protocols, protoMetadata := RetrieveAndProcessProtocols(funcInfoMap)
 
-	BackupFunctionsAndTypesAndTables(metadataFile, otherFuncs, types, tables, functionMetadata, typeMetadata, relationMetadata, tableDefs, constraints)
+	BackupFunctionsAndTypesAndTablesAndProtocols(metadataFile, otherFuncs, types, tables, protocols, functionMetadata, typeMetadata, relationMetadata, protoMetadata, tableDefs, constraints)
```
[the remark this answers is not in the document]
khuddlefish: The whole reason we do the function this way is because the dependencies are intertwined so you can't back them up in separate functions.

PR #101, opened 2018-07-10
Fixed indexOOBounds for empty restore plan when we have single-data-file

soumyadeep2007:
Now simply returning from the restoreDataFromTimestamp if the restore plan is empty.

### DIFF
--- a/restore/restore.go
+++ b/restore/restore.go
@@ -211,6 +211,11 @@ func restoreDataFromTimestamp(fpInfo utils.FilePathInfo, restorePlanTableFQNs []
 	filteredMasterDataEntries := toc.GetDataEntriesMatching(*includeSchemas,
 		*excludeSchemas, *includeRelations, *excludeRelations, restorePlanTableFQNs)
 
+	if len(filteredMasterDataEntries) == 0 {
+		gplog.Info("Found no tables to restore in restore plan for backup with timestamp: %s", fpInfo.Timestamp)
+		return
+	}
+
 	if backupConfig.SingleDataFile {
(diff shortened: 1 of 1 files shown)


chrishajas:
Why did this manifest only in the single-data-file case? We should also test single-data-file with this if the logic diverges.

restore/restore.go
```diff
@@ -211,6 +211,11 @@ func restoreDataFromTimestamp(fpInfo utils.FilePathInfo, restorePlanTableFQNs []
 	filteredMasterDataEntries := toc.GetDataEntriesMatching(*includeSchemas,
 		*excludeSchemas, *includeRelations, *excludeRelations, restorePlanTableFQNs)
 
+	if len(filteredMasterDataEntries) == 0 {
+		gplog.Info("Found no fresh entries to restore from backup with timestamp: %s", fpInfo.Timestamp)
```
chrishajas: We can say `Found no ~fresh entries~ tables to restore from backup with timestamp:

soumyadeep2007:
This manifested only in the single-data-file case as there was an explicit slice subscript access for the index 0 in the code branch. It did not manifest later on in the code as there were no similar accesses and all accesses were for ranges over the slice.

restore/restore.go
soumyadeep2007: Reason I added "fresh" is because I didn't want anyone seeing the log to mistakenly think that there are no tables in the backup. So how about "Found no tables to restore in restore plan for backup with timstamp.."?

chrishajas: Yep, that works

soumyadeep2007:
Sure, will do!

PR #102, opened 2018-07-11
Manually add protocol dependencies to external tables for GPDB4

chrishajas:
This builds upon the protocol dependencies that had been reverted.

In GPDB4, there is no entry in pg_depend that maps an external table to
its corresponding protocol. While this was fixed in GPDB5, we need to
work around this limitation in GPDB4 by manually adding the dependency.


chrishajas:
@soumyadeep2007 It's tested in the `constructs dependencies correctly if there is one table dependent on one protocol` scenario. I was torn between putting it within ConstructTableDependencies or putting it in a separate function. If I keep it within this function, I can make it private.

PR #103, opened 2018-07-11
Add concourse job for end_to_end tests using gpbackup 1.0.0 and latest restore

khuddlefish:
This will help us ensure that the changes we make to gprestore will not break our ability to restore backups taken with an older version of gpbackup.

This uses the USE_OLD_BACKUP_VERSION environment variable to tell the tests to run with the old version. If the variable is false, the tests will run as before. The suite does `git checkout 1.0.0`, builds the binaries, and installs, then checks out back to latest. We need to reinstall the two versions of gpbackup_helper between every backup and restore because the utilities simply use the version located in $GPHOME/bin and it needs to be the same version as backup/restore.

We skip running some tests because features were not supported or changed. The --include/exclude-table{file} tests would fail because we previously didn't support including/excluding sequences and views.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -187,10 +187,10 @@
   version = "v1.0.22"
 
 [[projects]]
+  branch = "v2"
   name = "gopkg.in/yaml.v2"
   packages = ["."]
   revision = "5420a8b6744d3b0345ab293f6fcba19c978f1183"
-  version = "v2.2.1"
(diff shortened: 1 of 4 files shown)


chrishajas:
We should add this job as a blocker for packaging.

khuddlefish:
Added the blocker to packaging :)

PR #104, opened 2018-07-12
Modify how we print Owner for views and sequences

khuddlefish:
* Print owner for views in all versions (we weren't doing this before)
* Print ALTER SEQUENCE/VIEW for sequences and views in 6+ but ALTER TABLE
before that

### DIFF
--- a/backup/predata_relations_test.go
+++ b/backup/predata_relations_test.go
@@ -653,7 +653,8 @@ SELECT pg_catalog.setval('public.seq_name', 7, true);`)
 
 SELECT pg_catalog.setval('public.seq_name', 7, false);`)
 		})
-		It("can print a sequence with privileges, an owner, and a comment", func() {
+		It("can print a sequence with privileges, an owner, and a comment for version < 6", func() {
+			testutils.SetDBVersion(connectionPool, "5.0.0")
 			sequenceMetadataMap := testutils.DefaultMetadataMap("SEQUENCE", true, true, true)
 			sequenceMetadata := sequenceMetadataMap[1]
 			sequenceMetadata.Privileges[0].Update = false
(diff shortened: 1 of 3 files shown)


backup/predata_relations_test.go
```diff
@@ -711,7 +742,30 @@ GRANT SELECT,USAGE ON SEQUENCE public.seq_name TO testrole WITH GRANT OPTION;`)
...
+COMMENT ON VIEW shamwow.shazam IS 'This is a view comment.';
+
+
+ALTER TABLE shamwow.shazam OWNER TO testrole;
+
+
+REVOKE ALL ON shamwow.shazam FROM PUBLIC;
+REVOKE ALL ON shamwow.shazam FROM testrole;
+GRANT ALL ON shamwow.shazam TO testrole;`)
+			testutils.SetDBVersion(connectionPool, "5.1.0")
```
chrishajas: Should these be in defers after the initial set in case the test fails it properly sets the version back?

backup/predata_shared_test.go
```diff
@@ -346,6 +346,40 @@ REVOKE ALL ON FOREIGN SERVER foreignserver FROM PUBLIC;
 REVOKE ALL ON FOREIGN SERVER foreignserver FROM testrole;
 GRANT ALL ON FOREIGN SERVER foreignserver TO testrole;`)
 		})
+		Context("Prints owners for views and sequences", func() {
```
soumyadeep2007: Context description text could be reworded to: "Views and Sequences have owners" as typically you would describe the scenario in the context text rather than what it does.

backup/predata_relations_test.go
khuddlefish: Probably a good idea. I'll change it

PR #109, opened 2018-07-13
Add content id argument to setup/cleanup commands

chrishajas:
Now passing in the content id for scope=master, segment to setup/cleanup methods. For scope=segment_host, content id is not passed in.

Co-authored-by: Soumyadeep Chakraborty <sochakraborty@pivotal.io>

### DIFF
--- a/helper/helper.go
+++ b/helper/helper.go
@@ -390,6 +390,6 @@ func DoCleanup() {
 }
 
 func log(s string, v ...interface{}) {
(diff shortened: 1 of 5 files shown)


khuddlefish:
This is passing in the segment backup directories now right?

chrishajas:
@khuddlefish Yes, that was fixed in https://github.com/greenplum-db/gpbackup/commit/238de87afa90001f93d0d4f31ec0f26fd1808448.

PR #112, opened 2018-07-17
Add support for tablespace modifications introduced in master

chrishajas:
The FILESPACE keyword no longer exists in the GPDB master branch and has
been replaced with new syntax that specifies the directory for each
segment. This commit adds support for this new syntax.

### DIFF
--- a/backup/queries_globals.go
+++ b/backup/queries_globals.go
@@ -320,15 +320,11 @@ ORDER BY roleid, member;`
 	return results
 }
 
-type SegmentTablespace struct {
-	Tablespace   string
-	FileLocation string
-}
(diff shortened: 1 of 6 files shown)


backup/queries_globals.go
```diff
@@ -320,15 +320,11 @@ ORDER BY roleid, member;`
 	return results
 }
 
-type SegmentTablespace struct {
-	Tablespace   string
```
soumyadeep2007: Would we ever want to access the tablespace or the file location in isolation? If we do, perhaps it would be best to keep it as a struct. It avoids the need for parsing.

PR #113, opened 2018-07-19
Order labels for enum types

khuddlefish:
* ORDER BY oid in gpdb5
* ORDER BY enumsortorder in gpdb6+. This is because a feature was added
in Postgres 9.1 to add enum labels BEFORE or AFTER other labels so the
oids might not tell you the right order.

### DIFF
--- a/backup/queries_types.go
+++ b/backup/queries_types.go
@@ -261,6 +261,10 @@ ORDER BY n.nspname, t.typname;`, SchemaFilterClause("n"), ExtensionFilterClause(
 }
 
 func GetEnumTypes(connection *dbconn.DBConn) []Type {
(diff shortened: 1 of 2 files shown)


backup/queries_types.go
```diff
@@ -261,6 +261,10 @@ ORDER BY n.nspname, t.typname;`, SchemaFilterClause("n"), ExtensionFilterClause(
 }
 
 func GetEnumTypes(connection *dbconn.DBConn) []Type {
+	enumSortClause := "ORDER BY e.enumsortorder"
+	if connection.Version.Is("5") {
+		enumSortClause = "ORDER BY e.oid"
```
chrishajas: Nice find!

PR #114, opened 2018-07-24
Update queries_relations.go

water32:
Let the planner choose which table to use for the hash, rather than the rough left join pg_attribute table.


chrishajas:
@water32 Thanks for your contribution!

PR #118, opened 2018-07-25
Add support for collations on domains, table columns, and composite type columns

chrishajas:

### DIFF
--- a/backup/predata_types_test.go
+++ b/backup/predata_types_test.go
@@ -47,12 +47,25 @@ ALTER TYPE public.enum_type OWNER TO testrole;`)
 		})
 	})
 	Describe("PrintCreateCompositeTypeStatement", func() {
-		oneAtt := pq.StringArray{"\tfoo integer COLLATE public.some_coll"}
-		twoAtts := pq.StringArray{"\tfoo integer", "\tbar text"}
-		compType := backup.Type{Oid: 1, Schema: "public", Name: "composite_type", Type: "c", Category: "U"}
+		var compType backup.Type
+		var oneAtt, oneAttWithCollation, twoAtts pq.StringArray
+		BeforeEach(func() {
+			compType = backup.Type{Oid: 1, Schema: "public", Name: "composite_type", Type: "c", Category: "U"}
+			oneAtt = pq.StringArray{"\tfoo integer"}
+			oneAttWithCollation = pq.StringArray{"\tfoo integer COLLATE public.some_coll"}
+			twoAtts = pq.StringArray{"\tfoo integer", "\tbar text"}
+		})
 
 		It("prints a composite type with one attribute", func() {
 			compType.Attributes = oneAtt
+			backup.PrintCreateCompositeTypeStatement(backupfile, toc, compType, typeMetadata)
+			testutils.ExpectEntry(toc.PredataEntries, 0, "public", "", "composite_type", "TYPE")
+			testutils.AssertBufferContents(toc.PredataEntries, buffer, `CREATE TYPE public.composite_type AS (
+	foo integer
+);`)
+		})
+		It("prints a composite type with one attribute with a collation", func() {
+			compType.Attributes = oneAttWithCollation
 			backup.PrintCreateCompositeTypeStatement(backupfile, toc, compType, typeMetadata)
 			testutils.ExpectEntry(toc.PredataEntries, 0, "public", "", "composite_type", "TYPE")
 			testutils.AssertBufferContents(toc.PredataEntries, buffer, `CREATE TYPE public.composite_type AS (
@@ -186,13 +199,13 @@ ALTER TYPE public.base_type
 		domainOne.Collation = "public.mycollation"
 		domainTwo := testutils.DefaultTypeDefinition("d", "domain2")
 		domainTwo.BaseType = "varchar"
-		It("prints a basic domain with a constraint", func() {
+		It("prints a domain with a constraint", func() {
 			backup.PrintCreateDomainStatement(backupfile, toc, domainOne, emptyMetadata, checkConstraint)
 			testutils.ExpectEntry(toc.PredataEntries, 0, "public", "", "domain1", "DOMAIN")
 			testutils.AssertBufferContents(toc.PredataEntries, buffer, `CREATE DOMAIN public.domain1 AS numeric DEFAULT 4 COLLATE public.mycollation NOT NULL
 	CONSTRAINT domain1_check CHECK (VALUE > 2);`)
 		})
-		It("prints a basic domain without constraint", func() {
+		It("prints a domain without constraint", func() {
 			backup.PrintCreateDomainStatement(backupfile, toc, domainOne, emptyMetadata, emptyConstraint)
 			testutils.AssertBufferContents(toc.PredataEntries, buffer, `CREATE DOMAIN public.domain1 AS numeric DEFAULT 4 COLLATE public.mycollation NOT NULL;`)
 		})
--- a/backup/predata_types_test.go
+++ b/backup/predata_types_test.go
@@ -183,17 +183,18 @@ ALTER TYPE public.base_type
 		domainOne.DefaultVal = "4"
 		domainOne.BaseType = "numeric"
 		domainOne.NotNull = true
+		domainOne.Collation = "public.mycollation"
 		domainTwo := testutils.DefaultTypeDefinition("d", "domain2")
 		domainTwo.BaseType = "varchar"
 		It("prints a basic domain with a constraint", func() {
 			backup.PrintCreateDomainStatement(backupfile, toc, domainOne, emptyMetadata, checkConstraint)
 			testutils.ExpectEntry(toc.PredataEntries, 0, "public", "", "domain1", "DOMAIN")
-			testutils.AssertBufferContents(toc.PredataEntries, buffer, `CREATE DOMAIN public.domain1 AS numeric DEFAULT 4 NOT NULL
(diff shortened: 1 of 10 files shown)


backup/predata_types_test.go
```diff
@@ -182,17 +182,18 @@ ALTER TYPE public.base_type
 		domainOne.DefaultVal = "4"
 		domainOne.BaseType = "numeric"
 		domainOne.NotNull = true
+		domainOne.Collation = "public.mycollation"
 		domainTwo := testutils.DefaultTypeDefinition("d", "domain2")
 		domainTwo.BaseType = "varchar"
 		It("prints a basic domain with a constraint", func() {
 			backup.PrintCreateDomainStatement(backupfile, toc, domainOne, emptyMetadata, checkConstraint)
 			testutils.ExpectEntry(toc.PredataEntries, 0, "public", "", "domain1", "DOMAIN")
-			testutils.AssertBufferContents(toc.PredataEntries, buffer, `CREATE DOMAIN public.domain1 AS numeric DEFAULT 4 NOT NULL
```
khuddlefish: Does it make sense for the collations to just be mixed in with other existing tests? Does this qualify as a "basic domain"? Also will we have an GPDB version issues or do these objects as a whole only exist in master?

backup/queries_relations.go
```diff
@@ -219,11 +221,8 @@ var storageTypeCodes = map[string]string{
 
 func GetColumnDefinitions(connection *dbconn.DBConn, columnMetadata map[uint32]map[string][]ACL) map[uint32][]ColumnDefinition {
 	// This query is adapted from the getTableAttrs() function in pg_dump.c.
-	optionsQuery := ""
```
khuddlefish: Are the only two differences between the queries the options and collation, or are there more? It is nice to see the queries all together, but if it is just inserting one extra section, it can actually make it more clear what the differences are if we are just working with the one query and formatting. The way it is now I can't easily see what changes between the two because they're both pretty large.

integration/predata_types_create_test.go
```diff
@@ -40,8 +40,8 @@ var _ = Describe("backup integration create statement tests", func() {
 			}
 			enumType = backup.Type{Type: "e", Schema: "public", Name: "enum_type", EnumLabels: "'enum_labels'", Category: "U"}
 			domainType = testutils.DefaultTypeDefinition("d", "domain_type")
-			domainType.BaseType = "numeric"
-			domainType.DefaultVal = "5"
```
khuddlefish: Does this determine the collation for the domain type? Otherwise, I can't see anywhere that connects domainType to public.some_coll

integration/predata_types_queries_test.go
```diff
@@ -56,7 +56,14 @@ var _ = Describe("backup integration tests", func() {
 			structmatcher.ExpectStructsToMatchIncluding(&shellType, &results[0], "Schema", "Name", "Type")
 		})
 		It("returns a slice of composite types", func() {
-			testhelper.AssertQueryRuns(connection, "CREATE TYPE public.composite_type AS (name int4, name2 numeric(8,2), name1 text);")
+			if connection.Version.AtLeast("6") {
```
khuddlefish: Should these be two separate tests?

backup/queries_relations.go
chrishajas: There's a couple more joins too. It was getting to the point where we'd need a couple more interpolated lines.

integration/predata_types_queries_test.go
```diff
@@ -157,9 +164,16 @@ var _ = Describe("backup integration tests", func() {
 		})
 		It("returns a slice for a domain type", func() {
 			domainType := backup.Type{
-				Oid: 1, Type: "d", Schema: "public", Name: "domain1", DefaultVal: "4", BaseType: "numeric(8,2)", NotNull: false,
+				Oid: 1, Type: "d", Schema: "public", Name: "domain1", DefaultVal: "'abc'::bpchar", BaseType: "character(8)", NotNull: false,
+			}
+			if connection.Version.Before("6") {
```
khuddlefish: We should be consistent with whether we do "if Before("6") else" or "if AtLeast("6") else"

Also this should probably be separate too.

backup/queries_relations.go
khuddlefish: Ok, that's fine then

backup/predata_types_test.go
chrishajas: I can add them in separate test. This is just a unit test, so it'll work fine on any version. We included default in the "basic" domain, so we can also just remove the "basic" word. Looking into the future, domains aren't changed at all in any future versions.

integration/predata_types_create_test.go
chrishajas: We can't have collations on numeric types, so this was just to allow reuse of the same `CREATE DOMAIN` statement.

integration/predata_types_queries_test.go
chrishajas: Yeah, I'll do that.

backup/predata_types_test.go
khuddlefish: That's actually an issue with some of our unit tests. The version we use is "5.1.0" so if there are any master specific things, we should do a separate test and set the fake GPDBVersion to be 6.

chrishajas: There isn't any version check in the vast majority of our Print* code, we keep most of it in the queries. We usually just check if the field exists and do logic based on that.

backup/queries_relations.go
```diff
@@ -247,10 +244,44 @@ LEFT JOIN pg_description d ON (d.objoid = a.attrelid AND d.classoid = 'pg_class'
 WHERE %s
 AND a.attnum > 0::pg_catalog.int2
 AND a.attisdropped = 'f'
-ORDER BY a.attrelid, a.attnum;`, optionsQuery, relationAndSchemaFilterClause())
+ORDER BY a.attrelid, a.attnum;`, relationAndSchemaFilterClause())
+
+	masterQuery := fmt.Sprintf(`
+SELECT
```
khuddlefish: What is the cost of this query? I think I remember this used to be expensive before adding collations and this can only make it worse.

backup/predata_types_test.go
```diff
@@ -47,7 +47,7 @@ ALTER TYPE public.enum_type OWNER TO testrole;`)
 		})
 	})
 	Describe("PrintCreateCompositeTypeStatement", func() {
-		oneAtt := pq.StringArray{"\tfoo integer"}
+		oneAtt := pq.StringArray{"\tfoo integer COLLATE public.some_coll"}
```
khuddlefish: Are collations default behavior in 6? I still feel weird about throwing them into default cases without specifically calling out that we're doing it. Especially since they only exist in master.

integration/predata_types_queries_test.go
```diff
@@ -57,7 +57,20 @@ var _ = Describe("backup integration tests", func() {
...
+
+			Expect(len(results)).To(Equal(1))
+			structmatcher.ExpectStructsToMatchIncluding(&compositeType, &results[0], "Type", "Schema", "Name", "Attributes")
+		})
+		It("returns a slice of composite types with collations", func() {
+			testutils.SkipIfBefore6(connection)
+			testhelper.AssertQueryRuns(connection, `CREATE COLLATION public.some_coll (lc_collate = 'POSIX', lc_ctype = 'POSIX');`)
+			defer testhelper.AssertQueryRuns(connection, "DROP COLLATION public.some_coll")
+			testhelper.AssertQueryRuns(connection, "CREATE TYPE public.composite_type AS (name int4, name2 numeric(8,2), name1 character(8) COLLATE public.some_coll);")
+			compositeType.Attributes = pq.StringArray{"\tname integer", "\tname2 numeric(8,2)", "\tname1 character(8) COLLATE public.some_coll"}
```
khuddlefish: Is this testing what we want it to test if we are setting Attributes ourselves like this? Do we have a function we should we calling that does this for us?

backup/queries_relations.go
chrishajas: Before: 20996.94, 120.354ms
After: 24047.05, 125.629ms

I don't anticipate users having many collations, so it shouldn't add much, but it is in one of the slower parts of our metadata backup (along with types and the incremental metadata). Since it's a 6 only query, we can optimize once master is in a better state if it's a slower query.

integration/predata_types_queries_test.go
chrishajas: This is modifying the expected compositeType instead of a separate compositeType for each test (since it's identical except for the Attribute).

khuddlefish: In that case can we move it down? Maybe right before we "ExpectStructsToMatch"? At least after we finish running the actual queries.

chrishajas: Done

backup/predata_types_test.go
chrishajas: They aren't default, but the query itself returns the entire attribute string for composite types (ie: If it contains a collation, it will return `foo integer COLLATE public.some_coll`, if it does not, it will return `foo integer COLLATE public.some_coll`). Since we mock this string, we're not actually testing anything new. I added a separate test for this, I can keep it if you'd like.

backup/predata_types_test.go
```diff
@@ -47,16 +47,28 @@ ALTER TYPE public.enum_type OWNER TO testrole;`)
 		})
 	})
 	Describe("PrintCreateCompositeTypeStatement", func() {
+		var compType backup.Type
 		oneAtt := pq.StringArray{"\tfoo integer"}
+		oneAttWithCollation := pq.StringArray{"\tfoo integer COLLATE public.some_coll"}
```
khuddlefish: Can we put these all in the BeforeEach for consistency?

PR #123, opened 2018-07-28
Fixes seg prefix resolution when multiple prefixed dirs are present

soumyadeep2007:
Before this fix: GetSegPrefix would incorrectly match on the
directory lexicographically occurring first in order to compute the
prefix. Now, we also incorporate the timestamp into the resolution
process ensuring a correct match.

Authored-by: Soumyadeep Chakraborty <sochakraborty@pivotal.io>

### DIFF
--- a/utils/filepath.go
+++ b/utils/filepath.go
@@ -162,14 +162,12 @@ func GetSegPrefix(connection *dbconn.DBConn) string {
 func ParseSegPrefix(backupDir string, timestamp string) string {
 	segPrefix := ""
 	if len(backupDir) > 0 {
-		masterDir, err := operating.System.Glob(fmt.Sprintf("%s/*-1/backups/*/%s", backupDir, timestamp))
-		if err != nil || len(masterDir) == 0 {
+		backupDirForTimestamp, err := operating.System.Glob(fmt.Sprintf("%s/*-1/backups/*/%s", backupDir, timestamp))
+		if err != nil || len(backupDirForTimestamp) == 0 {
 			gplog.Fatal(err, "Master backup directory in %s missing or inaccessible", backupDir)
 		}
-		indexOfBackupsSubstr := strings.LastIndex(masterDir[0], "-1/backups/")
-		segPrefix = masterDir[0][:indexOfBackupsSubstr]
-		_, segPrefix = path.Split(segPrefix)
-		fmt.Println(segPrefix)
+		indexOfBackupsSubstr := strings.LastIndex(backupDirForTimestamp[0], "-1/backups/")
+		_, segPrefix = path.Split(backupDirForTimestamp[0][:indexOfBackupsSubstr])
 	}
 	return segPrefix
 }
--- a/utils/filepath.go
+++ b/utils/filepath.go
@@ -159,15 +159,17 @@ func GetSegPrefix(connection *dbconn.DBConn) string {
 	return segPrefix
 }
 
-func ParseSegPrefix(backupDir string) string {
+func ParseSegPrefix(backupDir string, timestamp string) string {
 	segPrefix := ""
 	if len(backupDir) > 0 {
-		masterDir, err := operating.System.Glob(fmt.Sprintf("%s/*-1", backupDir))
+		masterDir, err := operating.System.Glob(fmt.Sprintf("%s/*-1/backups/*/%s", backupDir, timestamp))
 		if err != nil || len(masterDir) == 0 {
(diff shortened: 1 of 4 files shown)


utils/filepath.go
```diff
@@ -159,15 +159,17 @@ func GetSegPrefix(connection *dbconn.DBConn) string {
...
+		masterDir, err := operating.System.Glob(fmt.Sprintf("%s/*-1/backups/*/%s", backupDir, timestamp))
 		if err != nil || len(masterDir) == 0 {
 			gplog.Fatal(err, "Master backup directory in %s missing or inaccessible", backupDir)
 		}
-		_, segPrefix = path.Split(masterDir[0])
-		segPrefix = segPrefix[:len(segPrefix)-2] // Remove "-1" segment ID from string
+		indexOfBackupsSubstr := strings.LastIndex(masterDir[0], "-1/backups/")
+		segPrefix = masterDir[0][:indexOfBackupsSubstr]
+		_, segPrefix = path.Split(segPrefix)
+		fmt.Println(segPrefix)
```
chrishajas: Remove println

utils/filepath.go
```diff
@@ -159,15 +159,17 @@ func GetSegPrefix(connection *dbconn.DBConn) string {
 	return segPrefix
 }
 
-func ParseSegPrefix(backupDir string) string {
+func ParseSegPrefix(backupDir string, timestamp string) string {
 	segPrefix := ""
 	if len(backupDir) > 0 {
-		masterDir, err := operating.System.Glob(fmt.Sprintf("%s/*-1", backupDir))
+		masterDir, err := operating.System.Glob(fmt.Sprintf("%s/*-1/backups/*/%s", backupDir, timestamp))
```
chrishajas: This is no longer the masterDir, but rather the backup directory for the timestamp.

utils/filepath.go
```diff
@@ -159,15 +159,17 @@ func GetSegPrefix(connection *dbconn.DBConn) string {
...
 		if err != nil || len(masterDir) == 0 {
 			gplog.Fatal(err, "Master backup directory in %s missing or inaccessible", backupDir)
 		}
-		_, segPrefix = path.Split(masterDir[0])
-		segPrefix = segPrefix[:len(segPrefix)-2] // Remove "-1" segment ID from string
+		indexOfBackupsSubstr := strings.LastIndex(masterDir[0], "-1/backups/")
+		segPrefix = masterDir[0][:indexOfBackupsSubstr]
```
chrishajas: This isn't just the segPrefix right, does it also include the path up to the segPrefix?

soumyadeep2007: After the split is executed, it is just the prefix.

utils/filepath.go
soumyadeep2007: Will do!

utils/filepath.go
soumyadeep2007: I was wondering what to call that! Thanks!

utils/filepath.go
```diff
@@ -159,15 +159,15 @@ func GetSegPrefix(connection *dbconn.DBConn) string {
...
 	if len(backupDir) > 0 {
-		masterDir, err := operating.System.Glob(fmt.Sprintf("%s/*-1", backupDir))
-		if err != nil || len(masterDir) == 0 {
+		backupDirForTimestamp, err := operating.System.Glob(fmt.Sprintf("%s/*-1/backups/*/%s", backupDir, timestamp))
+		if err != nil || len(backupDirForTimestamp) == 0 {
 			gplog.Fatal(err, "Master backup directory in %s missing or inaccessible", backupDir)
 		}
-		_, segPrefix = path.Split(masterDir[0])
-		segPrefix = segPrefix[:len(segPrefix)-2] // Remove "-1" segment ID from string
+		indexOfBackupsSubstr := strings.LastIndex(backupDirForTimestamp[0], "-1/backups/")
```
khuddlefish: Do we have any concern that we might get more than 1 result?

khuddlefish: Also, does this get the index of "backups/" or "-1/backups/"?

soumyadeep2007: Now we are searching with the timestamp in the dir path. So there can be only one directory captured in backupDirForTimestamp.

Gets last index of "-1/backups/".

khuddlefish:
Do we have any tests that simulate the original issue we were seeing of getting multiple backup dirs so we can confirm that our original issue isn't present anymore?

chrishajas:
I think most of the logic is in the glob'd result, which we mock.

PR #124, opened 2018-07-31
Unified progress bars for restore

soumyadeep2007:
Previously, we had a separate progress bar for each backup in the backup
set. Now we have one across all backups.

Authored-by: Soumyadeep Chakraborty <sochakraborty@pivotal.io>

### DIFF
--- a/restore/restore.go
+++ b/restore/restore.go
@@ -197,46 +197,58 @@ func restoreData(fpInfoList []utils.FilePathInfo, gucStatements []utils.Statemen
 		return
 	}
 	latestRestorePlan := backupConfig.RestorePlan
+
+	totalTables := 0
+	filteredDataEntries := make([][]utils.MasterDataEntry, 0)
 	for i, fpInfo := range fpInfoList {
-		gplog.Info("Restoring data from backup with timestamp: %s", fpInfo.Timestamp)
+		tocFilename := fpInfo.GetTOCFilePath()
+		toc := utils.NewTOC(tocFilename)
 		restorePlanTableFQNs := latestRestorePlan[i].TableFQNs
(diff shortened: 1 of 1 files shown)


restore/restore.go
```diff
@@ -197,46 +197,58 @@ func restoreData(fpInfoList []utils.FilePathInfo, gucStatements []utils.Statemen
...
+		tocFilename := fpInfo.GetTOCFilePath()
+		toc := utils.NewTOC(tocFilename)
 		restorePlanTableFQNs := latestRestorePlan[i].TableFQNs
-		restoreDataFromTimestamp(fpInfo, restorePlanTableFQNs, gucStatements)
+		filteredDataEntriesForTimestamp := toc.GetDataEntriesMatching(MustGetFlagStringSlice(utils.INCLUDE_SCHEMA),
+			MustGetFlagStringSlice(utils.EXCLUDE_SCHEMA), MustGetFlagStringSlice(utils.INCLUDE_RELATION),
+			MustGetFlagStringSlice(utils.EXCLUDE_RELATION), restorePlanTableFQNs)
+		filteredDataEntries = append(filteredDataEntries, filteredDataEntriesForTimestamp)
+
+		totalTables += len(filteredDataEntriesForTimestamp)
```
chrishajas: Can we set totalTables to `len(filteredDataEntries)` outside the loop?

soumyadeep2007: Umm..filteredDataEntries is a slice of slices. So len(filteredDataEntries) would only give the no of backups in the backup set in that case.

khuddlefish: No because filteredDataEntries is a [][]Entries. The len will be the number of timestamps, not tables I think

PR #127, opened 2018-08-02
Add github release to packaging step

chrishajas:
Uploads gpbackup, gprestore, and gpbackup_helper binaries to github
releases when gpbackup is tagged.


khuddlefish:
In addition to publishing releases, it might be nice to start using annotated tags

chrishajas:
I agree. I thought that's part of the tagging process when we manually set the tag and the associated message for the tag?

khuddlefish:
Yes it is part of the tagging process, not technically related to this PR. We just aren't doing annotated tags right now and I think they might look nicer now that we're making proper releases.

PR #128, opened 2018-08-02
Support 5th slot in pg_statistics introduced in 9.2 merge

chrishajas:
Also, disable ACL test for external protocol due to bug in GPDB master.

### DIFF
--- a/backup/statistics.go
+++ b/backup/statistics.go
@@ -58,9 +58,14 @@ func GenerateAttributeStatisticsQuery(table Relation, attStat AttributeStatistic
 	starelidStr := fmt.Sprintf("'%s'::regclass::oid", utils.EscapeSingleQuotes(table.ToString()))
 	// The entry may or may not already exist, so we can't either just UPDATE or just INSERT without a DELETE.
 	inheritStr := ""
+	attributeSlotsQueryStr := ""
 	if connectionPool.Version.AtLeast("6") {
 		inheritStr = fmt.Sprintf("\n\t%t::boolean,", attStat.Inherit)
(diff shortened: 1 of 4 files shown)


backup/statistics.go
```diff
@@ -75,12 +75,28 @@ INSERT INTO pg_statistic VALUES (
 	 * type.  We can't restore statistics of array columns, so we'll zero and
 	 * NULL everything out.
 	 */
+
+	if connectionPool.Version.AtLeast("6") {
+		attributeQuery += generateAttributeQueryMaster(attStat)
```
khuddlefish: Appending it here is a little hard for me to follow. Would it be possible to generate the query and format it in?

backup/statistics.go
```diff
@@ -89,12 +105,79 @@ INSERT INTO pg_statistic VALUES (
...
+	%s::real[],
+	%s::real[],
+	%s::real[],
+	%s::real[],
+	%s::real[],
+	%s,
+	%s,
+	%s,
+	%s,
+	%s`, attStat.Kind1,
```
soumyadeep2007: Is formatting a bit off here?

backup/statistics.go
chrishajas: I've added another commit with this implemented to compare. Neither is pretty since the attribute order matters when we insert values into the table. Also consider that this part will not change in the foreseeable future (at least through PG11).

backup/statistics.go
chrishajas: Which part? Do you think the attStat should be less indented? The formatting is handled by goimports/gofmt.

backup/statistics.go
chrishajas: @khuddlefish Done!

PR #129, opened 2018-08-07
Optimize ACL logic in metadata portion of backup

chrishajas:
Previously, we compiled the regex within a loop, even though the
expression never changed. This took a considerable amount of time, in
some cases up to 1/3 of the total backup time. Now, we compute the regex
once and pass it to the ParseACL function.

A complex metadata backup went from 3m50s to 2m30s on my machine.

### DIFF
--- a/backup/predata_relations.go
+++ b/backup/predata_relations.go
@@ -7,7 +7,6 @@ package backup
 
 import (
 	"fmt"
-	"regexp"
 	"strings"
 
 	"math"
@@ -211,7 +210,6 @@ func ConstructColumnPrivilegesMap(results []ColumnPrivilegesQueryStruct) map[uin
 		 * a nested map indexed by table oid.
 		 */
 		tableMetadata = make(map[string][]ACL, 0)
-		aclRegex := regexp.MustCompile(`^(.*)=([a-zA-Z\*]*)/(.*)$`)
 		for _, result := range results {
 			privilegesStr := ""
 			if result.Kind == "Empty" {
@@ -231,7 +229,7 @@ func ConstructColumnPrivilegesMap(results []ColumnPrivilegesQueryStruct) map[uin
 				currentColumn = result.Name
 				columnMetadata = make([]ACL, 0)
 			}
-			privileges := ParseACL(privilegesStr, quotedRoleNames, aclRegex)
+			privileges := ParseACL(privilegesStr, quotedRoleNames, ACLRegex)
 			if privileges != nil {
 				columnMetadata = append(columnMetadata, *privileges)
 			}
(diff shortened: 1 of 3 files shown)


backup/predata_shared.go
```diff
@@ -148,6 +148,7 @@ func ConstructMetadataMap(results []MetadataQueryStruct) MetadataMap {
 		quotedRoleNames := GetQuotedRoleNames(connectionPool)
 		currentOid := uint32(0)
 		// Collect all entries for the same object into one ObjectMetadata
+		aclRegex := regexp.MustCompile(`^(.*)=([a-zA-Z\*]*)/(.*)$`)
```
soumyadeep2007: We can have this as a constant at the top of the file, this would remove the need to pass it to parseACL.

khuddlefish: A good idea though Go variable scope is weird if you put it at the top of the file. I'm not sure what the best place to define it would be.

chrishajas: I haven't been able to get this to work since MustCompile is actually doing logic rather than simply declaring a variable.

Additionally, it doesn't seem that regexp can be constants. We could make the string itself a constant, but we'd still have to pass aclRegex into the function.

soumyadeep2007: You can also just define it as a non-const variable at the top of the file and declare it public such that it can also be referenced in the test.

khuddlefish: Yeah you should be able to make it non const. I wasn't sure if it should go in predata_shared or predata_relations. I guess predata_shared because that is where ParseACL is defined.

chrishajas: Thanks, fixed!

backup/predata_relations.go
```diff
@@ -231,7 +229,7 @@ func ConstructColumnPrivilegesMap(results []ColumnPrivilegesQueryStruct) map[uin
 				currentColumn = result.Name
 				columnMetadata = make([]ACL, 0)
 			}
-			privileges := ParseACL(privilegesStr, quotedRoleNames, aclRegex)
+			privileges := ParseACL(privilegesStr, quotedRoleNames, ACLRegex)
```
soumyadeep2007: Do we still need to pass this is in to the function? As it is global..

khuddlefish: No we don't. We should take that out.

backup/predata_shared.go
```diff
@@ -24,6 +24,8 @@ type Schema struct {
 	Name string
 }
 
+var ACLRegex = regexp.MustCompile(`^(.*)=([a-zA-Z\*]*)/(.*)$`)
```
khuddlefish: Probably move this above the struct definition since that is used for the function immediately below.

backup/predata_relations.go
chrishajas: Ah yes, the whole point of making it global in the first place. Fixed.

backup/predata_shared.go
chrishajas: Done.

PR #130, opened 2018-08-08
Modify table validation in restore database

khuddlefish:
For a data only restore, we want to validate that all tables we will
attempt to restore data to already exist in the database.

For a metadata restore, we want to validate that the tables we will
attempt to restore are not already in the database. We were previously
only checking this for a filtered restore.

### DIFF
--- a/restore/validate_test.go
+++ b/restore/validate_test.go
@@ -1,17 +1,13 @@
 package restore_test
 
 import (
-	"regexp"
-
 	"github.com/greenplum-db/gp-common-go-libs/testhelper"
 	"github.com/greenplum-db/gpbackup/restore"
 	"github.com/greenplum-db/gpbackup/testutils"
 	"github.com/greenplum-db/gpbackup/utils"
-	sqlmock "gopkg.in/DATA-DOG/go-sqlmock.v1"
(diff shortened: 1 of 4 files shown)


restore/validate_test.go
```diff
@@ -59,31 +63,140 @@ var _ = Describe("restore/validate tests", func() {
...
+				})
+				It("passes if all tables are present in database", func() {
+					two_table_rows := sqlmock.NewRows([]string{"string"}).
+						AddRow("public.table1").AddRow("public.view1")
+					mock.ExpectQuery("SELECT (.*)").WillReturnRows(two_table_rows)
+					restore.ValidateRelationsInRestoreDatabase(connection, filterList)
+				})
+			})
+		})
+		Context("all stages restore", func() {
```
soumyadeep2007: What is meant by "all stages restore"?

khuddlefish: I wasn't sure what to call it when it is not data-only. I guess I was really thinking of sections since that is how we refer to it in the report file. If you have other ideas, I'd be happy to hear them.

soumyadeep2007: Calling it "Not data only" might be more clear.

PR #131, opened 2018-08-08
Do not fail from COPY error when using --on-error-continue

khuddlefish:
People may want to load as much data as possible and deal with the
errors later. We now log an error and continue with the restore.


khuddlefish:
We're going to modify this with our next story so I will close this in favor of whatever we decide to do.

PR #132, opened 2018-08-09
Multi file plugin

lisaoakley:
Support multi-file with plugin

Also, fix for metadata-only with plugin and refactor of compression branching

### DIFF
--- a/backup/backup.go
+++ b/backup/backup.go
@@ -86,7 +86,7 @@ func DoSetup() {
 	CreateBackupDirectoriesOnAllHosts()
 	globalTOC = &utils.TOC{}
 	globalTOC.InitializeMetadataEntryMap()
-	utils.InitializeCompressionParameters(!MustGetFlagBool(utils.NO_COMPRESSION), MustGetFlagInt(utils.COMPRESSION_LEVEL))
+	utils.InitializePipeThroughParameters(!MustGetFlagBool(utils.NO_COMPRESSION), MustGetFlagInt(utils.COMPRESSION_LEVEL))
 
 	pluginConfigFlag := MustGetFlagString(utils.PLUGIN_CONFIG)
 
--- a/backup/backup.go
+++ b/backup/backup.go
@@ -28,7 +28,7 @@ func initializeFlags(cmd *cobra.Command) {
 
 func SetFlagDefaults(flagSet *pflag.FlagSet) {
 	flagSet.String(utils.BACKUP_DIR, "", "The absolute path of the directory to which all backup files will be written")
-	flagSet.Int(utils.COMPRESSION_LEVEL, 0, "Level of compression to use during data backup. Valid values are between 1 and 9.")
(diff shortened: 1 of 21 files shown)


backup/data.go
```diff
@@ -55,22 +55,25 @@ type BackupProgressCounters struct {
 }
 
 func CopyTableOut(connectionPool *dbconn.DBConn, table Relation, backupFile string, connNum int) int64 {
-	usingCompression, compressionProgram := utils.GetCompressionParameters()
-	copyCommand := ""
+	compressionProgram := utils.GetCompressionProgram()
```
khuddlefish: Maybe add a comment saying that this will be cat if we are not using compression

chrishajas:
Let's add a bit more to describe the compression refactor in the `Support multi file with plugin.` commit too. We now use `cat` as a default compression program to simplify the logic.

backup/data.go
```diff
@@ -55,22 +55,25 @@ type BackupProgressCounters struct {
...
+	checkPipeExistsCommand := ""
+	pluginCommand := ">"
```
khuddlefish: These names are a bit confusing. Maybe we could use names that refer to the order we call them in for the bash command? For example, we start with gzip/cat then pass through the plugin command then write it into the backupfile. So maybe something like initial command, transition command, and destination? Just an idea.

backup/data.go
khuddlefish: I'm a little confused as to why we don't have a top level conditional with instead of setting it as a default in the compression file. That just hides the logic.
```
if !usingCompression{
    compressionProgram = "cat -"
}
```

backup/backup.go
```diff
@@ -28,7 +28,7 @@ func initializeFlags(cmd *cobra.Command) {
 
 func SetFlagDefaults(flagSet *pflag.FlagSet) {
 	flagSet.String(utils.BACKUP_DIR, "", "The absolute path of the directory to which all backup files will be written")
-	flagSet.Int(utils.COMPRESSION_LEVEL, 0, "Level of compression to use during data backup. Valid values are between 1 and 9.")
+	flagSet.Int(utils.COMPRESSION_LEVEL, 1, "Level of compression to use during data backup. Valid values are between 1 and 9.")
```
khuddlefish: Have we verified that we are still able to tell the difference between someone passing this flag with the default value and not passing the flag? Want to make sure we can still detect if they try to pass no-compression and compression-level together.

chrishajas: Yes, confirmed.

backup/data.go
chrishajas: Modified the names to be more descriptive:
```
copyCommand := fmt.Sprintf("PROGRAM '%s%s %s %s'", checkPipeExistsCommand, customPassThroughCommand, sendToDestinationCommand, destinationToWrite)
```

khuddlefish:
Merged

PR #144, opened 2018-09-11
Fix gprestore properly restore filtered leaf partitions for incremental backups

chrishajas:
Previously, gprestore would fail in validation if a user attempted a
filtered restore of a partition that was not in the most recent restore
plan entry. Now, gprestore will look through all restore plan entries to
determine whether a partition leaf is present in the restore set.

Additionally, when restoring a leaf partition, the root partition table
schema would not get restored if it no leaf partitions were in the most
recent restore plan entry. Now, we iterate through restore plans to and
add all roots to the include list.

### DIFF
--- a/restore/validate_test.go
+++ b/restore/validate_test.go
@@ -36,23 +36,23 @@ var _ = Describe("restore/validate tests", func() {
 			toc.AddPredataEntry("schema", "somesequence", "SEQUENCE", "", table1Len+table2Len, backupfile)
 			restore.SetTOC(toc)
 		})
-		It("schema exists in normal backup", func() {
+		It("passes when schema exists in normal backup", func() {
 			restore.SetBackupConfig(&utils.BackupConfig{})
 			filterList = []string{"schema1"}
 			restore.ValidateFilterSchemasInBackupSet(filterList)
 		})
-		It("schema does not exist in normal backup", func() {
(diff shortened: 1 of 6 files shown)


kyeap-pivotal:
Looks good to me, there might be a typo in the third line of the metadata commit. 

```
schema would not get restored if it no leaf partitions were in the most
```

restore/validate_test.go
```diff
@@ -228,10 +228,18 @@ var _ = Describe("restore/validate tests", func() {
 			defer testhelper.ShouldPanicWithMessage("Could not find the following relation(s) in the backup set: schema1.table3")
 			restore.ValidateFilterRelationsInBackupSet(filterList)
 		})
+		It("table exists in most recent restore plan entry", func() {
```
khuddlefish: I just realized the labels for these tests don't fit with the "It" pattern. It should be something like "It(passes when table exists in most recent restore plan entry)" and "It(passes when table exists in previous restore plan entry)"

chrishajas: Fixed

chrishajas:
> Looks good to me, there might be a typo in the third line of the metadata commit.
> 
> ```
> schema would not get restored if it no leaf partitions were in the most
> ```

Thanks, Fixed!

PR #150, opened 2018-09-20
Add support for Foreign Tables

chrishajas:
Foreign tables were introduced in GPDB in the Postgres 9.1 merge.

Foreign tables have a similar schema structure to regular tables, but do
not contain any data within GPDB. Therefore, while we need to generate
the metadata for foreign tables, we do not dump or restore any data.

Additionally, Exclude foreign tables from GetType functions

Foreign Data Wrappers can be dependent on functions, so FDWs, servers, user mappings, and foreign tables will need to be added to the dependency sorting logic. This is not included in this PR.

Co-authored-by: Nadeem Ghani <nghani@pivotal.io>


backup/backup.go
```diff
@@ -235,6 +235,14 @@ func backupPredata(metadataFile *utils.FileWithByteCount, tables []Relation, tab
 		BackupEnumTypes(metadataFile, typeMetadata)
 	}
 
+	if len(MustGetFlagStringSlice(utils.INCLUDE_SCHEMA)) == 0 {
+		if connectionPool.Version.AtLeast("6") {
```
khuddlefish: Do any of these objects depend on other objects that were previously being dumped before them? AKA sequences, constraints, protocols, types, functions or tables?

chrishajas: Yes. Foreign Data Wrappers can depend on functions. So all of these objects will need to be added to dependency sorting logic.

PR #151, opened 2018-09-22
Redoing dependencies

nadeemg:


backup/dependencies.go
```diff
@@ -152,11 +107,97 @@ func TopologicalSort(slice []Sortable) []Sortable {
...
+
+type DependencyMap map[DepEntry]map[DepEntry]bool
+
+type DepEntry struct {
+	Classid uint32
+	Objid   uint32
+}
+
+// This function only returns dependencies that are referenced in the backup set
+func GetDependencies(connectionPool *dbconn.DBConn, backupSet map[DepEntry]bool) DependencyMap {
```
chrishajas: Would be good to include a comment about the different cases this handles (we're taking the union of two queries which I assume captures 2 different cases). Additionally, should mention what id1/id2 are capturing.

backup/queries_relations.go
```diff
@@ -47,18 +47,21 @@ func GetAllUserTables(connection *dbconn.DBConn) []Relation {
...
-	Oid         uint32
-	Schema      string
-	Name        string
-	DependsUpon []string // Used for dependency sorting
-	Inherits    []string // Only used for printing INHERITS statement
+	SchemaOid uint32
+	Oid       uint32
+	Schema    string
+	Name      string
+	Inherits  []string // Only used for printing INHERITS statement
```
jmcatamney: Nitpick: This comment can probably be removed, as it was only necessary because `DependsUpon` and `Inherits` were named similarly but used in different ways.

backup/dependencies.go
```diff
@@ -152,11 +107,97 @@ func TopologicalSort(slice []Sortable) []Sortable {
 		gplog.Verbose("Failed to sort dependencies.")
 		gplog.Verbose("Not yet visited:")
 		for _, item := range slice {
-			if notVisited.MatchesFilter(item.FQN()) {
-				gplog.Verbose("Object: %s; Dependencies: %s", item.FQN(), item.Dependencies())
+			if notVisited[item.GetDepEntry()] {
+				gplog.Verbose("Object: %s; TODO:ADD DEPENDENCIES", item.FQN())
```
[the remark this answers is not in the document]
khuddlefish: We actually fixed this and forgot to push it

backup/dependencies.go
khuddlefish: We added a comment for the union stuff on Friday, but forgot to push. We could also add comments for id1 and id2 if you think it would be useful

integration/predata_relations_queries_test.go
```diff
@@ -1128,24 +1128,23 @@ SET SUBPARTITION TEMPLATE
 		child := backup.Relation{Schema: "public", Name: "child"}
```
kyeap-pivotal: set of tests no longer testing ConstructTableDependencies, change Describe to GetTableInheritance

khuddlefish: fixed

PR #152, opened 2018-09-22
Do not backup indexes that are not valid or not ready

chrishajas:


khuddlefish:
I assume this is fine, but where did we find the info about these fields?

PR #156, opened 2018-10-03
Fix backing up and restoring database names containing special characters

chrishajas:
See individual commit messages for details.

### DIFF
--- a/integration/metadata_globals_queries_test.go
+++ b/integration/metadata_globals_queries_test.go
@@ -40,16 +40,19 @@ var _ = Describe("backup integration tests", func() {
 	})
 	Describe("GetDatabaseInfo", func() {
 		It("returns a database info struct", func() {
-			if connectionPool.Version.Before("6") {
-				testhelper.AssertQueryRuns(connectionPool, "CREATE TABLESPACE test_tablespace FILESPACE test_dir")
-			} else {
-				testhelper.AssertQueryRuns(connectionPool, "CREATE TABLESPACE test_tablespace LOCATION '/tmp/test_dir'")
-			}
-			defer testhelper.AssertQueryRuns(connectionPool, "DROP TABLESPACE test_tablespace")
-
 			result := backup.GetDatabaseInfo(connectionPool)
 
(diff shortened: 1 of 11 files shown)


integration/metadata_globals_queries_test.go
```diff
@@ -40,18 +40,21 @@ var _ = Describe("backup integration tests", func() {
 	})
 	Describe("GetDatabaseInfo", func() {
 		It("returns a database info struct", func() {
-			if connectionPool.Version.Before("6") {
```
khuddlefish: do we not need this anymore?

utils/io_test.go
```diff
@@ -13,6 +13,38 @@ import (
...
+
+			Expect(resultString).To(Equal(`test`))
+		})
+		It("returns single quote when passed a single quote", func() {
+			dbname := `"`
+			resultString := utils.UnquoteIdent(dbname)
+
+			Expect(resultString).To(Equal(`"`))
+		})
+		It("returns empty string when passed and empty string", func() {
```
khuddlefish: passed **an** empty string. (Don't worry about this if you aren't changing other things though)

utils/io_test.go
```diff
@@ -13,6 +13,38 @@ import (
 )
 
 var _ = Describe("utils/io tests", func() {
+	Describe("UnquoteIdent", func() {
```
khuddlefish: Should there be some test case for things containing single quotes? `'`

integration/metadata_globals_queries_test.go
chrishajas: We're creating that tablespace `test_tablespace`, but backing up `testdb`, which using `pg_default`. So `test_tablespace` isn't being used at all here.

utils/io_test.go
chrishajas: quote_ident doesn't care about single quotes, they don't need to be escaped or treated differently.

chrishajas:
@khuddlefish I agree, we'll change the arg names so things are clearer.

PR #158, opened 2018-10-04
Handle aggregate sortop

nadeemg:


backup/queries_functions.go
```diff
@@ -305,7 +305,7 @@ SELECT
 	a.aggtransfn::regproc::oid,
 	a.aggprelimfn::regproc::oid,
 	a.aggfinalfn::regproc::oid,
-	a.aggsortop::regproc::oid,
```
khuddlefish: Are we worried about this not being fully qualified?

integration/predata_functions_queries_test.go
```diff
@@ -240,6 +240,31 @@ CREATE FUNCTION public.mypre_accum(numeric, numeric)
 			testhelper.AssertQueryRuns(connectionPool, "DROP FUNCTION public.mysfunc_accum(numeric, numeric, numeric)")
 			testhelper.AssertQueryRuns(connectionPool, "DROP FUNCTION public.mypre_accum(numeric, numeric)")
 		})
+		It("creates and aggregate with sortop in pg_catalog", func() {
```
khuddlefish: creates **an** aggregate

backup/queries_functions.go
chrishajas: As far as I can tell, casting to regoper should fully qualify the operator name.

PR #159, opened 2018-10-04
Fix type storage being printed in backup

chrishajas:
We were previously using the wrong values for storage type 'e' and 'x'.
We are now using the correct values. 'e' is external, and 'x' is
extended.

This would have been caught if we used create tests for all the
permutations, but I don't think we chose to do that due to the
overhead. Unfortunately, this one slipped through.

### DIFF
--- a/backup/predata_types.go
+++ b/backup/predata_types.go
@@ -95,11 +95,11 @@ func PrintCreateBaseTypeStatement(metadataFile *utils.FileWithByteCount, toc *ut
 	if base.Storage != "" {
 		switch base.Storage {
 		case "e":
-			metadataFile.MustPrintf(",\n\tSTORAGE = extended")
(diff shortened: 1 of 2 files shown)


jimmyyih:
Just interested, what are the repercussions of this?

chrishajas:
@jimmyyih This is the storage strategy strategy for a type. So if a user created a base type with extended, and then created a table with a column using that type, it would have been stored as external.

From https://www.postgresql.org/docs/9.5/static/sql-createtype.html:

plain specifies that data of the type will always be stored in-line and not compressed. extended specifies that the system will first try to compress a long data value, and will move the value out of the main table row if it's still too long. external allows the value to be moved out of the main table, but the system will not try to compress it. 

This isn't external in the external table sense, it's external in the on disk-format mentioned above. The repercussion I see is data possibly being compressed/uncompressed unintentionally on disk.

PR #160, opened 2018-10-05
Refactor backup predata functionality to remove duplication

nadeemg:
Previously, similar functionality was included in the predata backup for
all objects and predata backup for only relations. This commit combines
the two functions.

Co-authored-by: Nadeem Ghani <nghani@pivotal.io>


backup/wrappers.go
```diff
@@ -479,7 +479,7 @@ func BackupDependentObjects(metadataFile *utils.FileWithByteCount, tables []Rela
```
khuddlefish: The log statements above aren't really correct anymore. We should make a chore to clean them up

nadeemg: For now, just reduced it to one generic log message

PR #162, opened 2018-10-09
Only select database config that is not role specific

khuddlefish:
The pg_db_role_setting table was introduced in PG 9.0 which allows for
role-specific database settings instead of having a single datconfig
column in pg_database. setrole = 0 in this new table indicates settings
that are not role specific, so those are the only ones we want to
select.


chrishajas:
Do we want to add an ALTER statement in the `GetDatabaseGUCs` test scenario in `metadata_globals_queries_test` that would fail without this change?

khuddlefish:
@chrishajas Probably a good idea

kyeap-pivotal:
Test added in metadata_globals_queries_test.

PR #164, opened 2018-10-09
Update pipeline to use Data Domain hosted on AWS

kyeap-pivotal:
Check the commit messages for details.

### DIFF
--- a/end_to_end/end_to_end_suite_test.go
+++ b/end_to_end/end_to_end_suite_test.go
@@ -27,6 +27,10 @@ import (
 	"github.com/onsi/gomega/gexec"
 )
 
+/* The backup directory must be unique per test. There is test flakiness
+ * against Data Domain Boost mounted file systems due to how it handles
+ * directory deletion/creation.
+ */
 var custom_backup_dir string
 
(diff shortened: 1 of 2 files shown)


end_to_end/end_to_end_suite_test.go
```diff
@@ -398,7 +398,7 @@ var _ = Describe("backup end to end integration tests", func() {
 				os.RemoveAll(backupdir)
 			})
 			It("runs gpbackup and gprestore with single-data-file flag without compression", func() {
-				backupdir := filepath.Join(custom_backup_dir, "single_data_file")
+				backupdir := filepath.Join(custom_backup_dir, "single_data_file_no_compression")
```
khuddlefish: I assume these names were changed so we won't use the same dir names for different tests? Should we make that more clear somewhere so we don't mess it up when we add a new test in the future or do we think it will be ok?

chrishajas: I don't feel like a single comment at the top would keep someone from making this mistake. We added the comment `Must be unique` where each backupdir is set. It's a bit verbose, but is probably for the best.

PR #165, opened 2018-10-11
Log stack trace for panics that don't come from gplog

kyeap-pivotal:
* Create error file when helper panics to signal to gpbackup and
  gprestore an error has occurred
* Check for agent error file in backup
* Return error in CheckRowsRestored instead of panicking to avoid
  goroutine stacktrace

### DIFF
--- a/backup/backup.go
+++ b/backup/backup.go
@@ -3,6 +3,7 @@ package backup
 import (
 	"fmt"
 	"os"
+	"runtime/debug"
 	"sync"
 	"time"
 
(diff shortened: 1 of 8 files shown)


helper/helper.go
```diff
@@ -369,7 +371,15 @@ func removeFileIfExists(filename string) {
...
 
 func DoTeardown() {
-	_ = recover()
+	if err := recover(); err != nil {
+		// Check if gplog.Fatal did not cause the panic
+		if gplog.GetErrorCode() != 2 {
+			gplog.Error(fmt.Sprintf("%v: %s", err, debug.Stack()))
+			gplog.SetErrorCode(2)
+		}
+		handle := iohelper.MustOpenFileForWriting(fmt.Sprintf("%s_error", *pipeFile))
```
chrishajas: We do create an error file during backup if an error occurs. Do we need both still?

khuddlefish: This one covers more cases. We could probably remove the other one

backup/backup.go
```diff
@@ -368,7 +369,13 @@ func DoTeardown() {
 
 	errStr := ""
 	if err := recover(); err != nil {
-		errStr = fmt.Sprintf("%v", err)
+		// Check if gplog.Fatal did not cause the panic
+		if gplog.GetErrorCode() != 2 {
```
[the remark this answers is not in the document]
khuddlefish: Maybe a constant defined in gplog?

helper/helper.go
khuddlefish: Removed extra error file case because we don't need it anymore

khuddlefish:
Merged

PR #171, opened 2018-10-20
Remove fatals from helper

khuddlefish:
Modify helper functions to return errors instead of calling gplog.Fatal. Also does some refactoring. We kept the refactoring in separate commits so it is easier to review.


restore/data.go
```diff
@@ -139,18 +139,22 @@ func restoreDataFromTimestamp(fpInfo utils.FilePathInfo, dataEntries []utils.Mas
...
 	close(tasks)
 	workerPool.Wait()
+
+	var agentErr error
+	if backupConfig.SingleDataFile {
+		agentErr = utils.CheckAgentErrorsOnSegments(globalCluster, globalFPInfo)
+		if agentErr != nil {
+			if MustGetFlagBool(utils.ON_ERROR_CONTINUE) || fatalErr != nil {
+				gplog.Error(agentErr.Error())
+			} else if !MustGetFlagBool(utils.ON_ERROR_CONTINUE) {
```
chrishajas: Making this an else statement would be easier to read I think.

khuddlefish:
Merged

PR #174, opened 2018-10-25
Add support for backing up ALTER DEFAULT PRIVILEGES

chrishajas:
See individual commit messages. I should have added the tests before separating files/logic looking back, but review commit by commit instead of all at once.


backup/predata_acl.go
```diff
@@ -457,33 +457,35 @@ func ConstructDefaultPrivileges(results []DefaultPrivilegesQueryStruct) []Defaul
...
-		if result.Kind == "Empty" {
-			privilegesStr = "GRANTEE=/GRANTOR"
-		} else if result.Privileges.Valid {
-			privilegesStr = result.Privileges.String
-		}
-		if result.Oid != currentOid {
-			if currentOid != 0 {
-				priv.Privileges = sortACLs(priv.Privileges)
-				defaultPrivileges = append(defaultPrivileges, priv)
+	if len(results) > 0 {
```
khuddlefish: shouldn't the for loop already handle the case when len(results) == 0? It will just terminate immediately because there's nothing in it

chrishajas: In the empty case, it would end up populating the defaultPrivileges slice with 		
`defaultPrivileges = append(defaultPrivileges, priv)`. defaultPrivileges would then have a length of 1 even though no default privileges exist.

backup/predata_shared.go
```diff
@@ -501,3 +510,52 @@ func PrintDependentObjectStatements(metadataFile *utils.FileWithByteCount, toc *
...
+			objectType = "TABLE"
+		case "S":
+			objectType = "SEQUENCE"
+		case "f":
+			objectType = "FUNCTION"
+		case "T":
+			objectType = "TYPE"
+		}
+		alterPrefix := fmt.Sprintf("ALTER DEFAULT PRIVILEGES%s%s", roleStr, schemaStr)
+		statements = append(statements, fmt.Sprintf("%s REVOKE ALL ON %sS FROM PUBLIC;", alterPrefix, objectType))
```
nadeemg: Should that `%sS` be just `%s`? Or else, in line 537, should `PRIVILEGES` be just `PRIVILEGE`?

backup/queries_acl.go
```diff
@@ -0,0 +1,269 @@
...
+	query := fmt.Sprintf(`
+SELECT
+	'%s'::regclass::oid AS classid,
+	o.%s AS oid,
+	coalesce(description,'') AS comment
+	FROM %s o JOIN %s d ON (d.objoid = %s AND d.classoid = '%s'::regclass%s)%s;
+`, params.CatalogTable, params.OidField, params.CatalogTable, descTable, params.OidField, commentTable, subidStr, schemaStr)
+
+	results := make([]struct {
+		UniqueID
```
nadeemg: Is UniqueID mapping to the classid in the query or the oid?

khuddlefish: UniqueID is a nested struct that contains Oid and Classid. sqlx is smart enough to be able to select those two properties into the nested fields.

khuddlefish: We also didn't change this code, just moved it

backup/predata_acl.go
kyeap-pivotal: It might be better to check the inverse since it seems like the amount of indenting is starting to affect readability.
```
if len(results) == 0 {
    return []DefaultPrivileges{}
}
```
same for the function `ConstructMetadataMap`.

chrishajas: Good idea, fixed.

PR #175, opened 2018-10-25
Do not dump DB encoding info if it is the same as defaults

khuddlefish:
This encoding info includes
* LC_COLLATE
* LC_CTYPE
* ENCODING
Since this information is often platform specific, not dumping it unless
necessary will make it easier to port database statements across
platforms.

Also adds integration tests for PrintCreateDatabaseStatement


khuddlefish:
Passed the test pipeline

PR #176, opened 2018-10-25
Support restoring database with read only GUC set

chrishajas:
If a database has a ALTER DATABASE SET default_transaction_read_only
setting, restore would fail since the database would not allow writing.
Instead, set this GUC off during the restore session. This GUC was
introduced in GPDB6.


khuddlefish:
Would the GUCs for the database override this, or is it a DB level vs. session level thing?

chrishajas:
Session level will take precedent. Even if the database/user/cluster has this GUC set, session level will override for this session.

PR #177, opened 2018-10-26
Fix backup filter validation inconsistencies

kyeap-pivotal:
When backing up with a nonexistent table and schema using the
--exclude-table/--exclude-schema flag gpbackup would fatal. We now allow
this by removing the validation of the --exclude-table and
--exclude-schema flag as users may want to exclude tables/schemas that
may or may not exist at the time of backup.

When backing up with filtered schemas or tables that do not exist in the
database using --include-schema or --include-table, gpbackup will fail
to validate. This behavior is not intended. Now we properly validate
filter options

Additionally, log warnings when a table/schema does not exist with an --exclude flag.

Co-Authored-by: Kevin Yeap <kyeap@pivotal.io>
Co-Authored-by: Chris Hajas <chajas@pivotal.io>

### DIFF
--- a/backup/validate.go
+++ b/backup/validate.go
@@ -15,11 +15,13 @@ import (
  */
 
 func validateFilterLists() {
-	ValidateFilterSchemas(connectionPool, MustGetFlagStringSlice(utils.INCLUDE_SCHEMA))
-	ValidateFilterTables(connectionPool, MustGetFlagStringSlice(utils.INCLUDE_RELATION))
+	ValidateFilterSchemas(connectionPool, MustGetFlagStringSlice(utils.INCLUDE_SCHEMA), false)
+	ValidateFilterSchemas(connectionPool, MustGetFlagStringSlice(utils.EXCLUDE_SCHEMA), true)
+	ValidateFilterTables(connectionPool, MustGetFlagStringSlice(utils.INCLUDE_RELATION), false)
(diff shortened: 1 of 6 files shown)


backup/validate.go
```diff
@@ -15,29 +15,34 @@ import (
...
-	ValidateFilterTables(connectionPool, MustGetFlagStringSlice(utils.EXCLUDE_RELATION))
-	ValidateFilterTables(connectionPool, MustGetFlagStringSlice(utils.INCLUDE_RELATION))
+	ValidateFilterSchemas(connectionPool, MustGetFlagStringSlice(utils.INCLUDE_SCHEMA), false)
+	ValidateFilterSchemas(connectionPool, MustGetFlagStringSlice(utils.EXCLUDE_SCHEMA), true)
+	ValidateFilterTables(connectionPool, MustGetFlagStringSlice(utils.INCLUDE_RELATION), false)
+	ValidateFilterTables(connectionPool, MustGetFlagStringSlice(utils.EXCLUDE_RELATION), true)
 }
 
-func ValidateFilterSchemas(connectionPool *dbconn.DBConn, schemaList []string) {
+func ValidateFilterSchemas(connectionPool *dbconn.DBConn, schemaList []string, noFatal bool) {
```
khuddlefish: I'm not sure the name "noFatal" really communicates what we're doing here. We're specifically not fatalling because we're excluding.

PR #178, opened 2018-10-29
Refactor set implementation

chrishajas:
Previously, we were using sets for two purposes:
  1) As an include or exclude set to match table/schemas that were to be
  included or excluded from the backup or restore
  2) As a regular set

The first case uses a modified implementation, where an empty set would
always return that a object exists. This was done so that the default
behavior would backup/restore all objects. However, in cases where a
generic set without the property was wanted, it could lead to confusing
behavior.

This introduces a NewSet constructor and replaces instances that should
have been using a regular set with this call.

Additionally, this removes the Add and Delete set functionality since
they weren't being used.

### DIFF
--- a/backup/predata_relations.go
+++ b/backup/predata_relations.go
@@ -64,10 +64,8 @@ func SplitTablesByPartitionType(tables []Relation, tableDefs map[uint32]TableDef
 				if partType != "p" && partType != "i" {
 					dataTables = append(dataTables, table)
 				}
-			} else if len(includeList) > 0 {
-				if includeSet.MatchesFilter(table.FQN()) {
-					dataTables = append(dataTables, table)
-				}
+			} else if includeSet.MatchesFilter(table.FQN()) {
+				dataTables = append(dataTables, table)
 			}
 		}
 	} else {
--- a/backup/predata_relations.go
+++ b/backup/predata_relations.go
@@ -50,7 +50,7 @@ func SplitTablesByPartitionType(tables []Relation, tableDefs map[uint32]TableDef
 	metadataTables := make([]Relation, 0)
 	dataTables := make([]Relation, 0)
 	if MustGetFlagBool(utils.LEAF_PARTITION_DATA) || len(includeList) > 0 {
-		includeSet := utils.NewIncludeSet(includeList)
+		includeSet := utils.NewSet(includeList)
 		for _, table := range tables {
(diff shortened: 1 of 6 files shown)


backup/predata_relations.go
```diff
@@ -50,7 +50,7 @@ func SplitTablesByPartitionType(tables []Relation, tableDefs map[uint32]TableDef
 	metadataTables := make([]Relation, 0)
 	dataTables := make([]Relation, 0)
 	if MustGetFlagBool(utils.LEAF_PARTITION_DATA) || len(includeList) > 0 {
-		includeSet := utils.NewIncludeSet(includeList)
+		includeSet := utils.NewSet(includeList)
```
khuddlefish: If you change this then we should probably change line 67 to not check `if len(includeList) > 0` right? Wouldn't that we the whole point of changing this?

chrishajas: I was viewing it from the perspective of `NewIncludeSet` in the previous case could never be empty, so `AlwaysMatchesFilter` could never be set, therefore, it's safe to use a regular set rather than include set. I don't think we can remove `if len(includeList) > 0` without more work as unit tests start failing.

khuddlefish: I think the whole reason we were checking for len(includeList) > 0 is because of the weird behavior of the previous includeSet. We probably could have alternatively said 
```
includeSet := utils.NewIncludeSet(includeList)
includeSet.AlwaysMatchesFilter = false
```

khuddlefish: I'm having trouble thinking right now about cases where we wouldn't want to have the includeSet as long as the implementation does not return true when it is empty, but it does seem like this is exactly the kind of case where we could simplify logic that was weird because of the previously implementation.

chrishajas: Oh, I didn't read and thought you meant line 52. Yes, we can change line 67 to be an `else` instead of `else if`.

khuddlefish: No worries. Yeah, looks like we still need the logic on line 52, but we should change 
```
} else if len(includeList) > 0 {
    if includeSet.MatchesFilter(table.FQN()) {
        dataTables = append(dataTables, table)
    }
}
```
to be
```
} else if includeSet.MatchesFilter(table.FQN()) {
    dataTables = append(dataTables, table)
}
```

khuddlefish: Or I guess the
```
else {
    if ... {
 ```
is ok too, the other is just a bit more compressed

chrishajas: Fixed it to the `else if`.

PR #179, opened 2018-10-29
Fix on-error-continue case for restoreSchemas

khuddlefish:
We were not failing on errors if the schema already existed, but we were
failing if we encountered some other error. Now we will log and continue
if --on-error-continue is set.

### DIFF
--- a/restore/restore_suite_test.go
+++ b/restore/restore_suite_test.go
@@ -50,13 +50,10 @@ func TestRestore(t *testing.T) {
 
 var cmdFlags *pflag.FlagSet
 
-var _ = BeforeSuite(func() {
-	connectionPool, mock, stdout, stderr, logfile = testutils.SetupTestEnvironment()
-	buffer = gbytes.NewBuffer()
-})
-
 var _ = BeforeEach(func() {
-	connectionPool, mock = testutils.CreateAndConnectMockDB(1)
+	connectionPool, mock, stdout, stderr, logfile = testutils.SetupTestEnvironment()
+	restore.SetConnection(connectionPool)
+	buffer = gbytes.NewBuffer()
 
 	cmdFlags = pflag.NewFlagSet("gprestore", pflag.ExitOnError)
 	restore.SetCmdFlags(cmdFlags)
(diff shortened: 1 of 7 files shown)


restore/wrappers_test.go
```diff
@@ -35,4 +38,55 @@ var _ = Describe("wrapper tests", func() {
...
+			schemaArray        = []utils.StatementWithType{{Name: "foo", Statement: "create schema foo"}}
+		)
+		BeforeEach(func() {
+			ignoredProgressBar = utils.NewProgressBar(1, "", utils.PB_NONE)
+			ignoredProgressBar.Start()
+		})
+		AfterEach(func() {
+			ignoredProgressBar.Finish()
+		})
+		It("logs nothing if there are no errors", func() {
```
khuddlefish: This test case seems to be flaky

restore/restore_suite_test.go
```diff
@@ -50,12 +50,9 @@ func TestRestore(t *testing.T) {
 
 var cmdFlags *pflag.FlagSet
 
-var _ = BeforeSuite(func() {
-	connectionPool, mock, stdout, stderr, logfile = testutils.SetupTestEnvironment()
-	buffer = gbytes.NewBuffer()
-})
-
 var _ = BeforeEach(func() {
+	_, _, stdout, stderr, logfile = testutils.SetupTestEnvironment()
```
chrishajas: SetupTestEnvironment calls `CreateAndConnectMockDB`

khuddlefish: I saw that, but for some weird reason, I got a bunch of null pointers when I tried to use this call to set connectionPool and mock rather than the call below. We can investigate why that might have been happening to see if we can get that to work though.

nadeemg:
Merged into master

restore/restore_suite_test.go
nadeemg: We figured out there was another copy of CreateAndConnectMockDB with slightly different behavior that was tricking us, but we fixed it.

PR #181, opened 2018-10-31
Add support for security labels

chrishajas:
Security labels were introduced in GPDB6 during the 9.1 merge.

* This also refactors some tests to use the DefaultMetadata and
DefaultMetadataMap functions instead of manually constructing the
metadata
* In some of the create tests, we aren't actually asserting that the
metadata is logically correct, but rather only syntactically correct. Since
the bulk of this code is the same, it's likely ok for now.

I still need to figure out how to get this to run in CI, but it's reviewable in this state.

### DIFF
--- a/backup/predata_acl.go
+++ b/backup/predata_acl.go
@@ -15,9 +15,11 @@ var ACLRegex = regexp.MustCompile(`^(.*)=([a-zA-Z\*]*)/(.*)$`)
  */
 
 type ObjectMetadata struct {
-	Privileges []ACL
-	Owner      string
-	Comment    string
+	Privileges            []ACL
+	Owner                 string
+	Comment               string
+	SecurityLabelProvider string
+	SecurityLabel         string
 }
 
 type ACL struct {
@@ -63,6 +65,9 @@ func PrintObjectMetadata(file *utils.FileWithByteCount, obj ObjectMetadata, obje
 	if privileges := obj.GetPrivilegesStatements(objectName, objectType); privileges != "" {
 		file.MustPrintln(privileges)
 	}
+	if securityLabel := obj.GetSecurityLabelStatement(objectName, objectType); securityLabel != "" {
+		file.MustPrintln(securityLabel)
+	}
 }
 
 func ConstructMetadataMap(results []MetadataQueryStruct) MetadataMap {
@@ -91,6 +96,8 @@ func ConstructMetadataMap(results []MetadataQueryStruct) MetadataMap {
 			metadata.Privileges = make([]ACL, 0)
 			metadata.Owner = result.Owner
 			metadata.Comment = result.Comment
+			metadata.SecurityLabelProvider = result.SecurityLabelProvider
+			metadata.SecurityLabel = result.SecurityLabel
 		}
 
 		privileges := ParseACL(privilegesStr, quotedRoleNames)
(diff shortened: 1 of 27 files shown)


backup/predata_acl.go
```diff
@@ -403,6 +410,19 @@ func (obj ObjectMetadata) GetCommentStatement(objectName string, objectType stri
 	return commentStr
 }
 
+func (obj ObjectMetadata) GetSecurityLabelStatement(objectName string, objectType string) string {
+	providerStr := ""
+	securityLabelStr := ""
+	if obj.SecurityLabelProvider != "" {
```
khuddlefish: I'm pretty sure all security labels have a provider. It is optional to specify one when creating then if you only have one provider loaded, but that just means the label gets that provider. We should probably always specify the provider for correctness.

backup/predata_acl.go
```diff
@@ -91,6 +96,8 @@ func ConstructMetadataMap(results []MetadataQueryStruct) MetadataMap {
 			metadata.Privileges = make([]ACL, 0)
 			metadata.Owner = result.Owner
 			metadata.Comment = result.Comment
+			metadata.SecurityLabelProvider = result.SecurityLabelProvider
+			metadata.SecurityLabelTarget = result.SecurityLabelTarget
```
khuddlefish: Is target the actual label? Is so, can we just call it "SecurityLabel"? label is what postgres seems to call it.

integration/predata_acl_queries_test.go
```diff
@@ -511,7 +563,7 @@ LANGUAGE SQL`)
 				resultMetadataMap := backup.GetMetadataForObjectType(connectionPool, backup.TYPE_TYPE)
 
 				uniqueID := testutils.UniqueIDFromObjectName(connectionPool, "testschema", "testtype", backup.TYPE_TYPE)
-				expectedMetadata := testutils.DefaultMetadata("TYPE", false, true, true)
```
khuddlefish: I thought types can have security labels?

chrishajas: This is just testing that the logic works if the object is in a specific schema, there are already tests above that test it with security labels. The security label logic doesn't have any logic for schema filtering, so I just turned it off for those cases.

khuddlefish: Ok that's fine then

backup/queries_acl.go
```diff
@@ -103,10 +103,12 @@ func InitializeMetadataParams(connectionPool *dbconn.DBConn) {
...
 	UniqueID
-	Privileges sql.NullString
-	Kind       string
-	Owner      string
-	Comment    string
+	Privileges            sql.NullString
+	Kind                  string
+	Owner                 string
+	Comment               string
+	SecurityLabelTarget   string
```
khuddlefish: Remove "Target" here too

backup/queries_relations.go
```diff
@@ -253,21 +253,23 @@ WHERE r.parchildrelid != 0;
...
+	StatTarget            int `db:"attstattarget"`
+	StorageType           string
+	DefaultVal            string
+	Comment               string
+	ACL                   []ACL
+	Options               string
+	FdwOptions            string
+	Collation             string
+	SecurityLabelProvider string
+	SecurityLabelTarget   string
```
khuddlefish: Remove "Target" here too

chrishajas:
@khuddlefish Renamed all instances of SecurityLabelTarget to SecurityLabel. Anything else?

PR #182, opened 2018-10-31
Fix globals

khuddlefish:
We had several problems to fix
    * Dump resource queues/groups and roles before other globals so roles
      will exist when setting owners.
    * Restore CREATE DATABASE statement in the order it was backed up when
      restoring --with-globals so it will not error out for the missing role
      issue.
    * Separate out role gucs from other role information and dump last for
      globals. In gpdb6 users can set role gucs for specific databases so the
      database needs to exist before the guc can be set.

### DIFF
--- a/backup/metadata_globals.go
+++ b/backup/metadata_globals.go
@@ -135,17 +135,15 @@ func PrintCreateResourceGroupStatements(metadataFile *utils.FileWithByteCount, t
 			}
 
 			/* special handling for cpu properties */
+			start = metadataFile.ByteCount
 			if resGroup.CPURateLimit >= 0 {
 				/* cpu rate mode */
-				start = metadataFile.ByteCount
 				metadataFile.MustPrintf("\n\nALTER RESOURCE GROUP %s SET CPU_RATE_LIMIT %d;", resGroup.Name, resGroup.CPURateLimit)
-				toc.AddGlobalEntry("", resGroup.Name, "RESOURCE GROUP", start, metadataFile)
 			} else {
 				/* cpuset mode */
-				start = metadataFile.ByteCount
 				metadataFile.MustPrintf("\n\nALTER RESOURCE GROUP %s SET CPUSET '%s';", resGroup.Name, resGroup.Cpuset)
-				toc.AddGlobalEntry("", resGroup.Name, "RESOURCE GROUP", start, metadataFile)
 			}
+			toc.AddGlobalEntry("", resGroup.Name, "RESOURCE GROUP", start, metadataFile)
 			start = metadataFile.ByteCount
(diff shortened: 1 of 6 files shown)


restore/restore.go
```diff
@@ -170,7 +168,10 @@ func createDatabase(metadataFilename string) {
 }
 
 func restoreGlobal(metadataFilename string) {
-	objectTypes := []string{"SESSION GUCS", "DATABASE GUC", "DATABASE METADATA", "RESOURCE QUEUE", "RESOURCE GROUP", "ROLE", "ROLE GRANT", "TABLESPACE"}
+	objectTypes := []string{"SESSION GUCS", "DATABASE GUC", "DATABASE METADATA", "RESOURCE QUEUE", "RESOURCE GROUP", "ROLE", "ROLE GUCS", "ROLE GRANT", "TABLESPACE"}
```
nadeemg: Would it be easier to understand that the ordering of objectTypes is important if this was an enum instead of a slice?

backup/metadata_globals.go
```diff
@@ -265,15 +279,6 @@ func PrintCreateRoleStatements(metadataFile *utils.FileWithByteCount, toc *utils
 CREATE ROLE %s;
 ALTER ROLE %s WITH %s;`, role.Name, role.Name, strings.Join(attrs, " "))
 
-		for _, roleGUC := range roleGUCs[role.Name] {
```
chrishajas: We didn't have any unit tests for this logic?

backup/metadata_globals.go
```diff
@@ -284,6 +289,20 @@ ALTER ROLE %s WITH %s;`, role.Name, role.Name, strings.Join(attrs, " "))
 	}
 }
 
+func PrintRoleGUCStatements(metadataFile *utils.FileWithByteCount, toc *utils.TOC, roleGUCs map[string][]RoleGUC) {
```
chrishajas: We should add some unit tests to this.

backup/metadata_globals_test.go
```diff
@@ -160,15 +160,12 @@ COMMENT ON RESOURCE QUEUE "commentQueue" IS 'This is a resource queue comment.';
...
-
-ALTER RESOURCE GROUP default_group SET MEMORY_SHARED_QUOTA 25;
-
-ALTER RESOURCE GROUP default_group SET MEMORY_SPILL_RATIO 30;
-
-ALTER RESOURCE GROUP default_group SET CONCURRENCY 15;
-
-ALTER RESOURCE GROUP default_group SET CPU_RATE_LIMIT 10;`)
+			testutils.AssertBufferContents(toc.GlobalEntries, buffer,
+				`ALTER RESOURCE GROUP default_group SET MEMORY_LIMIT 20;`,
```
khuddlefish: Maybe declare these buffer args as a separate variable and pass them in with "..."

PR #183, opened 2018-11-02
Check error on functions in tests

chrishajas:
This checks that the functions under test don't return errors. There
were a few tests that we had missed, and in one case the function was
actually returning an error.

### DIFF
--- a/backup/data_test.go
+++ b/backup/data_test.go
@@ -72,7 +72,9 @@ var _ bool = Describe("backup/data tests", func() {
 			mock.ExpectExec(execStr).WillReturnResult(sqlmock.NewResult(10, 0))
 			filename := "<SEG_DATA_DIR>/backups/20170101/20170101010101/gpbackup_<SEGID>_20170101010101_3456.gz"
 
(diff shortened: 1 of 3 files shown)


backup/data_test.go
```diff
@@ -141,8 +151,9 @@ var _ bool = Describe("backup/data tests", func() {
 			backupFile := fmt.Sprintf("<SEG_DATA_DIR>/gpbackup_<SEGID>_20170101010101_pipe_(.*)_%d", testTable.Oid)
 			copyCmd := fmt.Sprintf(copyFmtStr, backupFile)
 			mock.ExpectExec(copyCmd).WillReturnResult(sqlmock.NewResult(0, 10))
-			backup.BackupSingleTableData(tableDef, testTable, rowsCopiedMap, &counters, 0)
+			err := backup.BackupSingleTableData(tableDef, testTable, rowsCopiedMap, &counters, 0)
```
khuddlefish: You could use Succeed() for this one since the err is the only return value.

PR #184, opened 2018-11-05
Separate history and filepath into their own packages

khuddlefish:
This will make it easier to import this functionality into
gpbackup_manager

Golang doesn't like underscores in package names, so we will try to change the name at some point.

See commits for additional info.


khuddlefish:
I had an idea about where might be a good place for these packages. While you can't have more than one package per directory, you can nest packages at different depths. I think it could be good to put them in gp-common-go-libs, but maybe create a `backup` directory for packages more specifically related to backup. Then we could shorten their names to history/config and filepath. The import would be `github.com/greenplum-db/gp-common-go-libs/backup/filepath`. What do you think?

nadeemg:
> I had an idea about where might be a good place for these packages. While you can't have more than one package per directory, you can nest packages at different depths. I think it could be good to put them in gp-common-go-libs, but maybe create a `backup` directory for packages more specifically related to backup. Then we could shorten their names to history/config and filepath. The import would be `github.com/greenplum-db/gp-common-go-libs/backup/filepath`. What do you think?

SGTM. At this point the code in backup_filepath is being used by gpbackup, gprestore and gpbackup_manager, which seems like enough to call it common. @jmcatamney do you agree that this fits in gp-common-go-libs?

jmcatamney:
On the one hand, our criterion for what goes into gp-common-go-lib` is "Is it used by, or would obviously be useful for, at least two different utilities?", so by that metric it would fit.

On the other hand, backup and backup manager are very tightly coupled and aren't really "separate utilities" in the way that gpbackup and gpupgrade are.  Also, backup file changes are likely to happen much more frequently than other changes to common libraries, at least while gpbackup_manager is in the initial development stage, so pinning the changes only to gpbackup and not gp-common-go-libs would be easier to work with.  Putting them in gp-common-go-libs, meanwhile, would probably lead to other utilities either updating frequently for no actual benefit when the repo is updated or getting lax about updates since they don't immediately matter and slowly drifting out of sync, neither of which is particularly desirable.

I think leaving them in the gpbackup repo would probably be better.  You could still change the names to just `history` and `filepath`; it's obvious what kind of history and filepaths they're dealing with because they're in the gpbackup repo, and they're just as useful for both backup and restore as `utils` is even if you're not actively importing them in gprestore at the moment.

chrishajas:
I agree with Jamie's points. Although this is being used by multiple utilities, it's still specific to backup. No other team would have a use for these packages. Is there a concern with importing packages from gpbackup in gpbackup_manager?

PR #186, opened 2018-11-09
Refactor NewHistory

nadeemg:
The backup history file is now being used by gpbackup_manager as well as gpbackup. This PR makes it easier to use the backup history file consistently from both utilities. See commit messages for details.

### DIFF
--- a/backup/incremental.go
+++ b/backup/incremental.go
@@ -37,7 +37,8 @@ func GetTargetBackupTimestamp() string {
 }
 
 func GetLatestMatchingBackupTimestamp() string {
-	history := backup_history.NewHistory(globalFPInfo.GetBackupHistoryFilePath())
+	history, err := backup_history.NewHistory(globalFPInfo.GetBackupHistoryFilePath())
+	gplog.FatalOnError(err)
 	latestMatchingBackupHistoryEntry := GetLatestMatchingBackupConfig(history, &backupReport.BackupConfig)
 	if latestMatchingBackupHistoryEntry == nil {
 		gplog.FatalOnError(errors.Errorf("There was no matching previous backup found with the flags provided. " +
--- a/backup/incremental.go
+++ b/backup/incremental.go
@@ -38,14 +38,15 @@ func GetTargetBackupTimestamp() string {
 }
 
 func GetLatestMatchingBackupTimestamp() string {
-	history := &backup_history.History{BackupConfigs: make([]backup_history.BackupConfig, 0)}
+	var history *backup_history.History
+	var latestMatchingBackupHistoryEntry *backup_history.BackupConfig
 	var err error
-
 	if iohelper.FileExistsAndIsReadable(globalFPInfo.GetBackupHistoryFilePath()) {
 		history, err = backup_history.NewHistory(globalFPInfo.GetBackupHistoryFilePath())
 		gplog.FatalOnError(err)
+		latestMatchingBackupHistoryEntry = GetLatestMatchingBackupConfig(history, &backupReport.BackupConfig)
 	}
-	latestMatchingBackupHistoryEntry := GetLatestMatchingBackupConfig(history, &backupReport.BackupConfig)
+
 	if latestMatchingBackupHistoryEntry == nil {
 		gplog.FatalOnError(errors.Errorf("There was no matching previous backup found with the flags provided. " +
 			"Please take a full backup."))
--- a/backup/incremental.go
+++ b/backup/incremental.go
@@ -2,6 +2,7 @@ package backup
 
 import (
 	"github.com/greenplum-db/gp-common-go-libs/gplog"
+	"github.com/greenplum-db/gp-common-go-libs/iohelper"
 	"github.com/greenplum-db/gpbackup/backup_history"
 	"github.com/greenplum-db/gpbackup/utils"
 	"github.com/pkg/errors"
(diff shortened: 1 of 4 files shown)


backup_history/history.go
```diff
@@ -91,7 +93,13 @@ func WriteBackupHistory(historyFilePath string, currentBackupConfig *BackupConfi
...
 
-	history := NewHistory(historyFilePath)
+	history := &History{BackupConfigs: make([]BackupConfig, 0)}
+	var err error
+
+	if iohelper.FileExistsAndIsReadable(historyFilePath) {
+		history, err = NewHistory(historyFilePath)
+		gplog.FatalOnError(err)
+	}
 	if len(history.BackupConfigs) == 0 {
```
khuddlefish: could this just be `else {` for the `if FileExistsAndIsReadable`?

backup/incremental.go
```diff
@@ -37,7 +38,13 @@ func GetTargetBackupTimestamp() string {
 }
 
 func GetLatestMatchingBackupTimestamp() string {
-	history := backup_history.NewHistory(globalFPInfo.GetBackupHistoryFilePath())
+	history := &backup_history.History{BackupConfigs: make([]backup_history.BackupConfig, 0)}
```
khuddlefish: maybe putting this in an `else {` for the if below would make the usage a bit more clear? Otherwise, it sort of looks like a mistake that you aren't doing anything if the file doesn't exist.

backup/incremental_test.go
```diff
@@ -192,4 +198,49 @@ var _ = Describe("backup/incremental tests", func() {
...
+			operating.System.Stat = func(string) (os.FileInfo, error) { return nil, nil }
+			operating.System.OpenFileRead = func(string, int, os.FileMode) (operating.ReadCloserAt, error) { return nil, nil }
+			operating.System.ReadFile = func(string) ([]byte, error) { return nil, errors.New("read error") }
+
+			Expect(func() { backup.GetLatestMatchingBackupTimestamp() }).Should(Panic())
+			Expect(log).Should(gbytes.Say("read error"))
+
+		})
+		It("fatals when gpbackup_history.yaml is an invalid format", func() {
+			_, _, log := testhelper.SetupTestLogger()
```
khuddlefish: Some of these could probably be in a BeforeEach. Maybe you could group the tests that do similar setup in a Context() block so you can do that?

backup/incremental_test.go
```diff
@@ -192,4 +198,49 @@ var _ = Describe("backup/incremental tests", func() {
...
+			backup.SetFPInfo(backup_filepath.FilePathInfo{UserSpecifiedBackupDir: "/tmp", UserSpecifiedSegPrefix: "/kevin"})
+			backup.SetReport(&utils.Report{})
+			_, _, log := testhelper.SetupTestLogger()
+
+			Expect(func() { backup.GetLatestMatchingBackupTimestamp() }).Should(Panic())
+			Expect(log.Contents()).To(ContainSubstring("There was no matching previous backup found with the flags provided. Please take a full backup."))
+
+		})
+		It("fatals when gpbackup_history.yaml can't be read", func() {
+			_, _, log := testhelper.SetupTestLogger()
```
khuddlefish: It seems like some of these tests might be overlapping with tests for New History.

backup_history/history.go
```diff
@@ -91,7 +93,13 @@ func WriteBackupHistory(historyFilePath string, currentBackupConfig *BackupConfi
 		_ = lock.Unlock()
 	}()
 
-	history := NewHistory(historyFilePath)
+	history := &History{BackupConfigs: make([]BackupConfig, 0)}
```
khuddlefish: It might be good to move this into the else block too

backup/incremental_test.go
khuddlefish: Or actually it looks like they aren't, but maybe they should be in NewHistory instead of here. Correct me if I'm wrong, but it looks like you're testing the specific errors that NewHistory can encounter.

backup_history/history.go
```diff
@@ -91,7 +93,13 @@ func WriteBackupHistory(historyFilePath string, currentBackupConfig *BackupConfi
...
+	var err error
```
khuddlefish: The error could be declared inside the conditional since you don't really need it outside

backup/incremental_test.go
nadeemg: True, but we felt like the tests on GetLatestMatchingBackupTimestamp() were specifying existing behavior and gave confidence that the refactor was correct. 

After the refactor, NewHistory doesn't seem worth testing. We basically call a couple of golang-provided functions and return any errors. Since the only logic worth testing (if backup_history.yaml was present or not and to panic) was moved out of NewHistory and into the calling function it seemed appropriate to move the tests.

backup/incremental_test.go
nadeemg: done

backup_history/history.go
nadeemg: Good point

backup_history/history.go
nadeemg: done

backup_history/history.go
nadeemg: Not really. There's an edge case where NewHistory might return an empty History object. We've updated the log message to be clearer.

backup/incremental.go
nadeemg: done

backup_history/history.go
khuddlefish: But if the history file exists, shouldn't it always have at least one entry in it?

backup/incremental.go
```diff
@@ -37,7 +38,13 @@ func GetTargetBackupTimestamp() string {
...
+	var err error
+
+	if iohelper.FileExistsAndIsReadable(globalFPInfo.GetBackupHistoryFilePath()) {
+		history, err = backup_history.NewHistory(globalFPInfo.GetBackupHistoryFilePath())
+		gplog.FatalOnError(err)
+	}
 	latestMatchingBackupHistoryEntry := GetLatestMatchingBackupConfig(history, &backupReport.BackupConfig)
```
khuddlefish: I think you should only need to run this function if the history file exists. Otherwise, you know there won't be entries.

backup/incremental_test.go
khuddlefish: It still seems to me like "fatals when gpbackup_history.yaml can't be read" and "fatals when gpbackup_history.yaml is an invalid format" should be tested for NewHistory, not here. This function should just care about whether the history file exists or not because that is logic where it branches. It should be able to trust that NewHistory can respond to errors appropriately.

backup/incremental.go
nadeemg: good point

PR #187, opened 2018-11-10
Separate entries

khuddlefish:
This is a POC I tossed together to try out a method for splitting statements into separate TOC entries. It's not perfect, but hopefully you can get the idea. The main thought I had was having functions generate a slice of statements and then pass those statements to a function that simply prints the statements and creates a TOC entry for each one.


nadeemg:
Looks good!

PR #188, opened 2018-11-13
Convert table Relations and TableDefinitions to Table struct

khuddlefish:
Tables needed a struct that was not shared by other objects so we can support struct methods that are unique to tables. We already have the Sequence struct which combines a Relation and SequenceDefinition, so using a similar method for tables makes sense.

Other changes:
* converts QuerySimpleDefinition to RuleDefinition and TriggerDefinition so rules and triggers can have unique structs.
* Split predata_relations_test.go into two files since the file was too big
* Other minor refactoring in predata and queries relations.


backup/queries_postdata.go
```diff
@@ -181,20 +186,44 @@ AND rulename NOT LIKE 'pg_%%'
...
 
-func GetTriggers(connectionPool *dbconn.DBConn) []QuerySimpleDefinition {
+type TriggerDefinition struct {
+	Oid          uint32
+	Name         string
+	OwningSchema string
+	OwningTable  string
+	Def          string
+}
+
```
nadeemg: Could we instead say: `type TriggerDefinition RuleDefinition`

khuddlefish: It looks like we can. Neat :)

PR #189, opened 2018-11-17
Create separate TOC entries for object creation and metadata

khuddlefish:
We discovered an issue where the way we are executing SQL queries on
restore does not recognize the block of statements we are trying to run
as separate statements. We can have a case where a user is running with
--on-error-continue and some statements will pass, but others may fail
and it will rollback the entire block. Our solution is to create
separate TOC entries for each statement so they will all be executed on
their own at restore time.

This modifies PrintObjectMetadata and AddMetadataEntry functions to take
in a struct that satisfies the TOCObject interface. We call the
GetMetadataEntry method on these structs to add new entries rather than
passing in arguments so it is easier to add entries from within
the PrintObjectMetadata function.

### DIFF
--- a/backup/metadata_globals.go
+++ b/backup/metadata_globals.go
@@ -20,13 +20,12 @@ func PrintSessionGUCs(metadataFile *utils.FileWithByteCount, toc *utils.TOC, guc
 	metadataFile.MustPrintf(`
 SET client_encoding = '%s';
 `, gucs.ClientEncoding)
-	toc.AddGlobalEntry("", "", "SESSION GUCS", start, metadataFile)
+	toc.AddMetadataEntry(gucs, start, metadataFile.ByteCount)
 }
 
 func PrintCreateDatabaseStatement(metadataFile *utils.FileWithByteCount, toc *utils.TOC, defaultDB Database, db Database, dbMetadata MetadataMap) {
-	dbname := db.Name
 	start := metadataFile.ByteCount
-	metadataFile.MustPrintf("\n\nCREATE DATABASE %s TEMPLATE template0", dbname)
+	metadataFile.MustPrintf("\n\nCREATE DATABASE %s TEMPLATE template0", db.Name)
 	if db.Tablespace != "pg_default" {
 		metadataFile.MustPrintf(" TABLESPACE %s", db.Tablespace)
 	}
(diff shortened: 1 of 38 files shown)


khuddlefish:
Edge cases to notice:

* I used AddMetadataEntryLongArgs with the previous argument style in a few places where the struct method approach didn't work well.
* Databases are kind of weird. I put DATABASE METADATA as object type for GetMetadataEntry for the PrintObjectMetadata call, but use the LongArgs for the actual DATABASE entry. I thought about just using DATABASE for everything, but we do want the ability to restore the database metadata without the database if someone does --with-globals but doesn't use --create-db.
* I changed the FQN function for functions and aggregates to use IdentArgs in the name because that is what PrintObjectMetadata needs. The change didn't appear to break anything, but something to watch out for.
* See what we're doing with the SEQUENCE OWNER entry.
* I modified the procedural language toc object type from "PROCEDURAL LANGUAGE" to "LANGUAGE" so it can be used to add an entry and Print Metadata. This doesn't seem to have broken things either.
* Also changed TABLESPACE METADATA -> TABLESPACE since we weren't using the distinction like we do for database.

backup/metadata_globals.go
```diff
@@ -276,28 +263,30 @@ func PrintCreateRoleStatements(metadataFile *utils.FileWithByteCount, toc *utils
...
 	for roleName := range roleGUCs {
-		start := metadataFile.ByteCount
 		for _, roleGUC := range roleGUCs[roleName] {
+			start := metadataFile.ByteCount
 			dbString := ""
 			if roleGUC.DbName != "" {
 				dbString = fmt.Sprintf("IN DATABASE %s ", roleGUC.DbName)
 			}
 			metadataFile.MustPrintf("\n\nALTER ROLE %s %s%s;", roleName, dbString, roleGUC.Config)
+			toc.AddMetadataEntryLongArgs("", roleName, "ROLE GUCS", "", start, metadataFile, "global")
```
kyeap-pivotal: Maybe add to RoleGUC struct?

utils/toc.go
```diff
@@ -261,24 +261,22 @@ func (toc *TOC) InitializeMetadataEntryMap() {
...
-func (toc *TOC) AddPostdataEntry(schema string, name string, objectType string, referenceObject string, start uint64, file *FileWithByteCount) {
-	toc.AddMetadataEntry(schema, name, objectType, referenceObject, start, file, "postdata")
+func (toc *TOC) AddMetadataEntry(obj TOCObject, start, end uint64) {
+	section, entry := obj.GetMetadataEntry(start, end)
+	*toc.metadataEntryMap[section] = append(*toc.metadataEntryMap[section], entry)
 }
 
-func (toc *TOC) AddStatisticsEntry(schema string, name string, objectType string, start uint64, file *FileWithByteCount) {
-	toc.AddMetadataEntry(schema, name, objectType, "", start, file, "statistics")
+func (toc *TOC) AddMetadataEntryLongArgs(schema string, name string, objectType string, referenceObject string, start uint64, file *FileWithByteCount, section string) {
```
kyeap-pivotal: I know we discussed this in person, but just noting here that it might easier to read if the MetadataEntry was constructed outside the function and passed in rather than passing in all the arguments required to make it.

backup/metadata_globals.go
khuddlefish: done

utils/toc.go
khuddlefish: We will make a chore to do that later

PR #193, opened 2018-12-03
Modify example plugin to place files in a timestamp dir

kyeap-pivotal:
Modify example plugin to place files in a timestamp dir

### DIFF
--- a/plugins/example_plugin.sh
+++ b/plugins/example_plugin.sh
@@ -10,7 +10,8 @@ setup_plugin_for_backup(){
   elif [ "$3" = "segment" ]
     then echo "setup_plugin_for_backup was called for scope = segment" >> /tmp/plugin_out.txt
   fi
-  mkdir -p /tmp/plugin_dest
+  timestamp_dir=`basename "$2"`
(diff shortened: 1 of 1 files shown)


plugins/example_plugin.sh
```diff
@@ -49,25 +50,35 @@ cleanup_plugin_for_restore(){
...
   echo "restore_data $1 $2" >> /tmp/plugin_out.txt
   filename=`basename "$2"`
-	cat /tmp/plugin_dest/$filename
+  timestamp_dir=`basename $(dirname "$2")`
+	cat /tmp/plugin_dest/$timestamp_dir/$filename
+}
+
+delete_dir() {
+  echo "delete_dir $1 $2" >> /tmp/plugin_out.txt
+  rm -fr /tmp/plugin_dest/$2
```
khuddlefish: `rm -fr` sounds so weird when I'm used to `rm -rf`, but it's not technically wrong :P

kyeap-pivotal: done.

PR #195, opened 2018-12-06
Fix plugin tests

kyeap-pivotal:
The last commit pushed a change to the example plugin which was being
exercised in a few tests. This change corrects test setup and asserts
affected by the example plugin change.

### DIFF
--- a/integration/helper_test.go
+++ b/integration/helper_test.go
@@ -19,8 +19,8 @@ import (
 )
 
 var (
-	testDir          = "/tmp/helper_test"
-	pluginDir        = "/tmp/plugin_dest"
+	testDir          = "/tmp/helper_test/20180101010101"
+	pluginDir        = "/tmp/plugin_dest/20180101010101"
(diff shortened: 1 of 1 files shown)


integration/helper_test.go
```diff
@@ -19,8 +19,8 @@ import (
 )
 
 var (
-	testDir          = "/tmp/helper_test"
-	pluginDir        = "/tmp/plugin_dest"
+	testDir          = "/tmp/helper_test/20180101010101"
```
tyacovone: Does the helper_test dir need this too?

kyeap-pivotal: Yep it does! The directory `/tmp/helper_test/[timestamp]` is the staging area for the data file the plugin will be consuming. And because some of the plugin commands extract the timestamp from the filepath of the data file, if the timestamp is not on the end of the path, the plugin will land the resulting files in the wrong directory.

PR #196, opened 2018-12-06
Separate types into different structs and simplify types queries

khuddlefish:


khuddlefish:
Pipeline failure for 4.3 and 5: https://dev.ci.gpdb.pivotal.io/teams/main/pipelines/gpbackup-separate-types/jobs/integrations-GPDB5/builds/1. This is why we run dev pipeline before pushing changes :) Will fix these issues before pushing

PR #200, opened 2018-12-11
 Add delete_dir test to plugin_test_bench

tyacovone:
We also updated the plugin api version to 0.4.0


plugins/README.md
```diff
@@ -286,7 +286,7 @@ test_plugin restore_data /home/test_plugin_config.yaml /data_dir/backups/2018010
 ```
 ### [plugin_api_version](#plugin_api_version)
 
-This command should echo the gpbackup plugin api version to stdout. The version for this gpbackup plugin api is 0.1.0.
+This command should echo the gpbackup plugin api version to stdout. The version for this gpbackup plugin api is 0.4.0.
```
larham: Is there a way to avoid a version number here?  That version may change, and it will be easy to miss.  It looks like it was missed for v.0.2, 0.3

plugins/plugin_test_bench.sh
```diff
@@ -140,6 +140,29 @@ fi
...
+echo "[RUNNING] delete_dir"
+echo $data | $plugin backup_data $plugin_config $testdata
+
+# ensures the dir exists prior to deletion
+$plugin restore_data $plugin_config $testdata > /dev/null
+
+$plugin delete_dir $plugin_config 20180101010101
+
+set +e
+# output=`$plugin restore_data $plugin_config $testdata 2>&1`
```
larham: can this commented-out line be removed?

plugins/plugin_test_bench.sh
```diff
@@ -140,6 +140,29 @@ fi
...
+output=$($plugin restore_data $plugin_config $testdata 2>/dev/null)
+retval=$(echo $?)
+if [ "${output}" = "${data}"  ] || [ "$retval" -eq "0" ] ; then
```
larham: please beware of "=" versus "=="

tyacovone: sh uses '=' for str compare. I would say, though, that having consistency in our condition would be nice -- ill use '=' for both conditions, since were stringifying the retval

plugins/plugin_test_bench.sh
tyacovone: yep, good catch

plugins/README.md
tyacovone: Updated the README. We no longer reference version in these lines, and we've appended a `Release Notes` section to the bottom for tracking version update details

khuddlefish:
@larham I think we addressed all of your suggestions. Would you like to take another look?

PR #207, opened 2019-01-08
add comments and more clear tests for quoted characters

larham:
* make it clear that any character is accepted for an identifier between quotes


kyeap-pivotal:
Do we still want to merge this in considering that we are working on resolving some of the special character inputs which might change of of this code?

larham:
yeah, I'll close this.

PR #210, opened 2019-01-15
add configuration info to README

larham:


khuddlefish:
I agree with @kyeap-pivotal Feel free to suggest recommended configuration flags for gpbackup here, but the other errors about which packages to install and which flags to use for troubleshooting do not belong here. The releng team will be adding a little more info about zstd to the main README as well.

PR #212, opened 2019-01-16
Makefile lists subdirs; fix linting errors

larham:


Makefile
```diff
@@ -12,6 +12,9 @@ GIT_VERSION := $(shell git describe --tags | perl -pe 's/(.*)-([0-9]*)-(g[0-9a-f
 BACKUP_VERSION_STR="-X github.com/greenplum-db/gpbackup/backup.version=$(GIT_VERSION)"
 RESTORE_VERSION_STR="-X github.com/greenplum-db/gpbackup/restore.version=$(GIT_VERSION)"
 HELPER_VERSION_STR="-X github.com/greenplum-db/gpbackup/helper.version=$(GIT_VERSION)"
+SUBDIRS_PRODUCTION=backup/ backup_filepath/ backup_history/ helper/ options/ restore/ utils/
```
kyeap-pivotal: Can we just put all these directories into $SUBDIRS_HAS_UNIT?

larham: consolidated variables as you suggested.  Added comment to identify "/testutils" as an unusual directory that has unit tests despite being test-only code.

PR #218, opened 2019-02-15
remove Collate information when backing up External tables

larham:
External tables cannot restore any COLLATE information about attributes. Therefore, if that information is present when backing up, ignore it within the metadata for external tables.

### DIFF
--- a/backup/predata_acl.go
+++ b/backup/predata_acl.go
@@ -50,6 +50,36 @@ type ACL struct {
 	ConnectWithGrant    bool
 }
 
+func (acl ACL) Clone() ACL {
+	return ACL{
+		Grantee:             acl.Grantee,
+		Select:              acl.Select,
+		SelectWithGrant:     acl.SelectWithGrant,
+		Insert:              acl.Insert,
+		InsertWithGrant:     acl.InsertWithGrant,
+		Update:              acl.Update,
+		UpdateWithGrant:     acl.UpdateWithGrant,
+		Delete:              acl.Delete,
+		DeleteWithGrant:     acl.DeleteWithGrant,
+		Truncate:            acl.Truncate,
+		TruncateWithGrant:   acl.TruncateWithGrant,
+		References:          acl.References,
+		ReferencesWithGrant: acl.ReferencesWithGrant,
+		Trigger:             acl.Trigger,
+		TriggerWithGrant:    acl.TriggerWithGrant,
+		Usage:               acl.Usage,
+		UsageWithGrant:      acl.UsageWithGrant,
+		Execute:             acl.Execute,
+		ExecuteWithGrant:    acl.ExecuteWithGrant,
+		Create:              acl.Create,
+		CreateWithGrant:     acl.CreateWithGrant,
+		Temporary:           acl.Temporary,
+		TemporaryWithGrant:  acl.TemporaryWithGrant,
+		Connect:             acl.Connect,
+		ConnectWithGrant:    acl.ConnectWithGrant,
+	}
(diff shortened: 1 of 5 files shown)


chrishajas:
Would it be possible to modify the query to not get that information in the first place if it's an external table (pg_class.relstorage should have this info)? ie in `	CASE WHEN a.attcollation <> t.typcollation THEN quote_ident(cn.nspname) || '.' || quote_ident(coll.collname) ELSE '' END AS collation,`

kyeap-pivotal:
I dug into this more because I couldn't reproduce this locally, which led me to the line Chris mentioned. 

@larham I believe this is an issue cause by the [collate hotfix commit](https://github.com/greenplum-db/gpbackup/commit/06a5051e931b9698c59bb3059c5413fb9a33209d) we made on the regression branch. It seems I incorrectly reasoned we could do away with the conditional. We made the hotfix in order to get past the gpbackup error`[CRITICAL]:-sql: Scan error on column index 12: unsupported Scan, storing driver.Value type <nil> into type *string` as seen on [this backup regression run](https://dev.ci.gpdb.pivotal.io/teams/main/pipelines/dev:gpbackup_regression/jobs/gpbackup-gpdb6-ICW-regression-DB/builds/21). It was theorized that the issue was caused by trying to quote_ident(NULL) which results in a NULL value. Meaning the DB driver was attempting to store the NULL value into what should have been a string. We made the hotfix, and gpbackup moved on.

Reverting this hotfix seems to correct the collate issue. The safer hotfix would have been `CASE WHEN a.attcollation <> t.typcollation THEN coalesce(quote_ident(cn.nspname) || '.' || quote_ident(coll.collname), '') ELSE '' END AS collation,`. 

@chrishajas According to the comment in pg_dump.c, it seems that the conditional  `CASE WHEN a.attcollation <> t.typcollation` is so pg_dump does not `"...dump COLLATE clauses for attributes whose collation is different from their type's default..."` which isn't immediately clear to me as how this relates to external tables.

Going forward I think it might be better make the safe hotfix change. Our theory for the cause of the gpbackup error seem to be good; we managed to bypass the error with our original hotfix. However, it is not apparent yet to me why coll.collname or cn.nspname can be NULL.

chrishajas:
@kyeap-pivotal Yes, we have that case statement to achieve the same goal as pg_dump's:
```
			 * attcollation is new in 9.1.  Since we only want to dump COLLATE
			 * clauses for attributes whose collation is different from their
			 * type's default, we use a CASE here to suppress uninteresting
			 * attcollations cheaply.
```

However, I think in gpdb for external tables, the attcollation is the same as the typcollation in all cases since that cannot be configured by a user, so your proposed fix looks good! It would be good to also have a test to ensure we don't accidentally do this again. The way we've done it in the past is by creating an object in a query integration test and ensure the struct contains the expected values.

kyeap-pivotal:
Hotfix has not yet been merged into the master branch. I agree we'll need tests to vet the change for it before it gets merged.

tyacovone:
Fix looks good to me – good work Kevin & thanks Chris for weighing in on this. I agree a test or two should be added prior to merging

larham:
Closing this in favor of a new PR https://github.com/greenplum-db/gpbackup/pull/220   which attempts to solve the issue with the query itself, rather than patching the results.

@chrishajas @khuddlefish we'd love to get your eyes on the new PR. Thanks for your help with this. BTW, Karen, both Kevin and I saw a doc that implied default for "JOIN" was "left join", which is not the case, as you mentioned.

PR #219, opened 2019-02-16
Support check constraints on parent and child tables

kyeap-pivotal:
Previously, it was possible to add check constraints to tables which
have inherited tables underneath it without the constraint propagating
to its children.  This is changed in GPDB6 since check constraints on
parent tables MUST also be applied to it's children.

It's unsure if we can just change all constraint statements to not use the ONLY keyword constraint so I took a safe approach.

### DIFF
--- a/backup/predata_shared.go
+++ b/backup/predata_shared.go
@@ -37,8 +37,9 @@ func PrintConstraintStatements(metadataFile *utils.FileWithByteCount, toc *utils
 		if constraint.IsDomainConstraint {
 			continue
 		}
+		// ConIsLocal should always return true from GetConstraints because we filter out constraints that are inherited using the INHERITS clause, or inherited from a parent partition table. This field only accurately reflects constraints in GPDB6+ because check constraints on parent tables must propogate to children. For GPDB versions 5 or lower, this field will default to false.
 		objStr := "TABLE ONLY"
-		if constraint.IsPartitionParent {
+		if constraint.IsPartitionParent || (constraint.ConType == "c" && constraint.ConIsLocal) {
 			objStr = "TABLE"
 		}
 		metadataFile.MustPrintf(alterStr, objStr, constraint.OwningObject, constraint.Name, constraint.ConDef)
--- a/backup/queries_shared.go
+++ b/backup/queries_shared.go
@@ -79,6 +79,7 @@ type Constraint struct {
 	Name               string
 	ConType            string
 	ConDef             string
+	ConIsLocal         bool
 	OwningObject       string
 	IsDomainConstraint bool
 	IsPartitionParent  bool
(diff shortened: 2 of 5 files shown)


backup/queries_shared.go
```diff
@@ -132,13 +139,14 @@ AND %s
 AND c.relname IS NOT NULL
 AND conrelid NOT IN (SELECT parchildrelid FROM pg_partition_rule)
 AND (conrelid, conname) NOT IN (SELECT i.inhrelid, con.conname FROM pg_inherits i JOIN pg_constraint con ON i.inhrelid = con.conrelid JOIN pg_constraint p ON i.inhparent = p.conrelid WHERE con.conname = p.conname)
-GROUP BY con.oid, conname, contype, c.relname, n.nspname, pt.parrelid`, "%s", ExtensionFilterClause("c"))
+GROUP BY con.oid, conname, contype, c.relname, n.nspname, %s pt.parrelid`, selectConIsLocal, "%s", ExtensionFilterClause("c"), groupByConIsLocal)
```
larham: Please consider a slightly different formatting here, just for readability.  The backtick seems subtle, easy to miss.

What about separating the line at the backtick?

Would a replacement for the entire "groupby" clause be more readable?  It would be something like

groupby := fmt.Sprintf("con.oid, conname, contype, c.relname, n.nspname, %s pt.parrelid", selectConIsLocal)

integration/predata_shared_create_test.go
```diff
@@ -69,6 +69,14 @@ var _ = Describe("backup integration create statement tests", func() {
 			partitionCheckConstraint = backup.Constraint{Oid: 0, Schema: "public", Name: "check1", ConType: "c", ConDef: "CHECK (id <> 0)", OwningObject: "public.part", IsDomainConstraint: false, IsPartitionParent: true}
 			testhelper.AssertQueryRuns(connectionPool, "CREATE TABLE public.testtable(a int, b text) DISTRIBUTED BY (b)")
 			conMetadataMap = backup.MetadataMap{}
+
+			if connectionPool.Version.AtLeast("6") {
+				uniqueConstraint.ConIsLocal = true
+				pkConstraint.ConIsLocal = true
+				fkConstraint.ConIsLocal = true
+				checkConstraint.ConIsLocal = true
+				partitionCheckConstraint.ConIsLocal = true
```
larham: Please consider adding a test where one of these `ConIsLocal` is turned false.  I don't have enough context to understand what that might mean for GetConstraint(), so if the test title explained how that affected some table and thus the constraint, it would help me understand the function better.

backup/predata_shared.go
```diff
@@ -38,7 +38,7 @@ func PrintConstraintStatements(metadataFile *utils.FileWithByteCount, toc *utils
 			continue
 		}
 		objStr := "TABLE ONLY"
-		if constraint.IsPartitionParent {
```
khuddlefish: I'm pretty sure we need a related clause to tell it not to print check constraints at all if they are on child partition tables. For example, if you have
```
parent
|         |
child1    child2
```
and they all have a check constraint, we used to print 3 `ALTER TABLE ONLY tablename` statements for the 3 tables. We now want to print `ALTER TABLE parent` and not print anything for the child tables. I'm not sure if that case is covered here yet, but it would be good to have an explicit test for it.

integration/predata_shared_create_test.go
kyeap-pivotal: I could not think of a scenario where ConIsLocal is false for GetConstraints() because of the filters we have on this query.

@khuddlefish 
Correct me if I'm wrong, but It looked to me that the query in GetConstraints is smart enough to exclude constraints gotten from a parent partition and inherited constraints because of these WHERE filters in getConstraints().
```
...
AND conrelid NOT IN (SELECT parchildrelid FROM pg_partition_rule)
AND (conrelid, conname) NOT IN (SELECT i.inhrelid, con.conname FROM pg_inherits i JOIN pg_constraint con ON i.inhrelid = con.conrelid JOIN pg_constraint p ON i.inhparent = p.conrelid WHERE con.conname = p.conname)
...

File: backup/predata_shared.go (hunk 3 above)
Comment (kyeap-pivotal) [reply backup/predata_shared.go]: We do have a couple of tests for these scenarios
```
It("does not return a constraint array for a table that inherits a constraint from another table", func() {...})
It("returns a constraint array for a parent partition table with one CHECK constraint", func() {...})
```
Currently, constraints on a table gotten by means of inheritance or by being passed from parent to child partition do not have their own ALTER statements.

integration/predata_shared_create_test.go
khuddlefish: If ConIsLocal is always false, why are we selecting it as a property?

backup/predata_shared.go
khuddlefish: Ok great. They shouldn't have their own ALTER statements so as long as you're confident we aren't printing them, it's probably good.

PR #220, opened 2019-02-20
Filter out attributes from indexes in GetColumnDefinitions

kyeap-pivotal:
* change inner join on pg_namespace to be a LEFT OUTER JOIN
  -- this avoids joining that table twice
* add where condition that we do not want to have attributes from
indices (where c.reltype <> 0)
* spell out "LEFT OUTER JOIN" to help readability

Co-authored-by: Larry Hamel <lhamel@pivotal.io>

### DIFF
--- a/backup/queries_table_defs.go
+++ b/backup/queries_table_defs.go
@@ -242,14 +242,15 @@ SELECT
 FROM pg_catalog.pg_attribute a
 JOIN pg_class c ON a.attrelid = c.oid
 JOIN pg_namespace n ON c.relnamespace = n.oid
-LEFT JOIN pg_catalog.pg_attrdef ad ON (a.attrelid = ad.adrelid AND a.attnum = ad.adnum)
-LEFT JOIN pg_catalog.pg_type t ON a.atttypid = t.oid
-LEFT JOIN pg_collation coll on (a.attcollation = coll.oid)
-LEFT JOIN pg_namespace cn on (coll.collnamespace = cn.oid)
-LEFT JOIN pg_catalog.pg_attribute_encoding e ON e.attrelid = a.attrelid AND e.attnum = a.attnum
-LEFT JOIN pg_description d ON (d.objoid = a.attrelid AND d.classoid = 'pg_class'::regclass AND d.objsubid = a.attnum)
-LEFT JOIN pg_seclabel sec ON (sec.objoid = a.attrelid AND sec.classoid = 'pg_class'::regclass AND sec.objsubid = a.attnum)
+LEFT OUTER JOIN pg_catalog.pg_attrdef ad ON (a.attrelid = ad.adrelid AND a.attnum = ad.adnum)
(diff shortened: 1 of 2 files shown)


larham:
@chrishajas @khuddlefish we'd love to get your eyes on the new PR. Thanks for your help with this. BTW, Karen, both Kevin and I saw a doc that implied default for "JOIN" was "left join", which is not the case, as you mentioned.

backup/queries_table_defs.go
```diff
@@ -235,21 +235,21 @@ SELECT
 	coalesce(pg_catalog.pg_get_expr(ad.adbin, ad.adrelid), '') AS defaultval,
 	coalesce(pg_catalog.array_to_string(a.attoptions, ','), '') AS options,
 	coalesce(array_to_string(ARRAY(SELECT option_name || ' ' || quote_literal(option_value) FROM pg_options_to_table(attfdwoptions) ORDER BY option_name), ', '), '') AS fdwoptions,
-	CASE WHEN a.attcollation <> t.typcollation THEN quote_ident(cn.nspname) || '.' || quote_ident(coll.collname) ELSE '' END AS collation,
+	CASE WHEN a.attcollation <> t.typcollation THEN quote_ident(n.nspname) || '.' || quote_ident(coll.collname) ELSE '' END AS collation,
```
khuddlefish: I'm not sure this schema would always match up correctly to the actual schema for the collation. n comes from joining pg_class c on pg_namespace where the relnamespace property from pg_class matches up with the oid of pg_namespace. The column that determines which schema goes with a collation is collnamespace from pg_collation so the first match will not do anything to pair the collation schemas. I think this will work ok as long as collations happen to be in the same schema as the table you are looking at, but I'm fairly certain you can't guarantee that.

I would recommend creating a collation in a different schema, and creating a table with some columns that use that collation and see if this query picks up all the schemas correctly.

khuddlefish:
This stack overflow answer indicates that the implied default for "JOIN" is "INNER JOIN" which is consistent with what we often saw as we were writing queries. We would often try "JOIN" first to see if it would work and we would not get all the rows back we expected. We would then change it to "LEFT JOIN" and would then get the correct results. I would be inerested to see the document you found that says something different. https://softwareengineering.stackexchange.com/questions/206035/join-vs-inner-join-and-full-outer-join

Also according to this the "OUTER" keyword is optional since all "LEFT JOIN"s are "OUTER". https://stackoverflow.com/questions/6625964/left-join-vs-left-outer-join

larham:
@chrishajas excellent test, we are adding it, and it did break the refactor.
@khuddlefish agreed about optionality of "OUTER", which I consider helpful for novices like me. And what I was trying to say is simply that you were right and I was wrong; I must have misread something, but I did it in good company :)

khuddlefish:
@larham No worries. Let me or @chrishajas know if you would like us to work with you to help understand why the refactor broke on that test case. The same table joins are definitely tricky.

kyeap-pivotal:
Updated with additional tests and a partial revert.

PR #221, opened 2019-02-22
Re-enable plugin backwards compatibility tests

kyeap-pivotal:
Previously, plugin backwards compatibilty was tested against the version
of gpbackup the plugins were released with. However due a change to
example plugin's backup directory structure and the fact that
gpbackup1.7.0 is only compatible with pluginAPI version 0.3.0, the only
way these tests could be re-enabled was if the oldest versions these
tested against was bumped up to when both of these issues were resolved.


kyeap-pivotal:
Closing this PR due to explanation provided by previous comment.

PR #223, opened 2019-02-27
Change WriteOidListToSegments() to use `scp`

larham:
Previously, OIDs were transmitted to segments without `scp`, via "echo". However, a large enough database will have a million OIDs, which can blow out the capacity of a single command line.

### DIFF
--- a/utils/agent_remote.go
+++ b/utils/agent_remote.go
@@ -2,8 +2,12 @@ package utils
 
 import (
 	"fmt"
+	"io"
+	"io/ioutil"
 	"path/filepath"
 	"strings"
+
+	"github.com/greenplum-db/gp-common-go-libs/iohelper"
 
 	"github.com/greenplum-db/gp-common-go-libs/cluster"
 	"github.com/greenplum-db/gp-common-go-libs/gplog"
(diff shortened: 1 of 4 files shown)


utils/agent_remote.go
```diff
@@ -27,15 +31,43 @@ func CreateFirstSegmentPipeOnAllHosts(oid string, c *cluster.Cluster, fpInfo bac
...
+	defer func() {
+		err = os.Remove(localOidFile.Name())
+		if err != nil {
+			gplog.Warn("cannot remove temporary oid file: %s", localOidFile.Name())
+		}
+	}()
+
+	writeOidsToFile(localOidFile.Name(), oidList)
+
+	for _, contentID := range c.ContentIDs {
```
chrishajas: This may be slower if you have many segments. I'm also not sure if rsync would be better than scp here. 

We do something similar when copying the plugin config to all the segment hosts here: https://github.com/greenplum-db/gpbackup/blob/master/utils/plugin.go#L209

larham: Awesome feedback as usual.  The need for multithreading is for clear, thanks for the heads-up.  Looking into the `CopyPluginConfigToAllHosts()` method, it uses `GenerateAndExecuteCommand()`, which means it does `ssh` to the remote, and then from there calls `rsync` back to copy from the master to what is now "local" in the `ssh` shell. The `ssh` part seems unnecessary. I'm guessing the tool for `ssh` was available, and got reused in a one-off situation. This might be an opportunity to build a similar tool, one that does local `scp` or `rsync` or whatever commands from master to segments, since there are at least 2 usages now.

chrishajas: Yes, you're right; it'd be best to do that.

PR #228, opened 2019-03-11
Create build-dependencies job for gpbackup pipeline

tyacovone:
* this addition is intended to lessen the impact of failures related
to a known 'make dep' flake. This new pipeline pattern fetches all dependencies in the
build-dependencies job and then passes the 'vendor' directory to other jobs
via an s3 bucket.
* pipeline reformatted – jobs only trigger after build_deps finishes (with the
exception of plugin jobs, which also trigger on new s3 or ddboost
commits)

Signed-off-by: Kevin Yeap <kyeap@pivotal.io>

### DIFF
--- a/ci/pipeline.yml
+++ b/ci/pipeline.yml
@@ -77,6 +77,15 @@ resources:
       access_key_id: {{bucket-access-key-id}}
       secret_access_key: {{bucket-secret-access-key}}
 
+- name: gpbackup-dependencies
+  type: s3
+  source:
+      bucket: gpbackup-dependencies
+      versioned_file: gpbackup-dependencies/dependencies.tar.gz
+      region_name: us-west-2
+      access_key_id: {{bucket-access-key-id}}
+      secret_access_key: {{bucket-secret-access-key}}
+
 - name: bin_gpdb_5x_stable
   type: s3
   source:
@@ -210,6 +219,21 @@ resources:
       bucket_path: clusters-aws/
 
 jobs:
+- name: build_dependencies
+  plan:
+  - aggregate:
+    - get: gpbackup
+      trigger: true
+    - get: gpbackup-dependencies
+    - get: nightly-trigger
+      trigger: true
+  - task: build-dependencies
+    file: gpbackup/ci/tasks/build-dependencies.yml
+    attempts: 5
+  - put: gpbackup-dependencies
+    params:
+        file: output_deps/dependencies.tar.gz
+
 - name: units
   plan:
   - aggregate:
@@ -217,6 +241,9 @@ jobs:
       trigger: true
     - get: nightly-trigger
       trigger: true
+    - get: gpbackup-dependencies
+      passed:
+       - build_dependencies
   - task: unit-tests
     file: gpbackup/ci/tasks/unit-tests.yml
     on_failure:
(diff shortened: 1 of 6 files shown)


ci/pipeline.yml
```diff
@@ -77,6 +77,15 @@ resources:
       access_key_id: {{bucket-access-key-id}}
       secret_access_key: {{bucket-secret-access-key}}
 
+- name: gpbackup-dependencies
+  type: s3
+  source:
+      bucket: gpbackup-dependencies
+      versioned_file: gpbackup-dependencies/dependencies.tar.gz
```
larham: Do we need to keep previous copies--do we need versioning?

larham:
Given a case where we *remove* a dependency from `dep`, how can we also remove it from the latest .tar ?  I see a "prune" option in the `dep` config, but when I remove a constraint, I don't see those files removed, in a quick test.  This request is NOT a show-stopper: just inquiring in case it is easy.

ci/tasks/build-dependencies.yml
```diff
@@ -0,0 +1,34 @@
...
+    export GOPATH=$PWD/go
+    export PATH=$GOPATH/bin:$PATH
+
+    tar -zxf gpbackup-dependencies/dependencies.tar.gz -C $GOPATH/src/github.com/greenplum-db/gpbackup/
+
+    pushd $GOPATH/src/github.com/greenplum-db/gpbackup
+      make depend
+      tar cfz dependencies.tar.gz vendor
+    popd
+    cp $GOPATH/src/github.com/greenplum-db/gpbackup/dependencies.tar.gz output_deps/
```
larham: instead of `cp`, consider using `rsync` because if the files are the same (no changes to dependencies), then `rsync` will finish quickly without doing anything

ci/pipeline.yml
```diff
@@ -210,13 +219,28 @@ resources:
...
+    params:
+        file: output_deps/dependencies.tar.gz
+
+- name: units
+  plan:
+  - aggregate:
+    - get: gpbackup
+    - get: gpbackup-dependencies
+      passed:
+       - build_dependencies
```
larham: consider that all these jobs need not depend on whether the `build_dependencies` job succeeded--they will redo download work if they need to.

ci/pipeline.yml
tyacovone: Versioning _of some sort_ appears to be required of a concourse resource. We have the option of using a versioned file, as it currently is, or having a specified semver/hardcoded version in the tarball's name (e.g. `dependencies-1.0.0`) which is parsed by regex.

tyacovone: Versioning _of some sort_ is required of a concourse resource. We have the option of using a versioned file, as it currently is, or having a specified semver/hardcoded version in the tarball's name (e.g. dependencies-1.0.0) which is parsed by regex. In the latter case, we would be using a 'dummy' version, since we would always intend to overwrite the the dependency tarball with the latest version.

tyacovone: Given that each versioned file upload is < 3MB and I'm encountering some strange issues when trying to use the dummy file name `dependencies-0.tar.gz`, im leaning towards leaving implementation as is.

tyacovone:
> Given a case where we _remove_ a dependency from `dep`, how can we also remove it from the latest .tar ? I see a "prune" option in the `dep` config, but when I remove a constraint, I don't see those files removed, in a quick test. This request is NOT a show-stopper: just inquiring in case it is easy.

This seems to be handled properly with dep ensure. The following steps remove a directory that previously existed:

ci/pipeline.yml
larham: Excellent bucket policy, that makes me happy.

ci/tasks/build-dependencies.yml
larham: I like the substitution.  When using `rsync`, I'm used to seeing `-a` as the common example, such as in docs like https://linux.die.net/man/1/rsync

This will preserve perms, timestamp, etc.  And it includes `-r` as part of `-a`.

kyeap-pivotal:
Can we also squash all these commits into 1 before pushing?

PR #231, opened 2019-03-19
Regression squashed

kyeap-pivotal:
I made a new branch so the differences can be seen more easily. I squashed most of the commits together and If the last commit looks good It will be squashed into the first commit.

These commits contains our pipeline that tests for diffs in gpbackup vs pg_dump.

### DIFF
--- a/ci/tasks/gprestore-regression.yml
+++ b/ci/tasks/gprestore-regression.yml
@@ -1,0 +1,79 @@
+PLATFORM: linux
+
+image_resource:
+  type: docker-image
+  source:
+    repository: pivotaldata/centos-gpdb-dev
+    tag: '6-gcc6.2-llvm3.7'
+
+inputs:
+- name: gpbackup
+  path: go/src/github.com/greenplum-db/gpbackup
+- name: ccp_src
+- name: cluster_env_files
+- name: gpdb_src
+- name: backup_artifact_resource
+- name: regression_dump
+
+run:
+  path: bash
+  args:
+  - -c
+  - |
+    set -ex
+
+    ccp_src/scripts/setup_ssh_to_cluster.sh
+    cp -r backup_artifact_resource/* /tmp
+    cp -r regression_dump/* /tmp
+
+    pushd /tmp
+      tar -zxf gpbackup_all.tar.gz
+      scp gpbackup_allsegments/gpbackup_mdw.tar.gz mdw:/tmp
+      scp gpbackup_allsegments/gpbackup_sdw1.tar.gz sdw1:/tmp
+    popd
+    ssh -t sdw1 'pushd /tmp ; tar -xzf gpbackup_sdw1.tar.gz ; popd'
+    ssh -t mdw 'pushd /tmp ; tar -xzf gpbackup_mdw.tar.gz ; popd'
+
+    # restore the backedup data to a new cluster and generate a pg_dump.
+    # do not fail here because might be possible for gpbackup to fail but still produce the same dump diff
+    scp go/src/github.com/greenplum-db/gpbackup/ci/scripts/gprestore_and_dump.bash mdw:/home/gpadmin/gprestore_and_dump.bash
+    set +e
+      ssh -t mdw "bash /home/gpadmin/gprestore_and_dump.bash"
+    set -e
+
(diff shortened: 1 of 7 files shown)


ci/tasks/gprestore-regression.yml
```diff
@@ -0,0 +1,97 @@
...
+    ccp_src/scripts/setup_ssh_to_cluster.sh
+    cp -r backup_artifact_resource/* /tmp
+    cp -r regression_dump/* /tmp
+
+    pushd /tmp
+      tar -zxf gpbackup_all.tar.gz
+      scp gpbackup_allsegments/gpbackup_mdw.tar.gz mdw:/tmp
+      scp gpbackup_allsegments/gpbackup_sdw1.tar.gz sdw1:/tmp
+    popd
+    ssh -t sdw1 'pushd /tmp ; tar -xzf gpbackup_sdw1.tar.gz ; popd'
```
larham: Consider that a reader might find it odd not to handle all segments+master in the same way:  it might be more readable to untarring both segment and master in commands right next to one another

ci/tasks/gprestore-regression.yml
```diff
@@ -0,0 +1,97 @@
...
+      set -ex
+      source env.sh
+
+      # extract backup tarball
+      pushd /tmp
+        tar -xzf gpbackup_mdw.tar.gz
+        ls backup_artifact/ | sort -g > /tmp/db_indices
+
+        # for each DB, find the timestamp & restore
+        # while read db_index ; do
```
larham: can this commented section be deleted?

ci/tasks/gprestore-regression.yml
```diff
@@ -0,0 +1,97 @@
...
+        # [ -z "\${db_index}" ] && continue
+        db_index=17 # 17 is the regression database
```
larham: Is there any failsafe to validate that 17 is still the right index?  e.g., look into the sql file for db name?

ci/tasks/gprestore-regression.yml
```diff
@@ -0,0 +1,97 @@
...
+        echo "##### pg_dump #####"
+        pg_dump regression -f /tmp/post_regression_dump.sql --schema-only
+        set -e
+        xz -z /tmp/post_regression_dump.sql
+      popd
+    SCRIPT
+
+    chmod +x /tmp/gprestore_and_dump.bash
+    scp /tmp/gprestore_and_dump.bash mdw:/home/gpadmin/gprestore_and_dump.bash
+    set +e
```
larham: consider indent for wrapping "+e/-e"

ci/tasks/gprestore-regression.yml
```diff
@@ -71,14 +71,14 @@ run:
...
-
     xz -d /tmp/regression_dump.sql.xz
     xz -d /tmp/post_regression_dump.sql.xz
     set +e
+
+    # Compare sqldump resource and the pg_dump that was newly generated
     diff -u /tmp/regression_dump.sql /tmp/post_regression_dump.sql > /tmp/diff.txt
-    # the diff file in the repo contains only trivial diffs, as described in README_regression.md
+
+    # The diff file in the repo contains only trivial diffs, as described in README_regression.md
```
larham: comment wording:  consider "Chop off timestamp and header/footer info so that we can test whether the diff file..."

tyacovone:
@larham @kyeap-pivotal 
Adding scripts provides much more clarity, thanks for this addition Larry. 
I fixed a minor typo in my first commit and then added a psql file load redirect to provide cleaner build output (if we want to see what was loaded, this info is easily accessible via input/output resources). All commits will be squashed when this PR is approved.

tyacovone:
> @tyacovone I like the idea of having a shorter output. The one issue I imagine is that this load takes a long time... 10 minutes? ... during which there would be no output indication of progress.

It looks like the last execution took ~5-6mins, during which there is WARNING, NOTICE, and ERROR output (which, as a separate matter, might create the illusion of an "real" issue ...). We could include a note about the estimated time required to load (e.g. "est. 5-10mins"). I don't think dynamically displaying the progress of the load would be worth our time to implement

PR #232, opened 2019-03-20
sense if gpexpand is running and refuse to backup if so

larham:


backup/gpexpand_sensor.go
```diff
@@ -0,0 +1,54 @@
...
+	// error has 3 possible states:
+	// nil: file exists
+	// error is "file not exists" kind of error
+	// real err that something went wrong
+	if err == nil {
+		return true, nil
+	}
+	if os.IsNotExist(err) {
+		// more work: check if it is in "phase 2" where a query will tell:
+		_, err = dbconn.SelectString(sensor.conn, GpexpandStatusQuery)
```
chrishajas: Is the intention of this PR to prevent gpbackup from running if either 1) the cluster is expanding or 2) there are any tables in the database (or cluster?) that have not yet been expanded? 

Is the intention to exit gpbackup if this query succeeds (that is, if the `gpexpand.status` table merely exists)?  Does this table exist in each database or just the `postgres` database? It's running `SELECT status FROM gpexpand.status ORDER BY updated DESC LIMIT 1`, yet the result of this query doesn't seem to be used anywhere.

If possible, please update the commit message to include this information/what circumstances we're preventing/allowing a user to run gpbackup wrt cluster/table expansion.

pivotal-ning-yu: The `gpexpand.status` table only exists in `postgres` database on gpdb 6 and master, can we ensure that `sensor.conn` is to the correct db?

larham: Pushed a commit to address these excellent comments:

* require and supply a connection to "postgres" database
* added comments and renamed query constant to clarify that the "phase 2" status is a query of a temporary table, so ANY successful reply indicates gpexpand is ongoing.

backup/gpexpand_sensor.go
```diff
@@ -0,0 +1,71 @@
...
+	"path/filepath"
+
+	"github.com/pkg/errors"
+
+	"github.com/blang/vfs"
+	"github.com/greenplum-db/gp-common-go-libs/dbconn"
+)
+
+const MasterDataDirQuery = `select datadir from gp_segment_configuration where content=-1 and role='p'`
+const GpexpandTemporaryTableStatusQuery = `SELECT status FROM gpexpand.status ORDER BY updated DESC LIMIT 1`
```
chrishajas: This doesn't need the `ORDER BY updated DESC` clause.

kyeap-pivotal: Removed order by clause. Thanks.

kyeap-pivotal:
There are only simulated unit and end-to-end tests for this feature because setting a test up to actually expand a cluster is such a significant cost. As a consequence we have leveraged existing tests from the gpexpand pipeline in order to hijack containers and test this feature manually, to ensure gpbackup does not during both phase1 and phase2 of gpexpand.

PR #235, opened 2019-04-17
Add job for testing 5X HEAD with fixed gpbackup version

tyacovone:
Since we are no longer continuing to rev the version of gpbackup that is
packaged with GPDB5, we need to be testing that the fixed version
~~(1.11.0)~~(1.12.1) of gpbackup is always compatible with the most recent commit of
GPDB5.


ci/pipeline.yml
```diff
@@ -79,6 +79,12 @@ resources:
     branch: 5X_STABLE
     tag_filter: 5.*
 
+- name: gpdb5_src_untagged
```
kyeap-pivotal: nit: instead of `gpdb5_src_untagged` can we use the name `gpdb5_src_head`?

ci/pipeline.yml
```diff
@@ -828,6 +846,45 @@ jobs:
...
+      gpdb_binary: bin_gpdb_5x_stable_intermediate
+      gpdb_src: gpdb5_src_untagged
+  - task: gpinitsystem
+    file: ccp_src/ci/tasks/gpinitsystem.yml
+  - task: setup-centos-env
+    file: gpbackup/ci/tasks/setup-centos-env.yml
+  - task: gpbackup-1.11.0-test
+    file: gpbackup/ci/tasks/integration-tests-fixed-version.yml
+    params:
+      GPBACKUP_VERSION: "1.11.0"
```
kyeap-pivotal: This will need tweaking due the recent event of backward compatibility issues.

ci/pipeline.yml
```diff
@@ -828,6 +846,45 @@ jobs:
...
+  - aggregate:
+    - get: gpbackup
+    - get: bin_gpdb_5x_stable_intermediate
+      trigger: true
+    - get: ccp_src
+    - get: gpdb5_src_untagged
+    - get: gpbackup-dependencies
+      passed:
+       - build_dependencies
+  - put: terraform
```
kyeap-pivotal: because this is running on every commit to 5X_STABLE. Do we want to setup a local cluster?

ci/pipeline.yml
tyacovone: updated to use 1.12.1

ci/pipeline.yml
tyacovone: I like it. Changed

kyeap-pivotal:
After pipeline refactor changes are merged in we'll need to update this commit before merging.

yijunkeun:
An updated PR is in https://github.com/greenplum-db/gpbackup/pull/235

This PR reflects the new templatized pipeline generation pattern.
Please reference the 5x-test branch.

PR #236, opened 2019-04-23
plugin: record API and plugin version within backup report, and supply this same information (during restore) to the plugin

larham:
During backup, save to the history file a record of the version of any plugin used.  During restore, add this information via an attribute of the plugin_config file supplied during restore.

* NEW FEATURE: previously, plugins were only checked to exist have sufficient API version, but there was no check that they were all of the same version. Now, there will be a fatal error if plugin versions are not uniform across all segments.

* plugin_config.yaml attribute name is "backup_plugin_version"
* it will have empty "" contents if there is no record of the version used when backing up; output a warning to log file

### DIFF
--- a/restore/wrappers.go
+++ b/restore/wrappers.go
@@ -142,7 +142,12 @@ func RecoverMetadataFilesUsingPlugin() {
 	var err error
 	pluginConfig, err = utils.ReadPluginConfig(MustGetFlagString(utils.PLUGIN_CONFIG))
 	gplog.FatalOnError(err)
+
 	pluginConfig.CheckPluginExistsOnAllHosts(globalCluster)
+
+	timestamp := MustGetFlagString(utils.TIMESTAMP)
+	historicalPluginVersion := FindHistoricalPluginVersion(timestamp)
+	pluginConfig.SetBackupPluginVersion(timestamp, historicalPluginVersion)
 
 	pluginConfig.CopyPluginConfigToAllHosts(globalCluster)
 	pluginConfig.SetupPluginForRestore(globalCluster, globalFPInfo)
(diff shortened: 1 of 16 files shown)


khuddlefish:
This looks pretty good :)

One thing I found slightly confusing was that many places in the code you are referring to `plugin version` not `plugin API version`. Since `plugin version` is actually a real thing that means something different than `plugin API version` it could be a bit confusing to future developers. I think your logs mostly said `API` so it's probably ok for customers. 

I have a quick question also. Are you checking the backup plugin version anywhere during restore currently? (not saying you should be, just want to make sure I understand the changes)

larham:
Thanks, @khuddlefish , the code should make more clear what version is being employed, and I think you are pointing out a problem with some of the logic.

Some of the motivation for this story is at https://github.com/greenplum-db/gpbackup/pull/203 . We hit an issue with being able to do backwards compatibility with plugins; we had to pause such tests. That led to discussion of how to handle plugin compatibility in general. To adjust for backwards compatibility, plugins need 2 things: knowledge of the provenance of old backups, and also code to make adjustments.

I'll adjust some naming and discuss the compatibility approach with team, thanks.

restore/wrappers.go
```diff
@@ -173,6 +178,24 @@ func RecoverMetadataFilesUsingPlugin() {
 	}
 }
 
+func FindHistoricalPluginVersion(timestamp string) string {
```
chrishajas: Do we actually need this function? Isn't the plugin version information already in backupConfig? As far as I remember, the history file simply contains a list of the backup configs for each timestamp. The reason we used it for incremental backups was so we could compare the dbname without having to pull down each backup config locally. We don't strictly enforce/assume that the history file must exist during restore, so this would be a new assumption.

larham: In order to access a backup, including its metadata backupConfig file, a plugin must get to those files in its storage.  If the plugin is doing some kind of backwards compatibility, then access to those files might imply a chicken/egg issue: in order to know to do the backwards compatibility to access the files, it must read the files first.  Thus, the history file seems like a better place to look for versioning info.

I agree that this access of the history file implies another (minor) dependence on the history file. If not found, in this case, the loss is not fatal, and the plugin will succeed if the backup is already compatible with its current version.

chrishajas: I see, thanks! Looks good then!

utils/plugin.go
```diff
@@ -277,3 +293,16 @@ func (plugin *PluginConfig) RestoreSegmentTOCs(c *cluster.Cluster, fpInfo backup
...
+	return plugin.backupPluginVersion
+}
+
+func (plugin *PluginConfig) SetBackupPluginVersion(timestamp string, historicalPluginVersion string) {
+	if historicalPluginVersion == "" {
+		gplog.Warn("cannot recover plugin version from history using timestamp %s, so using current plugin version. This is fine unless there is a backwards compatibility consideration within the plugin", timestamp)
+		plugin.backupPluginVersion = ""
+	} else {
+		plugin.backupPluginVersion = historicalPluginVersion
+	}
```
kyeap-pivotal: nit:
```
	if historicalPluginVersion == "" {
		gplog.Warn("cannot recover plugin version from history using timestamp %s, so using current plugin version. This is fine unless there is a backwards compatibility consideration within the plugin", timestamp)
	}
	plugin.backupPluginVersion = historicalPluginVersion
```

Also it doesn't seem like this warning belongs here. Maybe in `FindHistoricalPluginVersion`?

PR #238, opened 2019-05-16
Require 'delete_backup' command in plugin test bench

tyacovone:
* Removal of the conditional statement (which skipped delete tests when provided an
insufficient version) caused an indentation shift, causing a messy git
diff
* Also added a test to ensure that unknown commands exit non-zero

### DIFF
--- a/plugins/plugin_test_bench.sh
+++ b/plugins/plugin_test_bench.sh
@@ -185,95 +185,100 @@ cleanup_test_dir $testdir
 # ----------------------------------------------
 # Delete backup directory function
 # ----------------------------------------------
+time_second_for_del=$(date +"%Y%m%d%H%M%S")
+current_date_for_del=$(echo $time_second_for_del | cut -c 1-8)
+sleep 1
+time_second_for_del2=$(date +"%Y%m%d%H%M%S")
+current_date_for_del2=$(echo $time_second_for_del | cut -c 1-8)
 
-# `awk` call returns 1 for true, 0 for false (contrary to bash logic)
-if (( 1 == $(echo "0.4.0 $api_version" | awk '{print ($1 > $2)}') )) ; then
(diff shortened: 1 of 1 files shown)


larham:
Nice work. please do not push until s3 plugin gets this delete functionality

plugins/plugin_test_bench.sh
```diff
@@ -185,95 +185,100 @@ cleanup_test_dir $testdir
...
+$plugin backup_file $plugin_config $testfile_for_del2
+
+# here's the real test: can we delete and confirm deletion, while sibling backup remains?
+set +e
+$plugin delete_backup $plugin_config $time_second_for_del
+del_cmd_exit_code=`echo $?`
+
+if [ $del_cmd_exit_code -ne 0 ] ; then
+  # bash exit code 127 == command not found
+  if [ $del_cmd_exit_code -eq 127 ] ; then
```
kyeap-pivotal: I think an error code of 127 does mean command not found. However, It doesn't seem to apply to subcommands. The tests we tried on the command line were
```
  |2.3.1| melville in ~/go/src/github.com/greenplum-db/gpbackup-s3-plugin
± lh |master {5} ?:1 ✗| → foo
-bash: foo: command not found

  |2.3.1| melville in ~/go/src/github.com/greenplum-db/gpbackup-s3-plugin
± lh |master {5} ?:1 ✗| → echo $?
127
```
and 
```
  |2.3.1| melville in ~/go/src/github.com/greenplum-db/gpbackup-s3-plugin
± lh |master {5} ?:1 ✗| → gpbackup_ddboost_plugin del /Users/pivotal/ddboost_backup_config.yml 12309127
Invalid argument: del

  |2.3.1| melville in ~/go/src/github.com/greenplum-db/gpbackup-s3-plugin
± lh |master {5} ?:1 ✗| → echo $?
1
```

PR #242, opened 2019-06-05
resgroup: retire the `10 MB` syntax of memory_spill_ratio

pivotal-ning-yu:
In previous commits of gpdb we introduced a new syntax for
memory_spill_ratio, which allow to set the value in `10 MB` syntax, in
gpbackup we added support for this syntax.

However to simplify the user interface we retired this new syntax in
gpdb, so we now have to change in gpbackup accordingly.

Default resource group settings on gpdb 5 and 6 and later, however, are
different now, so we have to provide different default answer values in
the intergration tests.

Co-authored-by: Hubert Zhang <hzhang@pivotal.io>
Co-authored-by: Ning Yu <nyu@pivotal.io>


larham:
Merging to match the changed already pushed to master.

Thanks for the update!

PR #243, opened 2019-06-12
Test ddboost plugin on 4.3_STABLE

kyeap-pivotal:
Testing the entirety of boostfs with ddboost plugin was probably overkill in hindsight.

### DIFF
--- a/ci/pipeline.yml
+++ b/ci/pipeline.yml
@@ -10,7 +10,8 @@ groups:
   - master
   - s3_plugin_tests
   - backward-compatibility
-  - ddboost_plugin_and_boostfs_tests
+  - ddboost_plugin_and_boostfs_tests_43
+  - ddboost_plugin_and_boostfs_tests_5x
   - scale-43
   - scale-5x
   - scale-6x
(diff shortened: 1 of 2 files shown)


ci/pipeline.yml
```diff
@@ -468,7 +479,202 @@ jobs:
...
+            pgport: 5432
+            remote_hostname: {{datadomain_dest_host}}
+            remote_username: {{datadomain_user}}
+            remote_storage_unit: GPDB
+            remote_directory: gpbackup_tests
+            password: {{encrypted_datadomain_password}}
+            password_encryption: "on"
+            remote_password: {{encrypted_datadomain_password}}
+            remote_password_encryption: "on"
+            gpbackup_ddboost_plugin: 66706c6c6e677a6965796f68343365303133336f6c73366b316868326764
```
larham: Looks a little odd to have replacements for everything but the `gpbackup_ddboost_plugin` attribute, but not a deal breaker

kyeap-pivotal:
> looks good to me. One improvement would be to use a shared task yml for 4.3 & 5 rather than having near duplicate in-line `run_tests` steps, but this is not functionally necessary.

This also means we would have to duplicate the entire cluster generation too to get it running in parallel. For this reason alone I figured the increase in readability complexity was not worth the couple of lines saved.

tyacovone:
> > looks good to me. One improvement would be to use a shared task yml for 4.3 & 5 rather than having near duplicate in-line `run_tests` steps, but this is not functionally necessary.
> 
> This means we would have to duplicate the entire cluster generation too to get it running in parallel. For this reason alone I figured the increase in readability complexity was not worth the couple of lines saved.

I was suggesting keeping the jobs separate, but adding a task yml file which is flexible(no parallel clusters). I can add it to my to-dos and pick it up as a minor refactor in the future – im cool with merging this

ci/pipeline.yml
jimmyyih: I agree with this.  Assuming the value here is a SHA to a gpbackup_ddboost_plugin, we would have to keep editing and pushing this file instead of just changing the secrets file.

PR #244, opened 2019-06-20
build binaries & gppkgs in preliminary job, reuse those during testing

larham:
* update pivnet job to push with file groups
* use single-container "local" clusters where possible, instead of terraform VMs, where possible

This work is not finished, but the dev pipeline is green, and it seems worthy of getting the current state into master, instead of a long-running branch.


ci/pipeline.yml
```diff
@@ -99,6 +100,7 @@ resources:
     start: 6:00 AM
     stop: 7:00 AM
 
+# TODO mark these as src with name change
```
tyacovone: unresolved _TODO_. We should add `_src` to gpbackup and both the plugin resources below for consistency

ci/pipeline.yml
```diff
@@ -372,6 +381,195 @@ resources:
...
+            # cp $GOPATH/src/github.com/greenplum-db/gpbackup/ci/scripts/smoke_test_component_gpbackup.sh smoke_test_gpdb_component
+            # # Assemble binaries
+            # mkdir -p bin
+            # cp $GOPATH/bin/gpbackup bin/
+            # cp $GOPATH/bin/gpbackup_helper bin/
+            # cp $GOPATH/bin/gprestore bin/
+            # cp $GOPATH/bin/gpbackup_s3_plugin bin/
+            # cp $GOPATH/bin/gpbackup_manager bin/
+
+            # need to end up with a folder bin with everything
```
tyacovone: will this section of comments be useful in the future? As far as I know, we should be good to remove this section

ci/tasks/test-on-local-cluster.yml
```diff
@@ -63,16 +64,33 @@ run:
...
+      set -e
+      gppkg -i /tmp/untarred/gpbackup*gp\$gpdbversion*.gppkg
+    fi
+    set -e
+    
     # Build gpbackup
     pushd \$GOPATH/src/github.com/greenplum-db/gpbackup
       make depend
-      make build
+      # make build
```
tyacovone: nit: remove this line

ci/pipeline.yml
```diff
@@ -1535,57 +1535,60 @@ jobs:
...
+          trigger: true
+        - get: gpbackup_s3_plugin_tagged_src
+        - get: gpbackup_ddboost_plugin_tagged_src
+        - get: github_release_components
+          passed:
+            - build_binaries
+        # While the resource is not used, it ensures only releasing if
+        # tests are passing
+        - get: gppkgs
+          passed:
```
tyacovone: this job should be a gate for all tests in our pipeline (with the exception of `scale` tests). We should reinstate these passed conditions & ensure all jobs are accounted for

ci/pipeline.yml
```diff
@@ -19,7 +19,7 @@ groups:
   - scale-master
   - build_dependencies
   - push-to-github
-  - package_gpbackup_for_release
+  - gppkgs_smoketest
```
tyacovone: nit: commit message is inconsistent with job name: `{gppkg,gppkgs}_smoketest`.

ci/pipeline.yml
```diff
@@ -1535,57 +1535,60 @@ jobs:
...
+            # - GPDB6
+            - GPDB4.3
+            # - GPDB5-sles11
+            # - GPDB5-oracle7
+            - master
+            # - backward-compatibility
+            - s3_plugin_tests
+            # - ddboost_plugin_and_boostfs_tests
+        - get: gpbackup
+      # - get: gpdb6_src
```
tyacovone: the commented out lines below from copy/pasted dev work and will not be useful given the new 'build binaries' pattern. Lets remove them

ci/pipeline.yml
```diff
@@ -1623,10 +1632,11 @@ jobs:
   plan:
     - aggregate:
       - get: gppkgs
-        passed: [gppkgs_smoketest]
+        # passed: [gppkgs_smoketest]
```
tyacovone: we want this passed condition

Makefile
```diff
@@ -46,7 +46,7 @@ integration :
 
 test : lint unit integration
 
-end_to_end : build
```
kyeap-pivotal: Given today's discussion about creating a new makefile rule end_to_end_without_build that does not rebuild/reinstall over the binaries installed by our gppkgs.

ci/tasks/test-on-local-cluster.yml
```diff
@@ -67,39 +67,27 @@ run:
     out=\$(psql postgres -c "select version();")
     gpdbversion=\$(echo \$out | sed -n 's/.*Greenplum Database \([0-9]\).*/\1/p')
 
-    # We do not generate gppkgs for GPDB7/master, therefore we must compile during testing
```
tyacovone: This fixup seems to undo commit [2cfbb84](https://github.com/greenplum-db/gpbackup/pull/244/commits/2cfbb845b230ffec3f3784626c5af40c06905d32). Given that both commits exclusively edit this file, consider squashing.

However take note that the one-line addition [here](https://github.com/greenplum-db/gpbackup/pull/244/commits/d7ef64b2cc36bb2336beca7ee02fcc131ecf9369#diff-92376c8ebe48632ec920a4020a33fd94R57) in commit d7ef64b is what allowed for this fixup. This single addition can be squashed into commit `59bafb2 - Add build_binaries job to pipeline`

larham: fixed

Makefile
larham: done

ci/pipeline.yml
larham: now it is called "green_gate".  the version smoketest is done during the build phase.

ci/pipeline.yml
larham: done

PR #246, opened 2019-06-26
Create gpdb4 regression database and backup using gpbackup

kyeap-pivotal:
This is done using a replica of the icg_orca_centos job from gpdb4 pipeline

### DIFF
--- a/ci/regression_pipeline.yml
+++ b/ci/regression_pipeline.yml
@@ -1,4 +1,15 @@
 ---
+groups:
+- name: 6-to-6
+  jobs:
+  - build_dependencies
+  - gpbackup-gpdb6-ICW-regression-DB
+  - gprestore-gpdb6-ICW-regression-DB
+- name: 4-to-6
+  jobs:
+  - build_dependencies
+  - gpbackup_gp43_icg_regression
+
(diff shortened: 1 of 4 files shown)


ci/regression_pipeline.yml
```diff
@@ -123,6 +200,59 @@ jobs:
...
+      - get: gpdb_src
+        resource: gpdb43_src
+      - get: sync_tools_gpdb
+        resource: sync_tools_gpdb_centos
+      - get: bin_gpdb
+        resource: bin_gpdb4_centos6
+        trigger: true
+      - get: centos-gpdb-dev-5
+  - task: ic_gpdb4_backup
+    image: centos-gpdb-dev-5
```
tyacovone: I believe in LL we were told that rhel5 is no longer supported - do we still want to do the gp4 backup on this image?

jimmyyih: We should continue to use centos5 until the GPDB4 pipeline stops using it for its compilation.  Most likely... it will not stop using centos5 seeing how we will be working on 4 branches (4.3, 5, 6, and master) and very little attention will be given to 4.3... probably.

PR #247, opened 2019-06-27
Create release pipeline; templatize `gpbackup` pipeline

tyacovone:
We are now using Jinja to generate pipeline yml files, since the
gpbackup pipeline (testing against HEAD) and the gpbackup-release
pipeline(testing tagged releases) are very similar.

We now use tag_filtering to ensure that the `gpbackup` pipeline always tests
HEAD against HEAD and the `gpbackup-release` pipeline tests only tagged
versions (of DPM components *and* GPDB versions).
  * DDBoost is an exception to this pattern, as a tagged version of
    DDBoost is required for compilation. The version being tested
    remains consistent, it is only a one-off for the build_binaries job.

Modified nightly-trigger-off to nightly-trigger.
The default behavior is for nightly-trigger to be enabled by default for prod and disabled for dev pipelines

Other changes:
* Add pass condition for build_dependencies in build_binaries
* Remove scale tests from 'All' group
* Remove nightly-triggers & add cache trigger in release-pipeline
* Trigger tests on new gpdb binaries
* Add unique bucket for final github release candidates


ci/gpbackup-release-generated.yml
```diff
@@ -426,7 +448,7 @@ jobs:
             popd
             echo ${version} > go_components/gpbackup_version
 
-            if [[ "gpbackup version ${version}" != "$($GOPATH/bin/gpbackup --version)" ]]; then
+            if [ "gpbackup version ${version}" != "$($GOPATH/bin/gpbackup --version)" ]; then
```
yijunkeun: These double brackets for string comparisons were changed to single brackets because they were erroneously being resolved as Jinja expressions.

PR #248, opened 2019-07-03
Fix flaking resource group test

kyeap-pivotal:
The correct global variable defaults were not being properly set for one
of test. Depending on the order in which tests ran, a default resource
group test had the potential to flake. Move the setting of the GUCs to
before when all integration tests run.

The other test using the global defaults variables `memSharedDefault` and `memSpillDefault` are in metadata_globals_queries_test.go

### DIFF
--- a/integration/metadata_globals_create_test.go
+++ b/integration/metadata_globals_create_test.go
@@ -14,11 +14,7 @@ import (
 )
 
 var _ = Describe("backup integration create statement tests", func() {
-	var includeSecurityLabels bool
 	BeforeEach(func() {
-		if connectionPool.Version.AtLeast("6") {
-			includeSecurityLabels = true
-		}
 		toc, backupfile = testutils.InitializeTestTOC(buffer, "predata")
 	})
 	Describe("PrintCreateDatabaseStatement", func() {
--- a/integration/metadata_globals_create_test.go
+++ b/integration/metadata_globals_create_test.go
@@ -13,20 +13,10 @@ import (
 	"github.com/onsi/gomega/gbytes"
 )
 
-var (
-	concurrencyDefault = "20"
-	memSharedDefault   = "20"
-	memSpillDefault    = "20"
-	memAuditDefault    = "0"
-	cpuSetDefault      = "-1"
-)
-
 var _ = Describe("backup integration create statement tests", func() {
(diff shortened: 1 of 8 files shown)


jimmyyih:
What is the order of operations here for this test framework to make the defaults incorrect?  The original `BeforeEach()` call seems to be correct in setting up each test (with my very little understanding of this test framework).  Can you describe the ordering issue more?

integration/metadata_globals_create_test.go
```diff
@@ -13,20 +13,10 @@ import (
...
-	cpuSetDefault      = "-1"
-)
-
 var _ = Describe("backup integration create statement tests", func() {
 	var includeSecurityLabels bool
 	BeforeEach(func() {
 		if connectionPool.Version.AtLeast("6") {
-			memSharedDefault = "80"
-			memSpillDefault = "0"
 			includeSecurityLabels = true
```
tyacovone: the gpdb version will not vary between individual integration tests – I'd suggest declaring/setting this variable in the BeforeSuite clause, alongside the GUCs

kyeap-pivotal:
@jimmyyih ginkgo allows a tests of a package to be run in random order with `ginkgo`'s 
`--randomizeAllSpecs` flags. The `creates a resource group with defaults` test in   `metadata_globals_queries_test.go` (different file not edited in this PR), will flake depending on if any of the tests in `metadata_globals_create_test.go` have run, because those tests are setting the `memSharedDefault` and `memSpillDefault` globals variables.

Flake in action
```
creates a resource group with defaults
returns a resource group with defaults

Ran 2 of 488 Specs in 5.303 seconds
SUCCESS! -- 2 Passed | 0 Failed | 0 Pending | 486 Skipped
```
same tests but reversed
```
returns a resource group with defaults
   Mismatch on field MemorySharedQuota
   ...
   Mismatch on field MemorySpillRatio
   ...
creates a resource group with defaults

Ran 2 of 488 Specs in 5.166 seconds
FAIL! -- 1 Passed | 1 Failed | 0 Pending | 486 Skipped
```

integration/metadata_globals_create_test.go
kyeap-pivotal: Added 1 more commit to address your comment here.

PR #250, opened 2019-07-10
Set boostfs mounted dir

kyeap-pivotal:
Adding a duplicate boostfs job to run against GPDB43 introduced a test
flake because the two jobs were saving backups to the same directory on
DataDomain. This change fixes that.


shivzone:
This would be a lot more flake proof if the sink location is specific to the job (maybe based on some unique job id or timestamp) /data/gpdata/dd_dir/<id>. This way even if we were to multiple pipelines which exercise end to end we would stay flake free

jimmyyih:
Agree with Shivram here.  Maybe we can just add a random number or hash at the end to make it super unique.  It should be easy (but fair warning, using /dev/urandom or RANDOM is not really random unless you change the entropy on the box).

kyeap-pivotal:
Appended $RANDOM to the custom backup directory.

PR #252, opened 2019-07-11
Pass in same endtime for history and report file

yijunkeun:
The backup-history and backup-report files were previously resolving
each of its endtime field individually, which created inconsistencies
for the same backup. We now pass in the endtime field of the history
file into the report file to sync up the two files.


kyeap-pivotal:
This is probably what caused the backup report and and backup history to be out sync by 1 second which is what @oakbarrett was seeing.

yijunkeun:
> This is probably what caused the backup report and and backup history to be out sync by 1 second which is what @oakbarrett was seeing.

Yes!

PR #255, opened 2019-07-15
Modify pipeline template for scale tests

yijunkeun:
~~1. The template for the pipeline generation does not include the scale
tests as pass conditions for its final validation step.~~

2. Scale jobs were tested against a specific GPDB major versions but were not
    added to the pipeline group for that version.

### DIFF
--- a/ci/templates/gpbackup-tpl.yml
+++ b/ci/templates/gpbackup-tpl.yml
@@ -30,11 +30,13 @@ groups:
   jobs:
   - GPDB4.3
   - ddboost_plugin_and_boostfs_tests_43
+  - scale-43
 - name: GPDB5
   jobs:
   - GPDB5
   - GPDB5-oracle7
   - GPDB5-sles11
+  - scale-5x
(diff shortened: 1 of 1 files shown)


yijunkeun:
> I like the redistribution of the scale jobs to be within the GPDB major groups. Under this grouping, we could/should remove the `scale` group.
> 
> In terms of the other changes, we intentionally decided to exclude scale jobs from the final gates. We can discuss this as a team if we want to reevaluate, but previously we had agreed that the scale jobs (which run 4-7hrs) shouldn't block a release.

Cool, makes sense

yijunkeun:
@tyacovone Actually, since the scale tests aren't part of the 'All' group, shouldn't we keep reference to them in the 'scale' group to track them?

tyacovone:
> @tyacovone Actually, since the scale tests aren't part of the 'All' group, shouldn't we keep reference to them in the 'scale' group to track them?

Either way, I dont have a strong preference. No harm in leaving the 'scale' group if you'd like.

PR #257, opened 2019-07-15
Add gpbackup-manager to gppkgs and create CI test job

yijunkeun:
The gpbackup-manager binary is now packaged with the other binaries into
a single gppkg. The test for gpbackup-manager runs against GPDB6.

### DIFF
--- a/ci/tasks/test-gpbackup-manager.yml
+++ b/ci/tasks/test-gpbackup-manager.yml
@@ -1,0 +1,86 @@
+platform: linux
+
+image_resource:
+  type: docker-image
+  source:
+    repository: pivotaldata/centos-gpdb-dev
+    tag: '7-gcc6.2-llvm3.7'
+
+inputs:
+- name: gpbackup_manager_src
+  path: go/src/github.com/pivotal/gp-backup-manager
+- name: gpdb_src
+- name: bin_gpdb
+- name: gppkgs
+
+params:
+  GOLANG_VERSION: 1.12.7
+
+run:
+  path: bash
+  args:
+  - -c
+  - |
+    set -ex
+    if [ ! -f bin_gpdb/bin_gpdb.tar.gz ] ; then
+      mv bin_gpdb/*.tar.gz bin_gpdb/bin_gpdb.tar.gz
+    fi
+
+    # We download a fixed version GOLANG_VERSION of Go to run ginkgo, our testing framework, on gpbackup_manager.
+    # This is because gpbackup_manager has tests that require Go 1.12+
+    # TODO: bake a new image with Go 1.12+ pre-installed.
+    wget https://storage.googleapis.com/golang/go${GOLANG_VERSION}.linux-amd64.tar.gz && \
+    sudo rm -rf /usr/local/go && \
+    tar -xzf go${GOLANG_VERSION}.linux-amd64.tar.gz -C /usr/local
+
+    source gpdb_src/concourse/scripts/common.bash
+    time install_gpdb
+    time ./gpdb_src/concourse/scripts/setup_gpadmin_user.bash
+    time make_cluster
+
+    source /usr/local/greenplum-db-devel/greenplum_path.sh
+
+    # copy gpbackup-manager into the GOPATH used by user "gpadmin"
+    export GOPATH=/home/gpadmin/go
+    mkdir -p $GOPATH/src/github.com/pivotal
+    cp -R go/src/github.com/pivotal/gp-backup-manager $GOPATH/src/github.com/pivotal/
+
+    chown -R gpadmin $GOPATH
+
+    cat <<SCRIPT > /tmp/run_tests.bash
+    #!/bin/bash
+    set -ex
+    cd ~
+    source /usr/local/greenplum-db-devel/greenplum_path.sh
+
+    # use "temp build dir" of parent shell
+    source $(pwd)/gpdb_src/gpAux/gpdemo/gpdemo-env.sh
+    export GOPATH=\$HOME/go
+    mkdir -p \$GOPATH/bin
+    mkdir -p \$GOPATH/src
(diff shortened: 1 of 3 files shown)


ci/tasks/test-gpbackup-manager.yml
```diff
@@ -0,0 +1,96 @@
...
+  path: bash
+  args:
+  - -c
+  - |
+    set -ex
+    if [ ! -f bin_gpdb/bin_gpdb.tar.gz ] ; then
+      mv bin_gpdb/*.tar.gz bin_gpdb/bin_gpdb.tar.gz
+    fi
+
+    wget https://storage.googleapis.com/golang/go${GOLANG_VERSION}.linux-amd64.tar.gz && \
```
tyacovone: This is an antipattern. It may be worth creating an image with golang version >= 1.12 to run this job on.

If we stick with a manual download, we should make a note here about why the more up-to-date golang version is needed (`strings` package)

ci/templates/gpbackup-tpl.yml
```diff
@@ -864,6 +865,32 @@ jobs:
...
+- name: gpbackup-manager-tests
+  plan:
+    - aggregate:
+        - get: gpbackup_manager_src
+          trigger: true
+        - get: gpbackup
+        - get: bin_gpdb_6x_stable_centos6
+        - get: ccp_src
+        - get: gpdb6_src
+        - get: gppkgs
```
tyacovone: `gppkgs` should be the trigger - not the src code `gpbackup_manager_src`

ci/templates/gpbackup-tpl.yml
```diff
@@ -864,6 +865,32 @@ jobs:
...
+          passed:
+            - build_binaries
+        - get: gpbackup-dependencies
+          passed:
+            - build_dependencies
+  {% if nightly_trigger and "gpbackup" == pipeline_name %}
+        - get: nightly-trigger
+          trigger: true
```
tyacovone: This mustve slipped through the cracks when `build_gppkgs` was added. The nightly trigger should only apply to the build job, not the test jobs... I'll fix it in the template for all preexisting jobs prior to merging this in

tyacovone: Pushed. Please rebase.

yijunkeun: Okay

ci/templates/gpbackup-tpl.yml
yijunkeun: Nice catch

ci/tasks/test-gpbackup-manager.yml
```diff
@@ -0,0 +1,96 @@
...
+    sudo rm -rf /usr/local/go && \
+    tar -xzf go${GOLANG_VERSION}.linux-amd64.tar.gz -C /usr/local
+
+    source gpdb_src/concourse/scripts/common.bash
```
shivzone: A lot of these steps here seem similar to those in test-on-local-cluster.yml. Can we move the common steps to setup gpbackup to a separate script say setup-gpbackup.sh and call it from both the pipelines ?

ci/tasks/test-gpbackup-manager.yml
yijunkeun: Would using a golang docker image rather than the pivotal centos image work?

ci/tasks/test-gpbackup-manager.yml
yijunkeun: Yes, in fact the testing on remote clusters can be refactored too, and is probably good to think of as a chore.

shivzone: makes sense

yijunkeun:
https://github.com/greenplum-db/gpbackup/pull/257/files#diff-6ce55adbe1802a8ef0d717bbbe291c44L21

Where backup manager is being added to the gppkgs

ci/tasks/test-gpbackup-manager.yml
```diff
@@ -0,0 +1,96 @@
...
+    mkdir -p \$GOPATH/bin
+    mkdir -p \$GOPATH/src
+    # reference PATH defined by parent shell
+    export PATH=/usr/local/go/bin:$PATH:\$PATH:\$GOPATH/bin
+
+    # Install gpbackup gppkg
+    mkdir /tmp/untarred
+    tar -xzf gppkgs/gpbackup-gppkgs.tar.gz -C /tmp/untarred
+
+    # only install if not installed already
```
kyeap-pivotal: This section of checking if it is already installed is not needed here since this is the only task that is installing a gppkg. This was added to other yml files because in some cases we are reusing a yml file to run the same set of tests on the same aws cluster. It looks like you are running tests on a local cluster.

ci/templates/gpbackup-tpl.yml
```diff
@@ -860,6 +861,29 @@ jobs:
...
+        - get: ccp_src
+        - get: gpdb6_src
+        - get: gppkgs
+          trigger: true
+          passed:
+            - build_binaries
+        - get: gpbackup-dependencies
+          passed:
+            - build_dependencies
+  {% endif %}
```
kyeap-pivotal: leftover code?

ci/tasks/test-gpbackup-manager.yml
yijunkeun: Done

ci/templates/gpbackup-tpl.yml
yijunkeun: Deleted, it was an error on my part when I was deleting the nightly triggers.

ci/tasks/test-gpbackup-manager.yml
yijunkeun: Added a comment, with a TODO for a new image.

PR #258, opened 2019-07-16
Output COPY ON SEGMENT error CONTEXT string

jimmyyih:
When COPY ON SEGMENT errors out on distribution key issues, the error
CONTEXT contains the actual tuple that COPY was having issues
with. This is not outputted by default though in pgx.PgError.Error()
so we have to output it ourselves.

Before:
Error loading data into table public.foobar: ERROR: value of distribution key doesn't belong to segment with ID 0, it belongs to segment with ID 1  (seg0 127.0.0.1:25432 pid=18337) (SQLSTATE 22P04)

After:
Error loading data into table public.foobar: COPY foobar, line 1: "5": ERROR: value of distribution key doesn't belong to segment with ID 0, it belongs to segment with ID 1  (seg0 127.0.0.1:25432 pid=18337) (SQLSTATE 22P04)

### DIFF
--- a/restore/data.go
+++ b/restore/data.go
@@ -13,6 +13,7 @@ import (
 	"github.com/greenplum-db/gp-common-go-libs/gplog"
 	"github.com/greenplum-db/gpbackup/backup_filepath"
 	"github.com/greenplum-db/gpbackup/utils"
+	"github.com/jackc/pgx"
 	"github.com/pkg/errors"
 	pb "gopkg.in/cheggaaa/pb.v1"
 )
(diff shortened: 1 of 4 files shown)


restore/data.go
```diff
@@ -13,6 +13,7 @@ import (
 	"github.com/greenplum-db/gp-common-go-libs/gplog"
 	"github.com/greenplum-db/gpbackup/backup_filepath"
 	"github.com/greenplum-db/gpbackup/utils"
+	"github.com/jackc/pgx"
```
kyeap-pivotal: This looks like a new dependency. The Gopkg.toml file and the lock file should be updated.

lavjain: Is the error functionality not available using https://github.com/lib/pq library?

jimmyyih: @kyeap-pivotal It's already there in the lock file but I guess it's not the standard way? In the Gopkg.lock file currently:
```
[[projects]]
  digest = "1:1cbbe90c54d3308204c955163a6dce9a328839a166aadd2d13dd26ab290162f8"
  name = "github.com/jackc/pgx"
  packages = [
    ".",
    "chunkreader",
    "internal/sanitize",
    "pgio",
    "pgproto3",
    "pgtype",
    "stdlib",
  ]
  pruneopts = "NUT"
  revision = "c59c9cac59ab95eceb0c12ff338923c62f411ea2"
  version = "v3.3.0"
```

I added to the Gopkg.toml file and ran `dep ensure`.  Let me know if that commit is right.

jimmyyih: @lavjain Changing it to `lib/pq` should technically work.  The `jackc/pgx` library was chosen because the `gp-common-go-libs` library (used for our database connection) is using `jackc/pgx` instead of `lib/pq` so I wanted to match the structs properly.

From `gp-common-go-libs/dbconn/dbconn.go`:
```
        /*
         * We previously used github.com/lib/pq as our Postgres driver,
         * but it had a bug with the way it handled certain encodings.
         * pgx seems to handle these encodings properly.
         */
```

PR #262, opened 2019-07-23
Package libDDBoost.so in the gpbackup gppkg

jimmyyih:
The libDDBoost.so file is packaged with GPDB 4.3 and 5X due to it
being a gpcrondump dependency. On GPDB 6X, gpcrondump is removed so we
don't get this for free anymore. The 6X DDBoost tests were only
passing because the plugin's `make test` was installing the
libDDBoost.so file.

We now package the libDDBoost.so file into the gpbackup RPM which gets
consumed in the gpbackup gppkg build process.


tyacovone:
Add `gpbackup-release-generated.yml` to the commit (created via `./gen_pipeline -p gpbackup-release --prod`).

Otherwise, LGTM.

PR #263, opened 2019-07-23
Fix getModCount return a wrong result.

hidva:
There may be multiple segno files for an append-only table.

```
gpadmin=# select * from  pg_aoseg.pg_aoseg_26762;
 segno | eof | tupcount | varblockcount | eofuncompressed | modcount | formatversion | state 
-------+-----+----------+---------------+-----------------+----------+---------------+-------
     1 |   0 |        4 |             0 |               0 |        4 |             3 |     1
     2 |   0 |        2 |             0 |               0 |        2 |             3 |     1
(2 rows)
```

### DIFF
--- a/backup/queries_incremental.go
+++ b/backup/queries_incremental.go
@@ -84,7 +84,7 @@ func getAOSegTableFQNs(connectionPool *dbconn.DBConn) map[string]string {
 
 func getModCount(connectionPool *dbconn.DBConn, aosegtablefqn string) int64 {
 	query := fmt.Sprintf(`
-	SELECT modcount FROM %s
+	SELECT COALESCE(pg_catalog.sum(modcount), 0) AS modcount FROM %s
 `, aosegtablefqn)
 
 	var results []struct {
@@ -93,9 +93,6 @@ func getModCount(connectionPool *dbconn.DBConn, aosegtablefqn string) int64 {
 	err := connectionPool.Select(&results, query)
 	gplog.FatalOnError(err)
 
-	if len(results) == 0 {
-		return 0
-	}
 	return results[0].Modcount
 }
 
--- a/end_to_end/end_to_end_suite_test.go
+++ b/end_to_end/end_to_end_suite_test.go
@@ -276,8 +276,6 @@ var _ = Describe("backup end to end integration tests", func() {
 		if backupConn.Version.Before("6") {
 			testutils.DestroyTestFilespace(backupConn)
 		} else {
-			_ = exec.Command("dropdb", "testdb").Run()
-			_ = exec.Command("dropdb", "restoredb").Run()
 			_ = exec.Command("psql", "postgres", "-c", "DROP RESOURCE QUEUE test_queue").Run()
 			_ = exec.Command("psql", "postgres", "-c", "DROP TABLESPACE test_tablespace").Run()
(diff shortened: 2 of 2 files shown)


jimmyyih:
> Good catch! Can you explain what it means when there are multiple segno entries in a table? I haven't seen that before and there isn't a lot of documentation on these tables.

AO segno entries represent a lot of things.  One thing they represent is write isolation where one user session locks and inserts into segno 1 while another user session locks and inserts into segno 2.  If the first user in this scenario aborts their transaction while the second user commits their inserts into segno 2, gpbackup seems to not detect the incremental change as proposed by this PR.

hidva:
> > Good catch! Can you explain what it means when there are multiple segno entries in a table? I haven't seen that before and there isn't a lot of documentation on these tables.
> 
> AO segno entries represent a lot of things. One thing they represent is write isolation where one user session locks and inserts into segno 1 while another user session locks and inserts into segno 2. If the first user in this scenario aborts their transaction while the second user commits their inserts into segno 2, gpbackup seems to not detect the incremental change as proposed by this PR.

Exactly.

jimmyyih:
Thanks @hidva for the fix.  We'll most likely merge it in after an automated test is written (either by you or by us if you're not comfortable with the our test framework).

Doing some more validation, I found another common case which does not relate to parallel transactions.  If you create a table with CTAS and run VACUUM on it, we run into the same issue.  This is because CTAS (and a couple other batch insert commands) insert into special segno 0.

Example:
```
CREATE TABLE foobar WITH (appendonly=true) AS SELECT i FROM generate_series(1,5)i;
SELECT * FROM gp_toolkit.__gp_aoseg_name('foobar');
VACUUM foobar;
SELECT * FROM gp_toolkit.__gp_aoseg_name('foobar');
INSERT INTO foobar VALUES (1);
SELECT * FROM gp_toolkit.__gp_aoseg_name('foobar');

-- gpbackup incremental will always get modcount 0 due to segno 0 being the first row
```

That should be a good test to use without having to worry about parallel transactions.

hidva:
> Thanks @hidva for the fix. We'll most likely merge it in after an automated test is written (either by you or by us if you're not comfortable with the our test framework).
> 
> Doing some more validation, I found another common case which does not relate to parallel transactions. If you create a table with CTAS and run VACUUM on it, we run into the same issue. This is because CTAS (and a couple other batch insert commands) insert into special segno 0.
> 
> Example:
> 
> ```
> CREATE TABLE foobar WITH (appendonly=true) AS SELECT i FROM generate_series(1,5)i;
> SELECT * FROM gp_toolkit.__gp_aoseg_name('foobar');
> VACUUM foobar;
> SELECT * FROM gp_toolkit.__gp_aoseg_name('foobar');
> INSERT INTO foobar VALUES (1);
> SELECT * FROM gp_toolkit.__gp_aoseg_name('foobar');
> 
> -- gpbackup incremental will always get modcount 0 due to segno 0 being the first row
> ```
> 
> That should be a good test to use without having to worry about parallel transactions.

I want, but there isn't a document about writing an end-to-end test. So can you help me write an automated test for this PR?

jimmyyih:
Sure, we can write the test.

lavjain:
@hidva We have created a PR (https://github.com/hidva/gpbackup/pull/1) for adding the mentioned test

jimmyyih:
Thanks again @hidva for the contribution.  We'll be adding more documentation later to help community contributors.

PR #266, opened 2019-08-01
Backup plugin config

shivzone:
- Added the user provided plugin configuration to backup directory during backups
- Updated /tmp staged config file path to include a timestamp to handle concurrent backups/restores

### DIFF
--- a/backup/backup.go
+++ b/backup/backup.go
@@ -203,7 +203,7 @@ func DoBackup() {
 		if MustGetFlagBool(utils.WITH_STATS) {
 			pluginConfig.MustBackupFile(globalFPInfo.GetStatisticsFilePath())
 		}
-		utils.CopyFile(pluginConfigFlag, globalFPInfo.GetPluginConfigPath())
+		_ = utils.CopyFile(pluginConfigFlag, globalFPInfo.GetPluginConfigPath())
 		pluginConfig.MustBackupFile(globalFPInfo.GetPluginConfigPath())
 	}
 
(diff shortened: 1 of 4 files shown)


tyacovone:
Do these changes account for the case where "plugin_file.yml<timestamp>" doesn't exist in the backup dir?
For example, if I backed up to DataDomain a month ago, then try restoring using the newest version of gpbackup(1.13.0, for example), would the restore fail?

Nit: We should retain the file extension(plugin_config.yml_TIMESTAMP -> plugin_config_TIMESTAMP.yml). `filepath.Base()` and `filepath.Ext()` should help.

shivzone:
> ~Do these changes account for the case where "plugin_file.yml" doesn't exist in the backup dir? For example, if I backed up to DataDomain a month ago, then try restoring using the newest version of gpbackup(1.13.0, for example), would the restore fail?~
> 
> Nit: We should retain the file extension(plugin_config.yml_TIMESTAMP -> plugin_config_TIMESTAMP.yml). `filepath.Base()` and `filepath.Ext()` should help.

backup/backup.go
```diff
@@ -189,12 +191,14 @@ func DoBackup() {
...
 	}
 	metadataFile.Close()
-	if MustGetFlagString(utils.PLUGIN_CONFIG) != "" {
+	if pluginConfigFlag != "" {
 		pluginConfig.MustBackupFile(metadataFilename)
 		pluginConfig.MustBackupFile(globalFPInfo.GetTOCFilePath())
 		if MustGetFlagBool(utils.WITH_STATS) {
 			pluginConfig.MustBackupFile(globalFPInfo.GetStatisticsFilePath())
 		}
+		utils.CopyFile(pluginConfigFlag, globalFPInfo.GetPluginConfigPath())
```
kyeap-pivotal: It is possible for `CopyFile(...)` to return an error. If we are intentionally ignoring this error can we change this line to `_ = utils.CopyFile(...)` so we the fact that something is being returned.

PR #267, opened 2019-08-07
Refactor where random string is appended in plugin_test_bench.sh

jimmyyih:
When running this test against ddboost plugin in concurrent jobs, some
jobs would end up interfering with each other due to the last 3 path
string tokens being the same. This is because the ddboost plugin only
takes the last 3 path string tokens and appends it to the ddboost
remote path configured in the ddboost plugin config.

To fix the concurrent job race condition, move the random string
further up the test directory path where it will be included in the
last 3 path string tokens.

While at it, replace $RANDOM with a /dev/urandom implementation which
has a better chance of generating a non-predictable random number
(assuming cloud entropy is dependable of course).


tyacovone:
The last three path tokens reflect the way in which we take backups, and the `delete` commands for plugins generally make the assumption that this path is predictable given a timestamp. For the s3 plugin for example, we determine the delete path via the following:
```
// note that "backups" is a directory is a fact of how we save, choosing
// to use the 3 parent directories of the source file. That becomes:
// <s3folder>/backups/<date>/<timestamp>
deletePath := filepath.Join(plugin.config.Options["folder"], "backups", date, timestamp)
```

If we choose to change this assumption we must also change our plugins.

PR #268, opened 2019-08-08
Auto generate pipeline name for dev branches

lavjain:
Also, run the fly command so the user does not need to cut/paste


tyacovone:
I like the usage of the branch name in the pipeline name. 

However, I dont think executing the fly set cmd is a very safe pattern to follow, especially for prod pipelines. I consider the copy/pasting a "sanity check," to ensure that the user is aware of what pipeline theyre setting & which secrets files are required.

PR #269, opened 2019-08-09
Add gpbackup manager dependencies to gpbackup dependencies tarball

lavjain:
This will save us time by not downloading the dependencies for the backup manager during build_binaries phase


tyacovone:
It doesnt look like the gpbackup_manager test job is actually using these dependencies. Storing both vendor dirs in the same tarball should work if each is handled properly though

PR #271, opened 2019-08-13
Pass gpbackup_helper the `--on-error-continue` flag

tyacovone:
Context (from commit 7091a90):

Previously, gpbackup_helper did not have the flag --on-error-continue. This means that when gprestore was called with --single-data-file and COPY FROM encountered an error and aborts, gpbackup_helper would error out and exit while gprestore continued onward. This would cause the next gprestore COPY FROM query to possibly hang. This is more likely to happen with large tables with issues.


helper/restore_helper.go
```diff
@@ -45,54 +41,85 @@ func doRestoreAgent() error {
...
 		}
-		log(fmt.Sprintf("Closing pipe for oid %d", oid))
-		err = flushAndCloseRestoreWriter()
+		log(fmt.Sprintf("Data Reader discarded %d bytes", numDiscarded))
+
+		log(fmt.Sprintf("Restoring table with oid %d", oid))
+		bytesRead, err = io.CopyN(writer, reader, int64(end-start))
 		if err != nil {
-			return err
+			// In case COPY FROM or coptN fails in the middle of a load. We
```
yijunkeun: typo copyN

ci/tasks/all-tests.yml
```diff
@@ -50,6 +50,10 @@ run:
 
       make unit
       make integration
+
+      # NOTE: This is a temporary hotfix intended to skip this test when running on CCP cluster because the backup artifact that this test is using only works on local clusters.
+      sed -i 's|It(\`gprestore continues when encountering errors during data load with --single-data-file and --on-error-continue\`, func() {|PIt(\`gprestore continues when encountering errors during data load with --single-data-file and --on-error-continue\`, func() {|g' end_to_end/end_to_end_suite_test.go
+
```
lavjain: Instead of modifying multiple scripts, we could modify the test case to check if the master hostname is `mdw`

restore/data.go
```diff
@@ -149,25 +158,7 @@ func restoreDataFromTimestamp(fpInfo backup_filepath.FilePathInfo, dataEntries [
...
-		}
-	}
-
-	if fatalErr != nil {
-		fmt.Println("")
-		gplog.Fatal(fatalErr, "")
-	} else if numErrors > 0 {
+	if numErrors > 0 {
 		fmt.Println("")
 		gplog.Error("Encountered %d errors during table data restore; see log file %s for a list of table errors.", numErrors, gplog.GetLogFilePath())
```
shivzone: Minor nitpick and this is not really in the scope of this PR ... When on-error-continue is not set we will atmost have one error. Can we reword this to "Encountered %d error(s)"

helper/restore_helper.go
```diff
@@ -45,54 +41,85 @@ func doRestoreAgent() error {
...
+		log(fmt.Sprintf("Removing pipe for oid %d: %s", oid, currentPipe))
+		errRemove = removeFileIfExists(currentPipe)
+		if errRemove != nil {
+			_ = removeFileIfExists(nextPipe)
+			return errRemove
+		}
+
+		if err != nil {
+			if *onErrorContinue {
+				logError(fmt.Sprintf("Error encountered: %v", err))
```
shivzone: Might be useful to log the corresponding oid in the error encountered message

ci/tasks/all-tests.yml
kyeap-pivotal: I would advocate for keeping the sed replacements as they currently are. IMO the issue is more visible this way even if the code base is slightly messier. This is a temporary fix so it will be removed or the test will be disabled in a cleaner manner in the future.

helper/restore_helper.go
kyeap-pivotal: We are already logging the oid leading up to a potential error when opening the pipe, restoring data, closing the pipe, and deleteing the pipe.

PR #273, opened 2019-08-22
Enhance error logging in plugin test bench

tyacovone:
Previously, a number of calls to gpbackup or gprestore were piping their
error outputs to /dev/null. This made debugging more difficult.

We now maintain the same logging pattern, while also outputting the error
log when we hit an unexpected failure.

Authored-by: Trevor Yacovone <tyacovone@pivotal.io>

--

_NOTE:_ Remaining references to /dev/null are for calls we expect to fail.

### DIFF
--- a/plugins/plugin_test_bench.sh
+++ b/plugins/plugin_test_bench.sh
@@ -35,9 +35,12 @@ testfile="$testdir/testfile_$time_second.txt"
 testdata="$testdir/testdata_$time_second.txt"
 test_no_data="$testdir/test_no_data_$time_second.txt"
 
+logdir="/tmp/test_bench_logs"
+
 text="this is some text"
 data=`LC_ALL=C tr -dc 'A-Za-z0-9' </dev/urandom | head -c 1000 ; echo`
 mkdir -p $testdir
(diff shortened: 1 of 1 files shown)


plugins/plugin_test_bench.sh
```diff
@@ -306,7 +313,10 @@ set -e
 test_backup_and_restore_with_plugin() {
     flags=$1
     test_db=plugin_test_db
-    log_file=/tmp/plugin_test_log_file
+    log_file_backupTMP="$logdir/$RANDOM"
```
jimmyyih: What's up with naming all these log files with $RANDOM?  Why not just hardcode them to an actual distinct filename?

shivzone: +1 on above suggestion

tyacovone: This function is called twice, and I intend for each err output to be isolated and therefore assigned to a unique file. This could be done with distinct files or the same hardcoded filenames and `rm` call at the end of the function.

Good catch on the log_file – I had mistakenly assumed that it was being used/referenced elsewhere.

PR #281, opened 2019-09-05
Properly close TOC, history, report, and config files

shivzone:
Explicitly fsync and call close on TOC file
```
    It is possible that the write() succeeds but the underlying buffer flush
    to disk fails because we chmod 0444 the TOC file immediately after the
    write(). To prevent TOC data loss, explicitly fsync() and close the
    file before making the file read-only.
```

Close history, report, and config files after write
```
    Files should always be closed anyways so explicitly call them instead of
    leaving it up to the OS when the program exits.

    Disclaimer: A chmod 0444 happens at the end for these specific files. A
    file close does not guarantee that the OS buffers have flushed to disk
    before the file permissions change (based on the underlying filesystem
    and its buffer implementation). We should change this later to
    explicitly fsync() but that requires refactoring everything to not use
    io.Writer.
```

### DIFF
--- a/backup_history/history.go
+++ b/backup_history/history.go
@@ -60,6 +60,8 @@ func WriteConfigFile(config *BackupConfig, configFilename string) {
 	configContents, _ := yaml.Marshal(config)
 	_, err := configFile.Write(configContents)
 	gplog.FatalOnError(err)
+	err = configFile.Close()
+	gplog.FatalOnError(err)
 	err = operating.System.Chmod(configFilename, 0444)
 	gplog.FatalOnError(err)
 }
@@ -156,13 +158,11 @@ func (history *History) WriteToFileAndMakeReadOnly(filename string) error {
 		return err
 	}
 	historyFile := iohelper.MustOpenFileForWriting(filename)
-	defer func() {
-		err = historyFile.Close()
-		if err != nil {
-			gplog.Warn("cannot close history file: %v", err)
-		}
-	}()
 	_, err = historyFile.Write(historyFileContents)
+	if err != nil {
+		return err
+	}
+	err = historyFile.Close()
 	if err != nil {
 		return err
 	}
@@ -170,7 +170,6 @@ func (history *History) WriteToFileAndMakeReadOnly(filename string) error {
 	if err != nil {
 		return err
 	}
-
 	return nil
 }
 
--- a/utils/toc.go
+++ b/utils/toc.go
@@ -4,9 +4,9 @@ import (
 	"fmt"
 	"io"
 	"regexp"
+	"os"
 
 	"github.com/greenplum-db/gp-common-go-libs/gplog"
-	"github.com/greenplum-db/gp-common-go-libs/iohelper"
 	"github.com/greenplum-db/gp-common-go-libs/operating"
 	"gopkg.in/yaml.v2"
 )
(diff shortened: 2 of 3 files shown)


utils/toc.go
```diff
@@ -76,29 +76,53 @@ func NewSegmentTOC(filename string) *SegmentTOC {
...
-	tocFile := iohelper.MustOpenFileForWriting(filename)
 	tocContents, err := yaml.Marshal(toc)
 	gplog.FatalOnError(err)
-	MustPrintBytes(tocFile, tocContents)
-	err = operating.System.Chmod(filename, 0444)
+
+	tocFile, err := os.OpenFile(filename, os.O_CREATE | os.O_EXCL | os.O_WRONLY, 0644)
+	gplog.FatalOnError(err)
+
+	_, err = tocFile.Write(tocContents)
```
lavjain: The number of bytes written to the toc file should be tracked and logged.

utils/toc.go
```diff
@@ -76,29 +76,53 @@ func NewSegmentTOC(filename string) *SegmentTOC {
...
+	if err != nil {
+		return err
+	}
+
+	err = tocFile.Close()
+	if err != nil {
+		return err
+	}
+
+	err = os.Chmod(filename, 0444)
```
lavjain: The code segment above is repeated. Would be good to have a utility function that can be called to close/sync files.

utils/toc.go
```diff
@@ -4,9 +4,9 @@ import (
 	"fmt"
 	"io"
 	"regexp"
+	"os"
 
 	"github.com/greenplum-db/gp-common-go-libs/gplog"
-	"github.com/greenplum-db/gp-common-go-libs/iohelper"
```
lavjain: +1 on getting rid of iohelper

utils/toc.go
jimmyyih: Sure, we can log this out.  Something like "Wrote N bytes to <TOC File>" would be helpful.

utils/toc.go
jimmyyih: They look similar but handle errors differently... and the error handlings are very interleaved with the file system calls.

However, we can do this as part of a separate history, report, and config file refactor that should happen.  They all follow similar patterns so having a similar function that takes in a byte buffer (contents to write) and a boolean flag (fatal or return error) could be nice.

shivzone: The reason we kept it separate is because for other backup files we aren't in a position to sync and close as they use io.WriteCloser which doesn't expose a sync functionality. As Jimmy said, when we refactor all our backup files to use os.file directly we can have a util function to generalize the sync+close functionality

utils/toc.go
shivzone: updated

backup_history/history.go
```diff
@@ -156,21 +158,18 @@ func (history *History) WriteToFileAndMakeReadOnly(filename string) error {
...
-			gplog.Warn("cannot close history file: %v", err)
-		}
-	}()
 	_, err = historyFile.Write(historyFileContents)
 	if err != nil {
 		return err
 	}
+	err = historyFile.Close()
+	if err != nil {
+		return err
```
tyacovone: Do we expect this error to contain reference to the history file? If not, lets append this info to the error message

shivzone: Not entirely sure. But we didn't want to lose the original error message so went with returning err itself. For the other files as well we either return the err or `gplog.FatalOnError(err)`
We can possibly address this once we refactor sync/close consistently across all backup files

PR #282, opened 2019-09-06
Do not send sigterm to gpbackup_helpers on success

kyeap-pivotal:
Only send sigterm to the backup helpers if gpbackup was terminated or if
gpbackup fatals and must exit immediately. On a success case let the
helpers finish and close by themselves.

### DIFF
--- a/backup/backup.go
+++ b/backup/backup.go
@@ -389,8 +389,9 @@ func backupStatistics(tables []Table) {
 }
 
 func DoTeardown() {
+	backupFailed := false
 	defer func() {
-		DoCleanup()
+		DoCleanup(backupFailed)
 
 		errorCode := gplog.GetErrorCode()
 		if errorCode == 0 {
@@ -408,6 +409,7 @@ func DoTeardown() {
 		} else {
 			errStr = fmt.Sprintf("%v", err)
 		}
+		backupFailed = true
 	}
 	if wasTerminated {
 		/*
@@ -417,6 +419,7 @@ func DoTeardown() {
 		 * doesn't exit while cleanup is still in progress.
 		 */
 		CleanupGroup.Wait()
+		backupFailed = true
 		return
 	}
 	if errStr != "" {
@@ -464,7 +467,7 @@ func DoTeardown() {
 	}
 }
 
-func DoCleanup() {
+func DoCleanup(backupFailed bool) {
 	defer func() {
 		if err := recover(); err != nil {
 			gplog.Warn("Encountered error during cleanup: %v", err)
@@ -476,7 +479,10 @@ func DoCleanup() {
 	gplog.Verbose("Beginning cleanup")
 	if globalFPInfo.Timestamp != "" {
 		if MustGetFlagBool(utils.SINGLE_DATA_FILE) {
-			utils.CleanUpSegmentHelperProcesses(globalCluster, globalFPInfo, "backup")
+			if backupFailed {
+				// Cleanup only if terminated or fataled
+				utils.CleanUpSegmentHelperProcesses(globalCluster, globalFPInfo, "backup")
+			}
 			if wasTerminated {
 				// It is possible for the COPY command to become orphaned if an agent process is killed
 				utils.TerminateHangingCopySessions(connectionPool, globalFPInfo, "gpbackup")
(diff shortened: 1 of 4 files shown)


helper/helper.go
```diff
@@ -168,7 +168,7 @@ func removeFileIfExists(filename string) error {
  * Shared helper functions
  */
 
-func DoCleanup() {
+func DoCleanup(shouldCleanupHelpers bool) {
```
jimmyyih: Why does the helper need this flag?  Isn't this isolated from the backup and restore `DoCleanup()` implementations?

backup/backup.go
```diff
@@ -461,7 +463,7 @@ func DoTeardown() {
 	}
 }
 
-func DoCleanup() {
+func DoCleanup(shouldCleanupHelpers bool) {
```
tyacovone: Why is an argument needed here, or in gprestore's `DoCleanup`, if the shouldCleanHelpers variable is accessible as a global?

backup/backup.go
```diff
@@ -414,6 +415,7 @@ func DoTeardown() {
 		 * doesn't exit while cleanup is still in progress.
 		 */
 		CleanupGroup.Wait()
+		shouldCleanupHelpers = true
```
lavjain: We can call `utils.CleanUpSegmentHelperProcesses(globalCluster, globalFPInfo, "backup")` here directly and move the logic out of `DoCleanup()`

restore/restore.go
```diff
@@ -297,6 +297,7 @@ func DoTeardown() {
 		} else {
 			errStr = fmt.Sprintf("%v", err)
 		}
+		shouldCleanupHelpers = true
```
lavjain: We can call utils.CleanUpSegmentHelperProcesses(globalCluster, globalFPInfo, "restore") here directly and move the logic out of DoCleanup()

backup/backup.go
shivzone: I agree with @lavjain. Introducing another global variable seems unnecessary given the rather restricted use of it

backup/backup.go
kyeap-pivotal: This along with @shivzone's comment that a global is too heavy is true. This doesn't need to be a global. It can be a local variable.

backup/backup.go
kyeap-pivotal: This is a valid consideration. It was a choice by me to try to keep the helper cleanup function calls consolidated in `DoCleanup`. Moving the `CleanUpSegmentHelperProcesses` out also results in having to check the `utils.SINGLE_DATA_FILE` flag so we would be replacing the two `shouldCleanupHelpers = true` calls with at least
```
if MustGetFlagBool(utils.SINGLE_DATA_FILE) {
	utils.CleanUpSegmentHelperProcesses(globalCluster, globalFPInfo, "backup")
}
```
and if we wanted to gpbackup's termination behavior exactly the same (which we don't have to)
```
if globalFPInfo.Timestamp != "" {
	if MustGetFlagBool(utils.SINGLE_DATA_FILE) {
		utils.CleanUpSegmentHelperProcesses(globalCluster, globalFPInfo, "backup")
	}
}
```

backup/backup.go
```diff
@@ -386,8 +386,9 @@ func backupStatistics(tables []Table) {
 }
 
 func DoTeardown() {
+	backupFailed := false
 	defer func() {
-		DoCleanup()
+		DoCleanup(shouldCleanupHelpers)
```
tyacovone: This seems be be an outdated var reference -- did you mean `backupFailed` ? (also applies to `restoreFailed`).

restore/global_variables.go
```diff
@@ -21,15 +21,16 @@ import (
...
+	backupConfig         *backup_history.BackupConfig
+	connectionPool       *dbconn.DBConn
+	globalCluster        *cluster.Cluster
+	globalFPInfo         backup_filepath.FilePathInfo
+	globalTOC            *utils.TOC
+	pluginConfig         *utils.PluginConfig
+	restoreStartTime     string
+	version              string
+	wasTerminated        bool
+	shouldCleanupHelpers bool
```
tyacovone: This seems outdated, we should remove this global var since we've since switched to using `backupFailed`/`restoreFailed` local vars

backup/backup.go
kyeap-pivotal: gah good catch. I distinctly remember catching this also. forgot to force push.

PR #283, opened 2019-09-10
Fixed handling of dropped grantors during backup

shivzone:
When we create a role and add it to a member and a grantor and
subsequently drop the grantor, backup incorrectly logs the grantor as
'unknown (OID=xxx)'. This happens since pg_get_userbyid returns the
unknown string instead of NULL or empty string.

We could alternatively updated the existing query as follows
```
SELECT
quote_ident(pg_get_userbyid(pga.roleid)) AS a,
quote_ident(pg_get_userbyid(pga.member)) AS b,
case when pg_get_userbyid(pga.grantor) like 'unknown (OID='||pga.grantor||')' then '' else quote_ident(pg_get_userbyid(pga.grantor)) end AS c,
admin_option as isadmin
FROM pg_auth_members pga
ORDER BY roleid, member;
```
The issue with this is that we explictly look for a grantor string. We could actually have a grantor with the name 'unknown (OID='||{GRANTOR_OID}||') which would be returned incorrectly as an empty string.

### DIFF
--- a/backup/queries_globals.go
+++ b/backup/queries_globals.go
@@ -491,11 +491,11 @@ func (rm RoleMember) GetMetadataEntry() (string, utils.MetadataEntry) {
 func GetRoleMembers(connectionPool *dbconn.DBConn) []RoleMember {
 	query := `
 SELECT
-	quote_ident(pg_get_userbyid(roleid)) AS role,
-	quote_ident(pg_get_userbyid(member)) AS member,
-	quote_ident(pg_get_userbyid(grantor)) AS grantor,
-	admin_option as isadmin
-FROM pg_auth_members
+	quote_ident(pg_get_userbyid(pga.roleid)) AS role,
+	quote_ident(pg_get_userbyid(pga.member)) AS member,
+	CASE WHEN pg_get_userbyid(pga.grantor) like 'unknown (OID='||pga.grantor||')' THEN '' ELSE quote_ident(pg_get_userbyid(pga.grantor)) END AS grantor,
+	admin_option AS isadmin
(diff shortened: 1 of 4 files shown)


jimmyyih:
I'd rather have the alternative query mentioned in the opening PR message.  It's super extremely unlikely that there would be an existing role with that name (and if they did, that's their problem... not ours because we're handling the error of `pg_get_userbyid()` by changing it to `''`).

backup/queries_globals.go
```diff
@@ -491,12 +491,13 @@ func (rm RoleMember) GetMetadataEntry() (string, utils.MetadataEntry) {
...
-	admin_option as isadmin
-FROM pg_auth_members
-ORDER BY roleid, member;`
+ur.rolname AS role,
+um.rolname AS member,
+CASE WHEN ug.rolname != NULL THEN ug.rolname ELSE '' END AS grantor,
+a.admin_option AS isadmin
+FROM pg_auth_members a LEFT JOIN pg_authid ur on ur.oid = a.roleid
+LEFT JOIN pg_authid um on um.oid = a.member
+LEFT JOIN pg_authid ug on ug.oid = a.grantor ORDER BY 1,2`
```
lavjain: Do not recommend! The left joins are very expensive.

kyeap-pivotal:
The current change is based on the query from pg_dumpall.c
```

lavjain:
IMO we should go with the query in the PR comment

kyeap-pivotal:
Changed the query back to the original (faster one). merging.

PR #289, opened 2019-09-23
Organize imports (no logic change)

lavjain:


shivzone:
Since you are doing a code organization change can you also update toc.go where there is some formatting issues. run `make format` and add it as part of this commit

PR #290, opened 2019-09-24
Set enable_mergejoin off for GPDB4

kyeap-pivotal:
Enabling merge join on a GPDB4 system while taking a backup was causing
ACL backups to fail. Turn off the GUC in GPDB4 before taking a backup.


backup/wrappers.go
```diff
@@ -65,6 +65,9 @@ func SetSessionGUCs(connNum int) {
 	// getting the max value of the GUC.
 	connectionPool.MustExec("SELECT set_config('extra_float_digits', (SELECT max_val FROM pg_settings WHERE name = 'extra_float_digits'), false)")
 
+	if connectionPool.Version.Is("4") {
+		connectionPool.MustExec("SET enable_mergejoin TO off", connNum)
```
jimmyyih: IMO, just have it set to `off` for all versions.  Greenplum specifically has it turned `off` by default so we should do the same.  We'll add a version check later when it's turned `on` by default (possibly in GPDB 7?).

PR #291, opened 2019-09-25
Batch lock queries

kyeap-pivotal:
Previously, gpbackup was dispatching 1 LOCK TABLE query for each table. This
can cause strain on the network as the master would have to dispatch N
amount of LOCK TABLE queries. This change batches table LOCKs so that there
are at most 100 LOCK dispatches.

### DIFF
--- a/backup/queries_relations.go
+++ b/backup/queries_relations.go
@@ -330,11 +330,50 @@ AND %s;`, optionsStr, relationAndSchemaFilterClause(), ExtensionFilterClause("c"
 
 func LockTables(connectionPool *dbconn.DBConn, tables []Relation) {
 	gplog.Info("Acquiring ACCESS SHARE locks on tables")
+
 	progressBar := utils.NewProgressBar(len(tables), "Locks acquired: ", utils.PB_VERBOSE)
 	progressBar.Start()
-	for _, table := range tables {
-		connectionPool.MustExec(fmt.Sprintf("LOCK TABLE %s IN ACCESS SHARE MODE", table.FQN()))
-		progressBar.Increment()
+
+	const batchSize = 100
+	lastBatchSize := len(tables) % batchSize
+	tableBatches := generateTableBatches(tables, batchSize)
+	currentBatchSize := batchSize
+	for i, currentBatch := range tableBatches {
+		connectionPool.MustExec(fmt.Sprintf("LOCK TABLE %s IN ACCESS SHARE MODE", currentBatch))
+
+		if i == len(tableBatches)-1 && lastBatchSize > 0 {
+			currentBatchSize = lastBatchSize
+		}
+
+		progressBar.Add(currentBatchSize)
 	}
+
 	progressBar.Finish()
+}
+
+// generateTableBatches batches tables to reduce network congestion and
+// resource contention.  Returns an array of batches where a batch of tables is
+// a single string with comma separated tables
+func generateTableBatches(tables []Relation, batchSize int) []string {
+	var tableNames []string
+	for _, table := range tables {
+		tableNames = append(tableNames, table.FQN())
+	}
+
+	var end int
+	var batches []string
+	i := 0
+	for i < len(tables) {
+
+		if i+batchSize < len(tables) {
+			end = i + batchSize
+		} else {
+			end = len(tables)
+		}
+
+		batches = append(batches, strings.Join(tableNames[i:end], ", "))
+		i = end
+	}
+
+	return batches
 }
--- a/backup/queries_relation_internal_test.go
+++ b/backup/queries_relation_internal_test.go
@@ -1,0 +1,33 @@
+package backup
+
+import (
+	"fmt"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("backup internal tests", func() {
+	Describe("generateLockQueries", func() {
+		It("batches tables together and generates lock queries", func() {
+			tables := []Relation{}
+			for i := 0; i < 200; i++ {
+				tables = append(tables, Relation{0, 0, "public", fmt.Sprintf("foo%d", i)})
+			}
+
+			batchSize := 100
+			lockQueries := generateTableBatches(tables, batchSize)
+			Expect(len(lockQueries)).To(Equal(2))
+		})
+		It("batches up remaining leftover tables together in a single lock query", func() {
+			tables := []Relation{}
+			for i := 0; i < 101; i++ {
+				tables = append(tables, Relation{0, 0, "public", fmt.Sprintf("foo%d", i)})
+			}
+
+			batchSize := 50
+			lockQueries := generateTableBatches(tables, batchSize)
+			Expect(len(lockQueries)).To(Equal(3))
+		})
+	})
+})
--- a/utils/progress_bar.go
+++ b/utils/progress_bar.go
@@ -52,6 +52,7 @@ type ProgressBar interface {
 	Start() *pb.ProgressBar
 	Finish()
 	Increment() int
+	Add(int) int
 }
 
 type VerboseProgressBar struct {


backup/queries_relations.go
```diff
@@ -331,11 +331,38 @@ AND %s;`, optionsStr, relationAndSchemaFilterClause(), ExtensionFilterClause("c"
...
 		progressBar.Increment()
 	}
 	progressBar.Finish()
 }
+
+// generateLockQueries batches tables and generates
+// lock queries to reduce network congestion and
+// resource contention.
+func generateLockQueries(tables []Relation) []string {
+	var queriesToRun = 100
```
shivzone: Set this a `const`

backup/queries_relations.go
```diff
@@ -331,11 +331,38 @@ AND %s;`, optionsStr, relationAndSchemaFilterClause(), ExtensionFilterClause("c"
 
 func LockTables(connectionPool *dbconn.DBConn, tables []Relation) {
 	gplog.Info("Acquiring ACCESS SHARE locks on tables")
-	progressBar := utils.NewProgressBar(len(tables), "Locks acquired: ", utils.PB_VERBOSE)
+	queries := generateLockQueries(tables)
+	progressBar := utils.NewProgressBar(len(queries), "Locks acquired: ", utils.PB_VERBOSE)
```
jimmyyih: This is very deceiving since it doesn't report the actual number of locks required.  For example, taking a backup of 150 tables will always show 100 locks acquired.

backup/queries_relations.go
```diff
@@ -331,11 +331,38 @@ AND %s;`, optionsStr, relationAndSchemaFilterClause(), ExtensionFilterClause("c"
...
+		connectionPool.MustExec(query)
 		progressBar.Increment()
 	}
 	progressBar.Finish()
 }
+
+// generateLockQueries batches tables and generates
+// lock queries to reduce network congestion and
+// resource contention.
+func generateLockQueries(tables []Relation) []string {
```
jimmyyih: Add a comment stating that this function is round-robin batching tables.  I was under the impression at first that we were batching up to 100 tables into one LOCK query (better solution IMO but this works).  I was surprised when going through dlv debugger to see even distribution of tables to be locked.  Note that this round-robin solution possibly breaks when backing up ~3408704200 tables (random math involving max query length (INT_MAX) and max relation name length (64))... but having that many tables would probably cause other issues first.  Anyways, just wanted to point out that query length could get really big and that may not be good for us (I'd argue that bigger is better though).

lavjain:
Instead of defining the max number of lock queries, we should be defining a batch size (default 100).

backup/queries_relations.go
roicos: fixed

backup/queries_relations.go
```diff
@@ -331,11 +331,54 @@ AND %s;`, optionsStr, relationAndSchemaFilterClause(), ExtensionFilterClause("c"
...
+
+		for i := 0; i < currentBatchSize; i++ {
+			progressBar.Increment()
+		}
 	}
+
 	progressBar.Finish()
 }
+
+// generateBatchLockQueries batches tables and generates lock queries to reduce
```
shivzone: minor knitpick comment for function using 
```
/* 
*/
 block

File: backup/queries_relations.go [hunk 5]
```diff
@@ -331,11 +331,54 @@ AND %s;`, optionsStr, relationAndSchemaFilterClause(), ExtensionFilterClause("c"
...
+func generateBatchLockQueries(tables []Relation, batchSize int) []string {
+	var queriesToRun = len(tables) / batchSize
+	var tableNames []string
+	var queries []string
+
+	for _, table := range tables {
+		tableNames = append(tableNames, table.FQN())
+	}
+
+	for i := 0; i < queriesToRun; i++ {
```
shivzone: you could simplify the redundancy in this block and the one below by doing the following:
```
whilte(i < numTables {
    // Set end to min(i+batchSize, numTables)
    // batch := tableNames[i : end]
    // i = end
}
```

backup/queries_relations.go
jimmyyih: I think our gpbackup code base is using majority `//` so we should stick with that.

shivzone: In our codebase for functions i primarily see /* */ used for multiline comments. That said, the standard go convention used for go and major projects looks like // for multi line as well. Can ignore this :)

backup/queries_relations.go
```diff
@@ -331,11 +331,54 @@ AND %s;`, optionsStr, relationAndSchemaFilterClause(), ExtensionFilterClause("c"
 
 func LockTables(connectionPool *dbconn.DBConn, tables []Relation) {
 	gplog.Info("Acquiring ACCESS SHARE locks on tables")
+
+	const batchSize = 100
+	queries := generateBatchLockQueries(tables, batchSize)
```
lavjain: One idea here would be to generate a string array of comma separated tableNames, so the original logic can stay very similar.

backup/queries_relations.go
roicos: Done.

backup/queries_relations.go
roicos: done.

backup/queries_relations.go
```diff
@@ -331,11 +332,42 @@ AND %s;`, optionsStr, relationAndSchemaFilterClause(), ExtensionFilterClause("c"
...
 	progressBar.Start()
-	for _, table := range tables {
-		connectionPool.MustExec(fmt.Sprintf("LOCK TABLE %s IN ACCESS SHARE MODE", table.FQN()))
-		progressBar.Increment()
+
+	currentBatchSize := batchSize
+	for i, currentBatch := range tableBatches {
+		connectionPool.MustExec(fmt.Sprintf("LOCK TABLE %s IN ACCESS SHARE MODE", currentBatch))
+
+		if i == len(tableBatches)-1 && len(tables)%batchSize > 0 {
```
jimmyyih: We're using `len(tables) % batchSize` quite often in this loop when its computed value is not dynamic.  Let's initialize a variable for it outside the loop so we don't have to always compute it.

backup/queries_relations.go
```diff
@@ -331,11 +332,42 @@ AND %s;`, optionsStr, relationAndSchemaFilterClause(), ExtensionFilterClause("c"
...
+			currentBatchSize = len(tables) % batchSize
```
lavjain: currentBatchSize is not changing here (will always be the same value)

backup/queries_relations.go
```diff
@@ -331,11 +332,42 @@ AND %s;`, optionsStr, relationAndSchemaFilterClause(), ExtensionFilterClause("c"
...
+	const batchSize = 100
+	tableBatches := generateTableBatches(tables, batchSize)
 	progressBar := utils.NewProgressBar(len(tables), "Locks acquired: ", utils.PB_VERBOSE)
 	progressBar.Start()
-	for _, table := range tables {
-		connectionPool.MustExec(fmt.Sprintf("LOCK TABLE %s IN ACCESS SHARE MODE", table.FQN()))
-		progressBar.Increment()
+
+	currentBatchSize := batchSize
+	for i, currentBatch := range tableBatches {
```
lavjain: If we return the number of tables in the batch along with the string (i.e. returning two arrays), we don't need to worry about currentBatchSize

Alternatively, generateTableBatches could return the slices of tables based on the original array and length could be computed on the fly with String.join being done inside the loop here

backup/queries_relations.go
kyeap-pivotal: @jimmyyih @lavjain @roicos 
How does something like this look?
```
for i, currentBatch := range tableBatches[:len(tableBatches)-1] {
	connectionPool.MustExec(fmt.Sprintf("LOCK TABLE %s IN ACCESS SHARE MODE", currentBatch))
	progressBar.Add(batchSize)
}

// Handle the last batch manually because the number of tables in it can be
// less than const batchSize
lastBatchSize := len(tables) % batchSize
if lastBatchSize == 0 {
	lastBatchSize = batchSize
}
connectionPool.MustExec(fmt.Sprintf("LOCK TABLE %s IN ACCESS SHARE MODE", tableBatches[len(tableBatches)]))
progressBar.Add(lastBatchSize)

progressBar.Finish()
}

```

jimmyyih: That would assume that len(tables) is never 0 which I guess is always true if we have reached this function.  You can remove the conditional if you do it like this:
```
lastBatchSize := len(tables) % (batchSize + 1) // Always get a number between 1 and 100
```

lavjain:
Here is the pseudo code (may not be accurate) that I came up with:
```
func LockTables(connectionPool *dbconn.DBConn, tables []Relation) {
        gplog.Info("Acquiring ACCESS SHARE locks on tables")
        const batchSize = 100
        progressBar := utils.NewProgressBar(len(tables), "Locks acquired: ", utils.PB_VERBOSE) 
        progressBar.Start()
        for size, currentBatch := range generateTableBatches(tables, batchSize) {  
                connectionPool.MustExec(fmt.Sprintf("LOCK TABLE %s IN ACCESS SHARE MODE", strings.Join(currentBatch, ",")))
                progressBar.Add(size)
        }
        progressBar.Finish()
}       

func generateTableBatches(tables []Relation, batchSize int) []int, [][]string {
        var tableNames []string
        for _, table := range tables {
                tableNames = append(tableNames, table.FQN())
        }
        var sizes []int = new []int{}
        var batches [][]string = new [][]string{}
        for i := 0; i < len(tables); i += batchSize {
                if i + batchSize > len(tables) {
                     batchSize = len(tables) - i
                }
                sizes = append(sizes, batchSize)
                batches = append(batches, tableNames[i:i+batchSize])
        }
        return sizes, batches
}
```

backup/queries_relations.go
kyeap-pivotal: > ```
> lastBatchSize := len(tables) % (batchSize + 1) // Always get a number between 1 and 100
> ```
This seems to be only correct if len(tables) is less than batchSize. And off by -1 for numbers >= batchSize.

The last batch size is correct with `len(tables) % batchsize`. Except if the last batch size is also full, in which case `len(tables) % batchsize` results in 0. Having a lastBatchSize of 0 is impossible. 

Discovered a bug with the code I posted above when there are 0 tables to backup, best bet is probably to go with this.
```
	// Calculate size of the last batch
	lastBatchSize := len(tables) % batchSize
	if lastBatchSize == 0 {
		lastBatchSize = batchSize
	}

	currentBatchSize := batchSize
	for i, currentBatch := range tableBatches {
		connectionPool.MustExec(fmt.Sprintf("LOCK TABLE %s IN ACCESS SHARE MODE", currentBatch))

		// Handle the last batch progress bar manually because the number of
		// tables in it can be less than const batchSize
		if i == len(tableBatches)-1 {
			currentBatchSize = lastBatchSize
		}
		progressBar.Add(currentBatchSize)
	}
```

kyeap-pivotal:
> Here is the pseudo code (may not be accurate) that I came up with:
> 
> ```
> func LockTables(connectionPool *dbconn.DBConn, tables []Relation) {
>         gplog.Info("Acquiring ACCESS SHARE locks on tables")
>         const batchSize = 100
>         progressBar := utils.NewProgressBar(len(tables), "Locks acquired: ", utils.PB_VERBOSE) 
>         progressBar.Start()
>         ...
> ```

This is probably less performant in all scenarios, we are doubling the amount of appends and slice resizing, and memory used. `batchSize` always the same except the the last batch, no need to store the same constant so many times in the `sizes` slice.

kyeap-pivotal:
Layered one last commit based on Jimmy's Feedback and Lav's pseudo code.

backup/queries_relations.go
```diff
@@ -331,11 +332,49 @@ AND %s;`, optionsStr, relationAndSchemaFilterClause(), ExtensionFilterClause("c"
...
+func generateTableBatches(tables []Relation, batchSize int) ([]string, int) {
+	var tableNames []string
+	for _, table := range tables {
+		tableNames = append(tableNames, table.FQN())
+	}
+
+	i := 0
+	var batches []string
+	for i < len(tables) {
+		end := int(math.Min(float64(i+batchSize), float64(len(tables))))
```
jimmyyih: Why are we casting stuff to float64?... Oh okay I looked it up... It seems math.Min() only takes float64 vars.  That's... an interesting choice by Golang.  Could we just define our own simple min function that takes integers?  Casting an int to float and then back into an int is not good IMO.

backup/queries_relations.go
```diff
@@ -331,11 +332,49 @@ AND %s;`, optionsStr, relationAndSchemaFilterClause(), ExtensionFilterClause("c"
...
+			currentBatchSize = lastBatchSize
+		}
+		progressBar.Add(currentBatchSize)
 	}
+
 	progressBar.Finish()
 }
+
+// generateTableBatches batches tables to reduce network congestion and
+// resource contention.  Returns an array of batches where a batch of tables is
```
jimmyyih: Add to this function comment definition that this function is also returning the last batch's size for \<insert reason>.

backup/queries_relations.go
roicos: Manual comparison done in for loop.

backup/queries_relations.go
roicos: Actually, The last BatchSize in Jimmy's example is probably only calculated once despite being in the loop because in the conditional 
`if i == len(tableBatches)-1 && len(tables)%batchSize > 0 {`
The first clause `i == len(tableBatches)-1` will fail until the last loop iteration.

PR #292, opened 2019-09-27
Optimize composites

roicos:
This makes the following changes to getCompositeTypeAttributes queries:

- filters out implicitly created attributes
- removes ordering as we put data in hashmap after that anyway

This allows to increase query performance from 25 to 90%

### DIFF
--- a/backup/queries_types.go
+++ b/backup/queries_types.go
@@ -258,16 +258,18 @@ type Attribute struct {
 }
 
 func getCompositeTypeAttributes(connectionPool *dbconn.DBConn) map[uint32][]Attribute {
+	gplog.Verbose("Getting composite type attributes")
+
 	before6query := `SELECT
 	t.oid AS compositetypeoid,
 	quote_ident(a.attname) AS name,
 	pg_catalog.format_type(a.atttypid, a.atttypmod) AS type,
 	coalesce(quote_literal(d.description),'') AS comment
 	FROM pg_type t
+	JOIN pg_class c ON t.typrelid = c.oid
 	JOIN pg_attribute a ON t.typrelid = a.attrelid
 	LEFT JOIN pg_description d ON (d.objoid = a.attrelid AND d.classoid = 'pg_class'::regclass AND d.objsubid = a.attnum)
-	WHERE t.typtype = 'c'
-	ORDER BY t.oid, a.attnum;`
+	WHERE t.typtype = 'c' AND c.relkind = 'c';`
 
 	masterQuery := `SELECT
 	t.oid AS compositetypeoid,
--- a/backup/queries_types.go
+++ b/backup/queries_types.go
@@ -283,12 +283,12 @@ func getCompositeTypeAttributes(connectionPool *dbconn.DBConn) map[uint32][]Attr
 	coalesce(quote_literal(d.description),'') AS comment
 	FROM pg_type t
 	JOIN pg_attribute a ON t.typrelid = a.attrelid
+	JOIN pg_class c ON t.typrelid = c.oid
+	LEFT JOIN pg_description d ON (d.objoid = a.attrelid AND d.classoid = 'pg_class'::regclass AND d.objsubid = a.attnum)
 	LEFT JOIN pg_type at ON at.oid = a.atttypid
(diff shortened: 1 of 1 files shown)


pivotal-issuemaster:
@roicos Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/greenplum?repositoryId=greenplum-db/gpbackup&pullRequestId=292)!

[Click here](https://cla.pivotal.io/sync/greenplum?repositoryId=greenplum-db/gpbackup&pullRequestId=292) to manually synchronize the status of this Pull Request.

See the [FAQ](https://cla.pivotal.io/about) for frequently asked questions.

pivotal-issuemaster:
@roicos Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/greenplum?repositoryId=greenplum-db/gpbackup&pullRequestId=292)!

backup/queries_types.go
```diff
@@ -281,12 +283,12 @@ func getCompositeTypeAttributes(connectionPool *dbconn.DBConn) map[uint32][]Attr
 	coalesce(quote_literal(d.description),'') AS comment
 	FROM pg_type t
 	JOIN pg_attribute a ON t.typrelid = a.attrelid
+	JOIN pg_class c ON t.typrelid = c.oid
+	LEFT JOIN pg_description d ON (d.objoid = a.attrelid AND d.classoid = 'pg_class'::regclass AND d.objsubid = a.attnum)
```
jimmyyih: Is there a reason why this LEFT JOIN on pg_description was moved up?  I think JOIN ordering usually matters for LEFT/RIGHT JOINs.  I'm not sure which is correct or if it matters but it would be good to double check.

kyeap-pivotal: It was an effort to make the `before6query` and the `masterQuery` behave and look similar, but we actually made a slight mistake.  The only change to the master query was some additional sql to get a column's collation. 
`masterQuery` should've been 
```
...
FROM pg_type t
JOIN pg_class c ON t.typrelid = c.oid
JOIN pg_attribute a ON t.typrelid = a.attrelid
LEFT JOIN pg_description d ON (d.objoid = a.attre...
...
```

lavjain: @jimmyyih I don't think that the LEFT JOIN order would matter in this case.

backup/queries_types.go
```diff
@@ -258,35 +258,37 @@ type Attribute struct {
...
 	LEFT JOIN pg_description d ON (d.objoid = a.attrelid AND d.classoid = 'pg_class'::regclass AND d.objsubid = a.attnum)
-	WHERE t.typtype = 'c'
-	ORDER BY t.oid, a.attnum;`
+	WHERE t.typtype = 'c' AND c.relkind = 'c';`
 
 	masterQuery := `SELECT
 	t.oid AS compositetypeoid,
 	quote_ident(a.attname) AS name,
 	pg_catalog.format_type(a.atttypid, a.atttypmod) AS type,
+	coalesce(quote_literal(d.description),'') AS comment
```
jimmyyih: Does this mean description and collation were swapped before in the metadata dump?

roicos: It was moved up from line 281.

jimmyyih: To clarify, was the description and collation swapped in the metadata SQL (not the query gpbackup calls against GPDB) that we generate for the user.  If so, a user using that metadata SQL to restore composite types could be wrong and generate errors during the metadata restore stage of gprestore.

lavjain: According to the defined struct, the fields were incorrect but it affected MASTER branch only.

kyeap-pivotal: The ordering of the attributes in queries will not affect the metadata SQL generated. The query results are put into a struct and a statement built from the values of the struct. If wrong metadata was printed, the issue would be in the function that generates composite type create statements.

jimmyyih: @lavjain The variable name of the stored query is not very good currently.  It says `masterQuery` but it actually means GPDB 6+ (the other one is called `before6query`).  So if the fields being incorrect creates wrong metadata SQL, we would need to put out a note for GPDB 6 users.

@kyeap-pivotal Cool, so we're fine then.  Nice.

PR #293, opened 2019-10-01
Go modules

kyeap-pivotal:
This is a PR for visibility. This is not intended to be merged. I believe having go1.13 might be a requirement for this branch to work. This is a small step to switch to using go modules.

The change to use sqlmock-go is a workaround. The change to common-go-lib has already been pushed in.


shivzone:
Would be useful for context to add the command you ran to convert dependencies from `dep` to go modules. Also which file is the auto generated and which is the actual file that track new depedancies

PR #294, opened 2019-10-02
Use pgcrypto extension in GPDB 5X_STABLE tests

jimmyyih:
GPDB 5X_STABLE recently had a commit merged in that changed pgcrypto
from a contrib module to an extension. Instead of calling the install
SQL file, we now need to use CREATE EXTENSION. This should go into
effect starting GPDB 5.22.0.

GPDB reference:
https://github.com/greenplum-db/gpdb/commit/5c751a06d52e43b2


mkiyama:
I have a some questions.
--Is there a doc update for this addition? A limitation of backing and restoring between different versions of 5.x?

--Do you know if this is going into 6.x? the 6.x FIPS PR is currently open https://github.com/greenplum-db/gpdb/pull/8383

jimmyyih:
@mkiyama 
> I have a some questions.
> --Is there a doc update for this addition? A limitation of backing and restoring between different versions of 5.x?

Probably not... although I haven't give it much thought until now.  @shivzone did mention it off-PR though.  Currently, we give directions to go to GPDB 5 latest docs to see how to install pgcrypto and the user would be able to choose which version of 5X they have... so as long as the GPDB 5.22 docs are updated to install pgcrypto as an extension, we should be good to go.  The gpbackup/gprestore utilities will just call the pgcrypto functions without caring how it was installed.

> --Do you know if this is going into 6.x? the 6.x FIPS PR is currently open [greenplum-db/gpdb#8383](https://github.com/greenplum-db/gpdb/pull/8383)

Are you asking if the 6X FIPS PR will affect gpbackup?  If so, most likely not.  GPDB 6 already has pgcrypto as an extension.  If you're asking if the 6X FIPS PR will be going into GPDB 6, I haven't been following the PR very closely.  I expect there to be more activity on the PR when the author Daniel comes back later this month or if someone on the Pivotal GPDB server team picks it up for him.

mkiyama:
thanks.

PR #296, opened 2019-10-08
Compute QuotedRoles only once per backup session

lavjain:


backup/predata_acl_test.go
```diff
@@ -322,60 +322,53 @@ ALTER DEFAULT PRIVILEGES FOR ROLE testrole GRANT USAGE ON TABLES TO somerole WIT
 		})
 	})
 	Describe("ParseACL", func() {
-		var quotedRoleNames map[string]string
-		BeforeEach(func() {
-			quotedRoleNames = map[string]string{
-				"testrole":  "testrole",
-				"Test|role": `"Test|role"`,
-			}
-		})
```
jimmyyih: How is this test working without the initial data? 🤔

lavjain: I was surprised at well but then saw https://github.com/greenplum-db/gpbackup/blob/d817df1d06137ceace4d3e63f28d7d2bbab2ac24/backup/predata_acl_test.go#L274-L276

backup/queries_acl.go
```diff
@@ -285,10 +285,11 @@ ORDER BY n.nspname, a.defaclobjtype, r.rolname`
 }
 
 func GetQuotedRoleNames(connectionPool *dbconn.DBConn) map[string]string {
-	quotedRoleNames := make(map[string]string)
-
-	quotedRoleNamesQuery := `
-SELECT rolname AS rolename, quote_ident(rolname) AS quotedrolename FROM pg_authid`
+	if len(quotedRoleNames) > 0 {
```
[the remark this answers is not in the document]
lavjain: Agree. I refactored this code over multiple iterations and totally forgot that it is being called in one place only now.

backup/predata_acl_test.go
jimmyyih: I'm not sure how lines 274 through 276 are related.  It doesn't include a quoted role and seems to just be two simple roles.

I looked a bit closer and the initial data is actually needed.  The `ParseACL()` function checks if quotedRoleNames global var is instantiated (uses its contents) or not (use the regex output).  In fact, the ParseACL test to test this was actually incorrect which gave you your false positive.

For more clarification, try this test fix patch:
```
diff --git a/backup/predata_acl_test.go b/backup/predata_acl_test.go
index b2d26ef..10fd1a3 100644
--- a/backup/predata_acl_test.go
+++ b/backup/predata_acl_test.go
@@ -346,7 +351,7 @@ ALTER DEFAULT PRIVILEGES FOR ROLE testrole GRANT USAGE ON TABLES TO somerole WIT
                        structmatcher.ExpectStructsToMatch(&expected, result)
                })
                It("parses an ACL string containing a role name with special characters", func() {
-                       aclStr := `"Test|role"=a/gpadmin`
+                       aclStr := `Test|role=a/gpadmin`
                        expected := backup.ACL{Grantee: `"Test|role"`, Insert: true}
                        result := backup.ParseACL(aclStr)
                        structmatcher.ExpectStructsToMatch(&expected, result)
```

I tried to fill in the global `quotedRoleNames` map but it wasn't working for some reason.  We'll need to look into this before we merge this PR.

PR #297, opened 2019-10-09
Fix warnings that were flagged by Goland IDE

lavjain:


backup/incremental.go
```diff
@@ -50,6 +50,7 @@ func GetLatestMatchingBackupTimestamp() string {
 	if latestMatchingBackupHistoryEntry == nil {
 		gplog.FatalOnError(errors.Errorf("There was no matching previous backup found with the flags provided. " +
 			"Please take a full backup."))
+		return ""
```
jmcatamney: This `return` will never occur because `FatalOnError` panics.  If this fix is due to an erroneous Goland warning, is there a way to "tell" it about the panic to silence the warning instead?

backup/queries_relations.go
```diff
@@ -364,16 +361,13 @@ func generateTableBatches(tables []Relation, batchSize int) []string {
 	var batches []string
 	i := 0
 	for i < len(tables) {
-
```
jimmyyih: This is the only newline removal I agree with in this file.

backup/queries_table_defs.go
```diff
@@ -307,47 +305,44 @@ func GetDistributionPolicies(connectionPool *dbconn.DBConn) map[uint32]string {
...
+	query := `SELECT oid, reloftype::pg_catalog.regtype AS value from pg_class WHERE reloftype != 0`
 	return selectAsOidToStringMap(connectionPool, query)
 }
 
 func GetTableReplicaIdentity(connectionPool *dbconn.DBConn) map[uint32]string {
 	if connectionPool.Version.Before("6") {
 		return map[uint32]string{}
 	}
-	query := `select oid, relreplident AS value from pg_class`
+	query := `SELECT oid, relreplident AS value from pg_class`
```
jimmyyih: Should probably uppercase the `from` keyword too.

PR #299, opened 2019-10-09
Simply extraneous SQL and format it uniformly

lavjain:


backup/queries_externals.go
```diff
@@ -14,65 +14,60 @@ import (
 )
 
 func GetExternalTableDefinitions(connectionPool *dbconn.DBConn) map[uint32]ExternalTableDefinition {
+	gplog.Verbose("Retrieving external table information")
 	execOptions := "'ALL_SEGMENTS', 'HOST', 'MASTER_ONLY', 'PER_HOST', 'SEGMENT_ID', 'TOTAL_SEGS'"
 	version4query := fmt.Sprintf(`
```
shivzone: Not a fan of keeping separate variables for query based on version. Can we instead go with `query` and set it based on the gpdb version

jimmyyih: Very much agreed.  The version check should wrap around the query formation, not the running of the query down below.  It would only make sense if the results variable from the query becomes different between versions... if that matters (it apparently didn't matter for resource group results from a previous PR).

PR #301, opened 2019-10-11
Fix gprestore verbose output when using --jobs flag

jimmyyih:
When the --jobs flag is given, gprestore would output "(table 1 of N)"
multiple times depending on the --jobs input. The last table to be
restored would also not output the expected "(table N of N)". This is
because the first restore for each parallel worker gets the same table
number (in this case 1) which will be outputted before restoring
data. After the restore is finished, each parallel worker atomically
increments the table number.

To fix the incorrect gprestore output, we initialize the table number
to 0 and have each parallel worker atomically increment the table
number after they are done with their restore.

---

Also, do not build automatically when running end_to_end tests

The end_to_end tests were still automatically building and installing
gpbackup and gprestore. This is not always desirable since we may want
to run current tests with old binaries. This somewhat relates to how
we earlier separated the `end_to_end` Makefile target to have an
`end_to_end_without_install` target that doesn't call the `build`
target.

Reference:
https://github.com/greenplum-db/gpbackup/commit/bf60576277b621bc

### DIFF
--- a/end_to_end/end_to_end_suite_test.go
+++ b/end_to_end/end_to_end_suite_test.go
@@ -222,6 +222,7 @@ var _ = Describe("backup end to end integration tests", func() {
 	const (
 		TOTAL_RELATIONS               = 37
 		TOTAL_RELATIONS_AFTER_EXCLUDE = 21
+		TOTAL_CREATE_STATEMENTS       = 9
 	)
 
 	var gpbackupPath, backupHelperPath, restoreHelperPath, gprestorePath, pluginConfigPath string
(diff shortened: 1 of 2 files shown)


shivzone:
Looks good. The only reservation i have is that we use uint32 to keep track of numTables during restore while we use int64 during backup.  Not sure if this limitation will be a problem during restore ?

jimmyyih:
> Looks good. The only reservation i have is that we use uint32 to keep track of numTables during restore while we use int64 during backup. Not sure if this limitation will be a problem during restore ?

Huh, that is an interesting observation.  Technically... yes, it is a problem.  Realistically... most likely very much not a problem.  Having a single database with uint32_max + 1 (~4.2 billion) number of tables to reach wraparound and break gprestore log output (only log output) would be rare.  Nonetheless, we should keep them consistent and use int64 to future-proof ourselves (e.g. if someone puts Greenplum on a year 2030 quantum server).  I'll make that change later.

jimmyyih:
Added commit to change the counter from uint32 to int64.

PR #303, opened 2019-10-16
Cancel blocked LOCK TABLE query during gpbackup DoCleanup

jimmyyih:
The gpbackup utility can leak a blocked SQL session when a user sends
SIGINT (Ctrl-C) while gpbackup is executing its LOCK TABLE query. This
only occurs when the query is blocked trying to obtain a simple
AccessShareLock on a table when another session has already acquired
an AccessExclusiveLock on that same table. The AccessExclusiveLock
could have been acquired by the other session explicitly via LOCK
TABLE or implicitly via commands like TRUNCATE and ALTER TABLE.

The reason the SQL session is leaked is because the Golang SQL
libraries do not cancel the queries for you when closing
connections. To do this, we must introduce the concept of Golang
Contexts which gives us a very basic cancel function we can call
during clean up. The query cancel itself does not block and there is
no cancel verification built in so we explicitly queue a ROLLBACK on
the main gpbackup SQL session to make sure the SIGINT makes it through
the libpq connection. The Context and its cancel function are stored
as globals for now as one-off variables but I can see it grow into a
big list since all other gpbackup queries could technically encounter
this too (but unlikely).

----

This PR also updates the DATA-DOG/go-sqlmock dependency since the gp-common-go-libs dependency needs to be updated and the go-sqlmock was not in sync.  I would suggest reviewing this PR by individual commits so that the random go-sqlmock dependency updates don't cause a distraction to the actual fix patch.

### DIFF
--- a/backup/queries_relations.go
+++ b/backup/queries_relations.go
@@ -6,6 +6,7 @@ package backup
  */
 
 import (
+	"context"
 	"fmt"
 	"strings"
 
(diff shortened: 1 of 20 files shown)


backup/queries_relations.go
```diff
@@ -348,6 +356,11 @@ func LockTables(connectionPool *dbconn.DBConn, tables []Relation) {
 		progressBar.Add(currentBatchSize)
 	}
 
+	// We're done grabbing table locks. Unset the Context globals
+	// so we don't use them during DoCleanup.
+	queryContext = context.TODO()
```
shivzone: isn't there an alternate way to unset the context apart form TODO ?

jimmyyih: I can and would like to set it to `nil`... but the Golang documentation states the following:

```
Do not pass a nil Context, even if a function permits it. Pass context.TODO if you are unsure about which Context to use.
```

It's a bit hard to interpret IMO but I just took it as don't ever set a Context to `nil`.  IMO, it shouldn't matter.  Thoughts?

PR #304, opened 2019-10-16
Fix restore --with-stats behaviour when --include-table is specified

roicos:

Previously when run gprestore --with-stats it ignored --include-table flag for restore statistics.

This happened because we set `filterRelations` to `false` when call `GetRestoreMetadataStatement` in `restoreStatistics`, so it didn't filter relations and restored statistics for all the tables.

We set it to `true` and added end-to-end tests for:

- runs gprestore with --include-table flag to only restore tables specified
- restores statistics only for tables specified in --include-table flag when runs gprestore with with-stats flag

### DIFF
--- a/end_to_end/end_to_end_suite_test.go
+++ b/end_to_end/end_to_end_suite_test.go
@@ -997,6 +997,34 @@ var _ = Describe("backup end to end integration tests", func() {
 			assertDataRestored(restoreConn, schema2TupleCounts)
 
 		})
+		It("runs gprestore with --include-table flag to only restore tables specified ", func() {
+			testhelper.AssertQueryRuns(backupConn, "CREATE TABLE public.table_to_include_with_stats(i int)")
+			defer testhelper.AssertQueryRuns(backupConn, "DROP TABLE public.table_to_include_with_stats")
+			testhelper.AssertQueryRuns(backupConn, "INSERT INTO public.table_to_include_with_stats VALUES (1)")
+			timestamp := gpbackup(gpbackupPath, backupHelperPath, "--backup-dir", backupDir)
+			gprestore(gprestorePath, restoreHelperPath, timestamp, "--redirect-db", "restoredb", "--backup-dir", backupDir, "--include-table", "public.table_to_include_with_stats")
+
+			assertRelationsCreated(restoreConn, 1)
+
+			localSchemaTupleCounts := map[string]int{
+				`public."table_to_include_with_stats"`: 1,
+			}
+			assertDataRestored(restoreConn, localSchemaTupleCounts)
+			assertArtifactsCleaned(restoreConn, timestamp)
+		})
+		It("restores statistics only for tables specified in --include-table flag when runs gprestore with with-stats flag", func() {
+			testhelper.AssertQueryRuns(backupConn, "CREATE TABLE public.table_to_include_with_stats(i int)")
+			defer testhelper.AssertQueryRuns(backupConn, "DROP TABLE public.table_to_include_with_stats")
+			timestamp := gpbackup(gpbackupPath, backupHelperPath, "--with-stats", "--backup-dir", backupDir)
+			statFiles, err := filepath.Glob(filepath.Join(backupDir, "*-1/backups/*", timestamp, "*statistics.sql"))
+			Expect(err).ToNot(HaveOccurred())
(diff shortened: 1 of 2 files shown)


end_to_end/end_to_end_suite_test.go
```diff
@@ -997,6 +997,32 @@ var _ = Describe("backup end to end integration tests", func() {
...
+			assertRelationsCreated(restoreConn, 1)
+
+			localSchemaTupleCounts := map[string]int{
+				`public."table_to_include"`: 1,
+			}
+			assertDataRestored(restoreConn, localSchemaTupleCounts)
+			assertArtifactsCleaned(restoreConn, timestamp)
+		})
+		It("restores statistics only for tables specified in --include-table flag when runs gprestore with with-stats flag", func() {
+			testhelper.AssertQueryRuns(backupConn, "CREATE TABLE public.table_to_include(i int)")
```
jimmyyih: Could we use a different unique table name here?  It makes it much easier to see the order of operations in GPDB server logs for debugging test failures.  The tests run in random order so you don't know which test runs first.

end_to_end/end_to_end_suite_test.go
```diff
@@ -997,6 +997,32 @@ var _ = Describe("backup end to end integration tests", func() {
...
+			defer testhelper.AssertQueryRuns(backupConn, "DROP TABLE public.table_to_include")
+			timestamp := gpbackup(gpbackupPath, backupHelperPath, "--with-stats", "--backup-dir", backupDir)
+			filepath.Glob(filepath.Join(backupDir, "*-1/backups/*", timestamp, "*statistics.sql"))
```
jimmyyih: What is this Glob for?  It doesn't seem like we're using the returned output of the Glob function to do any kind of validation.

restore/restore.go
```diff
@@ -270,7 +270,7 @@ func restoreStatistics() {
 	}
 	statisticsFilename := globalFPInfo.GetStatisticsFilePath()
 	gplog.Info("Restoring query planner statistics from %s", statisticsFilename)
-	statements := GetRestoreMetadataStatements("statistics", statisticsFilename, []string{}, []string{}, true, false)
+	statements := GetRestoreMetadataStatements("statistics", statisticsFilename, []string{}, []string{}, true, true)
```
lavjain: Looks like the fix was simple :-)

end_to_end/end_to_end_suite_test.go
```diff
@@ -997,6 +997,32 @@ var _ = Describe("backup end to end integration tests", func() {
...
+
+			gprestore(gprestorePath, restoreHelperPath, timestamp, "--redirect-db", "restoredb", "--with-stats", "--backup-dir", backupDir, "--include-table", "public.table_to_include")
+
+			rawCount := dbconn.MustSelectString(restoreConn, "select count(*) from pg_stat_all_tables where schemaname='public'")
```
lavjain: Please convert all sql keywords to uppercase for consistency.

shivzone: not a fan of enforcing lot of convention with tests :)

end_to_end/end_to_end_suite_test.go
roicos: Any preferences?

end_to_end/end_to_end_suite_test.go
jimmyyih: I am a fan of enforcing uppercase SQL keywords.  It's good practice.

end_to_end/end_to_end_suite_test.go
jimmyyih: IMO, good practice would be to name your test table very similar to what the test is doing for uniqueness.  I guess you could name this one `public.table_to_include_stats_only`.

roicos: Renamed to 'table_to_include_with_stat'

end_to_end/end_to_end_suite_test.go
roicos: Added assertions.

end_to_end/end_to_end_suite_test.go
roicos: Done.

restore/restore.go
roicos: As soon as you know what causes the bug:)

PR #306, opened 2019-10-21
Adds test to NOT VALID clause.

roicos:
- Gpbackup supports NOT VALID option on constraints
  This allow to backup and restore legacy data that violate
  constraints being introduced later.
  This commit add behavior test to make sure that
    We can backup and restore legacy data
    It also restores new constraint
    New constraint works for new data

Authored-by: Ekaterina Dontsova <edontsova@pivotal.io>

### DIFF
--- a/end_to_end/end_to_end_suite_test.go
+++ b/end_to_end/end_to_end_suite_test.go
@@ -1125,7 +1125,7 @@ var _ = Describe("backup end to end integration tests", func() {
 			// session trying to acquire AccessShareLock on foo2.
 			var afterLockCount int
 			_ = backupConn.Get(&afterLockCount, checkLockQuery)
-                        Expect(afterLockCount).To(Equal(0))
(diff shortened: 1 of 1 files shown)


end_to_end/end_to_end_suite_test.go
```diff
@@ -1245,6 +1245,30 @@ PARTITION BY LIST (gender)
 			assertDataRestored(restoreConn, map[string]int{"public.corrupt_table": 0, "public.good_table1": 10, "public.good_table2": 10})
 
 		})
+		It("backup and restore all data when NOT VALID option on constraints is specified", func() {
+			testhelper.AssertQueryRuns(backupConn, "CREATE TABLE legacy_table_violate_constrains (a int)")
```
jimmyyih: s/legacy_table_violate_constrains/legacy_table_violate_constraints/

PR #311, opened 2019-10-22
Log error tables for metadata and data restore

shivzone:
Updated `gprestore` to log tables with errors during restore when `on-error-continue` flag is set. The idea behind this change is to allow users to 1. identify errors 2. allow users to use `--exclude-table-file` with the generated error table list for a follow up `gprestore` iteration.
error tables are tracked and logged separately for metadata errors and data related errors as there could be overlapping tables. Separate files gives users more granularity in controlling subsequent iterations of `gprestore` with the error file.

Updated the `end_to_end/resources/corrupt-db.tar.gz` (used as an input to handling of corrupt database restore) to use a new username `e2euser` instead of `pivotal`


jimmyyih:
Seems kinda weird to have repeats of the same relation:
```
(dlv) p errorTablesMetadata
[]string len: 2, cap: 2, [
	"public.foobar",
	"public.foobar",
]
(dlv) p errorTablesData
[]string len: 1, cap: 1, [
	"public.foobar",
]
```

This was done by simply changing the distribution key of the table in the metadata SQL file.  We get 2 errors for metadata (`CREATE TABLE` and `ALTER TABLE OWNER TO`) and 1 error for the data restore since the table doesn't exist.  We should probably make the entries unique for each file since there could be many repeats and users would probably not want to uniq the files themselves.  Maybe use sets?

shivzone:
@jimmyyih Good catch. I missed that during testing

shivzone:
Here is why golang doesn't have sets https://stackoverflow.com/questions/34018908/golang-why-dont-we-have-a-set-datastructure

restore/parallel.go
```diff
@@ -15,6 +15,7 @@ import (
 )
 
 func executeStatementsForConn(statements chan utils.StatementWithType, fatalErr *error, numErrors *int32, progressBar utils.ProgressBar, whichConn int) {
+	var mutex = &sync.Mutex{}
```
lavjain: What is the reasoning behind using the address of the Mutex instead of Mutex?

lavjain: On second thoughts, the mutex here should not be a local variable.

shivzone: you are right. this didn't fail in the tests since metadata restore isn't parallel

shivzone: address of muex vs the mutex itself has the same effect

PR #312, opened 2019-10-31
Support Materialized views

kyeap-pivotal:
Support materialized views

    Add backup support for materialized views in GPDB6.1. Since they are
    closely related to views, the query was reused and GetViews was changed
    to also query information about materialized views. The reason they are
    then sorted into separate object types is to make it easier to implement
    materialized views ACL.

### DIFF
--- a/backup/predata_acl.go
+++ b/backup/predata_acl.go
@@ -233,7 +233,7 @@ func ParseACL(aclStr string) *ACL {
 func (obj ObjectMetadata) GetPrivilegesStatements(objectName string, objectType string, columnName ...string) string {
 	statements := make([]string, 0)
 	typeStr := fmt.Sprintf("%s ", objectType)
-	if objectType == "VIEW" || objectType == "FOREIGN TABLE" {
+	if objectType == "VIEW" || objectType == "FOREIGN TABLE" || objectType == "MATERIALIZED VIEW" {
 		typeStr = ""
 	} else if objectType == "COLUMN" {
 		typeStr = "TABLE "
@@ -311,7 +311,7 @@ func createPrivilegeStrings(acl ACL, objectType string) (string, string) {
 	case "SEQUENCE":
 		hasAllPrivileges = acl.Select && acl.Update && acl.Usage
 		hasAllPrivilegesWithGrant = acl.SelectWithGrant && acl.UpdateWithGrant && acl.UsageWithGrant
-	case "TABLE":
+	case "TABLE", "VIEW", "MATERIALIZED VIEW":
 		hasAllPrivileges = acl.Select && acl.Insert && acl.Update && acl.Delete && acl.Truncate && acl.References && acl.Trigger
 		hasAllPrivilegesWithGrant = acl.SelectWithGrant && acl.InsertWithGrant && acl.UpdateWithGrant && acl.DeleteWithGrant &&
 			acl.TruncateWithGrant && acl.ReferencesWithGrant && acl.TriggerWithGrant
@@ -321,10 +321,6 @@ func createPrivilegeStrings(acl ACL, objectType string) (string, string) {
 	case "TYPE":
 		hasAllPrivileges = acl.Usage
 		hasAllPrivilegesWithGrant = acl.UsageWithGrant
-	case "VIEW":
-		hasAllPrivileges = acl.Select && acl.Insert && acl.Update && acl.Delete && acl.Truncate && acl.References && acl.Trigger
-		hasAllPrivilegesWithGrant = acl.SelectWithGrant && acl.InsertWithGrant && acl.UpdateWithGrant && acl.DeleteWithGrant &&
-			acl.TruncateWithGrant && acl.ReferencesWithGrant && acl.TriggerWithGrant
 	}
 	if hasAllPrivileges {
 		privStr = "ALL"
--- a/backup/predata_relations.go
+++ b/backup/predata_relations.go
@@ -281,11 +281,29 @@ func PrintAlterSequenceStatements(metadataFile *utils.FileWithByteCount, toc *ut
 	}
 }
 
+// A view's column names are automatically factored into it's definition.
 func PrintCreateViewStatement(metadataFile *utils.FileWithByteCount, toc *utils.TOC, view View, viewMetadata ObjectMetadata) {
 	start := metadataFile.ByteCount
+	// Option's keyword WITH is expected to be prepended to its options in the SQL statement
 	metadataFile.MustPrintf("\n\nCREATE VIEW %s%s AS %s\n", view.FQN(), view.Options, view.Definition)
 
 	section, entry := view.GetMetadataEntry()
 	toc.AddMetadataEntry(section, entry, start, metadataFile.ByteCount)
 	PrintObjectMetadata(metadataFile, toc, viewMetadata, view, "")
+}
+
+// A materialized view's column names are automatically factored into it's definition.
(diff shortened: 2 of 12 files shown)


jimmyyih:
Apparently, it's not going into GPDB 6.1.  Since these commits rely heavily on the version string, I'd wait until that PR is committed and the GPDB version string is 100% established instead of guessing (some users actually do pull down gpbackup master and run latest and greatest so having the wrong version string would be bad).

backup/predata_acl.go
```diff
@@ -325,6 +325,10 @@ func createPrivilegeStrings(acl ACL, objectType string) (string, string) {
 		hasAllPrivileges = acl.Select && acl.Insert && acl.Update && acl.Delete && acl.Truncate && acl.References && acl.Trigger
 		hasAllPrivilegesWithGrant = acl.SelectWithGrant && acl.InsertWithGrant && acl.UpdateWithGrant && acl.DeleteWithGrant &&
 			acl.TruncateWithGrant && acl.ReferencesWithGrant && acl.TriggerWithGrant
+	case "MATERIALIZED VIEW":
```
lavjain: Materialized view can be combined with View case above

backup/queries_relations.go
```diff
@@ -306,27 +308,101 @@ func (v View) FQN() string {
...
+			materializedViews = append(materializedViews, makeMaterializedView(view))
+		} else {
+			regularViews = append(regularViews, view)
+		}
+	}
+
+	return regularViews, materializedViews
+}
+
+type MaterializedView struct {
```
lavjain: This structure may not be needed if there is isMaterialized field in the View struct

shivzone:
> Apparently, it's not going into GPDB 6.1. Since these commits rely heavily on the version string, I'd wait until that PR is committed and the GPDB version string is 100% established instead of guessing (some users actually do pull down gpbackup master and run latest and greatest so having the wrong version string would be bad).

Instead i would check for > 7.0 which would help us test this against master builds. We can always update this to a specific 6.x version later

backup/predata_acl.go
kyeap-pivotal: It can be. We could also combine with TABLE. 
TYPE, LANGUAGE, FOREIGN SERVER, FOREIGN DATA WRAPPER can also be combined. But since half the cases are unique, I would say lets keep them all separate.

backup/queries_relations.go
kyeap-pivotal: True too, but I would like to keep the code base consistent.

jimmyyih:
> > Apparently, it's not going into GPDB 6.1. Since these commits rely heavily on the version string, I'd wait until that PR is committed and the GPDB version string is 100% established instead of guessing (some users actually do pull down gpbackup master and run latest and greatest so having the wrong version string would be bad).
> 
> Instead i would check for > 7.0 which would help us test this against master builds. We can always update this to a specific 6.x version later

Yeah, that seems like the best approach.  Let's do that.

backup/predata_acl.go
jimmyyih: Agree with Lav here.  Letting it fall through is good practice.  We'll change the other ones later.

backup/predata_relations.go
```diff
@@ -281,11 +281,29 @@ func PrintAlterSequenceStatements(metadataFile *utils.FileWithByteCount, toc *ut
 	}
 }
 
+// A view's column names are automatically factored into it's deinition.
```
jimmyyih: s/deinition/definition/

backup/predata_relations.go
```diff
@@ -281,11 +281,29 @@ func PrintAlterSequenceStatements(metadataFile *utils.FileWithByteCount, toc *ut
...
 	start := metadataFile.ByteCount
+	// Option's keyword WITH is expected to be prepended to its options in the SQL statement
 	metadataFile.MustPrintf("\n\nCREATE VIEW %s%s AS %s\n", view.FQN(), view.Options, view.Definition)
 
 	section, entry := view.GetMetadataEntry()
 	toc.AddMetadataEntry(section, entry, start, metadataFile.ByteCount)
 	PrintObjectMetadata(metadataFile, toc, viewMetadata, view, "")
 }
+
+// A materialized view's column names are automatically factored into it's deinition.
```
jimmyyih: s/deinition/definition/

backup/predata_relations.go
```diff
@@ -281,11 +281,29 @@ func PrintAlterSequenceStatements(metadataFile *utils.FileWithByteCount, toc *ut
...
+func PrintCreateMaterializedViewStatement(metadataFile *utils.FileWithByteCount, toc *utils.TOC, mview MaterializedView, mviewMetadata ObjectMetadata) {
+	start := metadataFile.ByteCount
+	var tablespaceClause string
+	if mview.Tablespace != "" {
+		tablespaceClause = fmt.Sprintf(" TABLESPACE %s", mview.Tablespace)
+	}
+	// Option's keyword WITH is expected to be prepended to its options in the SQL statement
+	// Remove trailining ';' at the end of materialized view's definition
```
jimmyyih: s/trailining/trailing/

backup/queries_relations.go
```diff
@@ -306,27 +308,101 @@ func (v View) FQN() string {
...
+
+	var whereClause string
+	if connectionPool.Version.Before("6.1") {
+		fromClause += `
+	WHERE c.relkind = 'v'::"char"`
+	} else {
+		fromClause += `
+	WHERE c.relkind IN ('m', 'v')`
+	}
+	whereClause += fmt.Sprintf(`
```
jimmyyih: It doesn't look like `whereClause` needs to be explicitly declared and could be a short assignment (`:=`).

backup/predata_acl.go
kyeap-pivotal: >Edit: Ugh.... I just saw how Golang does case fall through (they swapped default fall through for default break)... I think we should still do it though.

I can also put in explicit breaks for readability.

jimmyyih: No, that's not necessary.  Just keep it with what the Golang devs want.

If you want, you can just keep it as is and we'll refactor it later this week to add all the fall throughs in a one-sweep refactoring commit.

backup/queries_relations.go
kyeap-pivotal: oh oops. I accidentally appended to the wrong clause. got lucky because it was all  concatenated in the end.

backup/predata_acl.go
```diff
@@ -283,48 +283,57 @@ func createPrivilegeStrings(acl ACL, objectType string) (string, string) {
...
 		hasAllPrivilegesWithGrant = acl.CreateWithGrant
+		break
 	case "TYPE":
 		hasAllPrivileges = acl.Usage
 		hasAllPrivilegesWithGrant = acl.UsageWithGrant
-	case "VIEW":
-		hasAllPrivileges = acl.Select && acl.Insert && acl.Update && acl.Delete && acl.Truncate && acl.References && acl.Trigger
-		hasAllPrivilegesWithGrant = acl.SelectWithGrant && acl.InsertWithGrant && acl.UpdateWithGrant && acl.DeleteWithGrant &&
-			acl.TruncateWithGrant && acl.ReferencesWithGrant && acl.TriggerWithGrant
+		break
```
jimmyyih: Adding explicit break keywords to all the other cases is out of scope for this PR.  Although I like it, it's best to keep it as the default implicit break so no one thinks the default is fallthrough if they don't have the break.

Also, does `case "TABLE" || "VIEW" || "MATERIALIZED VIEW":` even work?  Trying the same in a small go program gives an error.  Seems comma split works though as `case "TABLE", "VIEW", "MATERIALIZED VIEW":`.

PR #313, opened 2019-11-01
gprestore restores schemas based on include-tables

shivzone:
If include-table is configured only schemas in this list will be restored.

`GetRestoreMetadataStatements` used for deriving the schema statements now use the filtered relations to infer the schemas.


shivzone:
@jimmyyih i felt the same and lets tackle that separately as you said

PR #315, opened 2019-11-01
Ubuntu debian spike

roicos:
To support Ubuntu 18 on  GPDB 6 we added building .deb packages and then building .gppkg 
We also added test-on-local-cluster-ubuntu task and GPDB6-ubuntu job to test on remote cluster.
Some refactoring was done in the process to separate building go binaries from building .gppkgs for RHEL, SLES and ubuntu.


jimmyyih:
Could you restructure the commits a bit?  I spent ~5 minutes trying to figure out what the new spike pipeline would give us and then found out that the next commit deleted it.

PR #316, opened 2019-11-06
Sort package imports

lavjain:
Also remove a redundant `make depend` call

### DIFF
--- a/ci/scripts/test-gpbackup-manager.bash
+++ b/ci/scripts/test-gpbackup-manager.bash
@@ -46,7 +46,6 @@ psql -d postgres -c "create extension pgcrypto"
 
 # Test gpbackup manager
 pushd \$GOPATH/src/github.com/pivotal/gp-backup-manager
-  make depend
   make unit
   make integration
   make end_to_end_without_install
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -68,7 +68,7 @@
   version = "v1.0"
 
 [[projects]]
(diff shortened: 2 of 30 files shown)


shivzone:
Can you instead run `make format` and commit changes based on that instead of hunting down things manually or using an IDE.

shivzone:
Why do we need the updates to `Gopkg.lock` ?

kyeap-pivotal:
> Why do we need the updates to `Gopkg.lock` ?

Gopkg.lock is usually also updated whenever Gopkg.toml is changed. Gopkg.lock is the output of the dependency solver. Contains information such as the exact commit of the dependency we are using.

ci/scripts/test-gpbackup-manager.bash
```diff
@@ -46,7 +46,6 @@ psql -d postgres -c "create extension pgcrypto"
 
 # Test gpbackup manager
 pushd \$GOPATH/src/github.com/pivotal/gp-backup-manager
-  make depend
```
kyeap-pivotal: does this break the pipeline? make depend installs ginkgo. which is required for `make unit`, `integration`, and `end_to_end`

lavjain: According to the makefile for gp-backup-manager, `make integration` is already running depend
```
integration : depend
		ginkgo $(GINKGO_FLAGS) integration 2>&1
```

lavjain: The dev pipeline is green

PR #319, opened 2019-11-19
Verify plugin config attributes executablepath and options

kyeap-pivotal:
Previously, gpbackup would attempt to continue with plugin API call if
the required yaml field 'executablepath' was missing due to being
forgotten or mispelled.  The missing executablepath would then cause
gpbackup to format the plugin API call badly and fail.  This commit add
a validation check which will cause gpbackup to exit early if either
'exectuablepath' and give the users a better error message.


kyeap-pivotal:
yaml unmarshal is unable to tell is options is missing or if it has no attributes.

kyeap-pivotal:
removed verification against `options` attribute

PR #320, opened 2019-11-20
Support Materialized views in GPDB6.2+

kyeap-pivotal:
Materialized views were backported from GPDB7 to GPDB6.2+. We changed
the version control logic to check for matviews on GPDB version 6.2 and
refactored the query to get all views.

### DIFF
--- a/backup/queries_relations.go
+++ b/backup/queries_relations.go
@@ -298,7 +298,7 @@ func (v View) FQN() string {
 }
 
 // This function retrieves both regular views and materialized views.
-// Materialized views were introduced in GPDB 7.
+// Materialized views were introduced in GPDB 7 and backported to GPDB 6.2.
 func GetAllViews(connectionPool *dbconn.DBConn) (regularViews []View, materializedViews []MaterializedView) {
 	selectClause := `
 	SELECT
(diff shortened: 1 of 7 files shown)


backup/queries_relations.go
```diff
@@ -308,31 +308,18 @@ func GetAllViews(connectionPool *dbconn.DBConn) (regularViews []View, materializ
 		pg_get_viewdef(c.oid) AS definition`
 	if connectionPool.Version.AtLeast("6") {
```
jimmyyih: This should technically be okay, but was there any effort to make sure materialized views on GPDB 6.0 or 6.1 would be dumped correctly to not change this to `AtLeast("6.2")`?  And yes, 6.0 and 6.1 will be able to have materialized views technically via binary swap downgrade (hopefully they added that case to binary swap tests).  I'll have to look at the backport PR again to see if anything major changed related to materialized view catalog metadata.

jimmyyih: Actually, creating a materialized view on GPDB 6.2 and then downgrading would make all subsequent backups not restorable.  I'll give this a try in a bit just to confirm.

jimmyyih: I've confirmed my scenario to be true.  This needs to be changed to check for version string 6.2.

kyeap-pivotal: The fix would be then to bring back and modify the WHERE clause filtering originally deleted in this PR.
```
	var whereClause string
	if connectionPool.Version.Before("6.2") {
		whereClause += `
	WHERE c.relkind = 'v'::"char"`
	} else {
		whereClause += `
	WHERE c.relkind IN ('m', 'v')`
	}
```
Materialized views created in 6.2 can continue to exist in gpdb 6.0 and 6.1 after binary downgrade. Backups can still succeed, but will not be able to backup materialized view.

jimmyyih:
Looks fine.  Only complaint I have is the inconsistency of "GPDB6.2" and "gpdb6.2".  Preferably, I would make it "GPDB 6.2" (with the space in between).

kyeap-pivotal:
> Looks fine. Only complaint I have is the inconsistency of "GPDB6.2" and "gpdb6.2". Preferably, I would make it "GPDB 6.2" (with the space in between).

Ressolved.

PR #323, opened 2019-12-13
Support incremental restore without DDL

roicos:

 - Adds --incremental flag for gprestore
 - Only restores the last entry fro restore plan
 - Filters schemas and relations based on DB state (existing schemas and relations) 
    and restore plan
- Respects --include/exclude-schemas/tables flag if provided

### DIFF
--- a/restore/restore.go
+++ b/restore/restore.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"os"
 	"runtime/debug"
+	"strings"
 	"sync"
 
 	"github.com/greenplum-db/gp-common-go-libs/cluster"
@@ -41,6 +42,7 @@ func SetFlagDefaults(flagSet *pflag.FlagSet) {
 	flagSet.StringSlice(utils.INCLUDE_SCHEMA, []string{}, "Restore only the specified schema(s). --include-schema can be specified multiple times.")
 	flagSet.StringSlice(utils.INCLUDE_RELATION, []string{}, "Restore only the specified relation(s). --include-table can be specified multiple times.")
 	flagSet.String(utils.INCLUDE_RELATION_FILE, "", "A file containing a list of fully-qualified relation(s) that will be restored")
+	flagSet.Bool(utils.INCREMENTAL, false, "Only restore data for all heap tables and only AO tables that have been modified since the last backup")
 	flagSet.Bool(utils.METADATA_ONLY, false, "Only restore metadata, do not restore data")
 	flagSet.Int(utils.JOBS, 1, "Number of parallel connections to use when restoring table data and post-data")
 	flagSet.Bool(utils.ON_ERROR_CONTINUE, false, "Log errors and continue restore, instead of exiting on first error")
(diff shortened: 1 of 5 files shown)


restore/restore.go
```diff
@@ -254,7 +333,14 @@ func restorePostdata(metadataFilename string) {
 		return
 	}
 	gplog.Info("Restoring post-data metadata")
-	statements := GetRestoreMetadataStatements("postdata", metadataFilename, []string{}, []string{}, true, true)
+
+	inSchemas := MustGetFlagStringSlice(utils.INCLUDE_SCHEMA)
+	exSchemas := MustGetFlagStringSlice(utils.EXCLUDE_SCHEMA)
+	inRelations := MustGetFlagStringSlice(utils.INCLUDE_RELATION)
+	exRelations := MustGetFlagStringSlice(utils.EXCLUDE_RELATION)
+	filters := NewFilters(inSchemas, exSchemas, inRelations, exRelations)
```
lavjain: The above 5 lines could be refactored into a function that returns a filter since `restoreStatistics` also uses the same code.

restore/wrappers.go
```diff
@@ -317,3 +345,49 @@ func RestoreSchemas(schemaStatements []utils.StatementWithType, progressBar util
...
+	return existingTableFQNs, err
+}
+
+func GetExistingSchemas()([]string, error){
+	existingSchemas := make([]string, 0)
+
+	query := `SELECT n.nspname AS "Name"
+FROM pg_catalog.pg_namespace n
+WHERE n.nspname !~ '^pg_' AND n.nspname <> 'information_schema'
+ORDER BY 1;`
```
lavjain: Please indent the SQL statement

restore/wrappers.go
```diff
@@ -317,3 +345,49 @@ func RestoreSchemas(schemaStatements []utils.StatementWithType, progressBar util
...
+
+	query := `SELECT quote_ident(n.nspname) || '.' || quote_ident(c.relname)
+FROM pg_catalog.pg_class c
+    LEFT JOIN pg_catalog.pg_namespace n ON n.oid = c.relnamespace
+WHERE c.relkind IN ('r','')
+AND c.relstorage IN ('h', 'a', 'c','')
+     AND n.nspname <> 'pg_catalog'
+     AND n.nspname <> 'information_schema'
+     AND n.nspname !~ '^pg_toast'
+ORDER BY 1;`
```
lavjain: Please indent the SQL statement

restore/wrappers.go
```diff
@@ -317,3 +345,49 @@ func RestoreSchemas(schemaStatements []utils.StatementWithType, progressBar util
 		gplog.Error("Encountered %d errors during schema restore; see log file %s for a list of errors.", numErrors, gplog.GetLogFilePath())
 	}
 }
+
+func filtersEmpty(filters Filters) bool {
+	return len(filters.includeSchemas) == 0 && len(filters.excludeSchemas) == 0 && len(filters.includeRelations) == 0 && len(filters.excludeRelations) == 0
+}
```
lavjain: `filtersEmpty` should be moved to Line 46

restore/wrappers.go
```diff
@@ -224,36 +243,39 @@ func FindHistoricalPluginVersion(timestamp string) string {
  * Metadata and/or data restore wrapper functions
  */
 
-func GetRestoreMetadataStatements(section string, filename string, includeObjectTypes []string, excludeObjectTypes []string, filterSchemas bool, filterRelations bool) []utils.StatementWithType {
+func GetRestoreMetadataStatements(section string, filename string, includeObjectTypes []string, excludeObjectTypes []string) []utils.StatementWithType {
+	filters := Filters{}
+	var statements []utils.StatementWithType
+	statements = GetRestoreMetadataStatementsFiltered(section, filename, includeObjectTypes, excludeObjectTypes, filters)
+	return statements
```
lavjain:
```
return GetRestoreMetadataStatementsFiltered(section, filename, includeObjectTypes, excludeObjectTypes, Filters{})
```

restore/restore.go
```diff
@@ -214,31 +281,43 @@ func restorePredata(metadataFilename string) {
...
-func restoreData(fpInfoList []backup_filepath.FilePathInfo, gucStatements []utils.StatementWithType) {
+func restoreData() {
 	if wasTerminated {
 		return
 	}
-	latestRestorePlan := backupConfig.RestorePlan
+	restorePlan := backupConfig.RestorePlan
+	restorePlanEntries := make([]backup_history.RestorePlanEntry, 0)
+	if MustGetFlagBool(utils.INCREMENTAL) {
+		restorePlanEntries = append(restorePlanEntries, restorePlan[len(backupConfig.RestorePlan)-1])
```
lavjain: Not very clear on why we need just the last entry from the plan

roicos: It's the way we are doing incremental restore. If we have t1,t2,t3,t4,t5 in restore plan and the actual state of db is t2 (user skip t3  and t4 in restore) we only apply t5 on top of the existing state. We do not track all the restores customer have done. So to incrementally restore db from t2 to t5  he has to go through all the steps t3, t4, t5.

restore/restore.go
roicos: We could have more refactoring in mind to separate user filters from system filters in the future.

PR #324, opened 2019-12-14
Use manually cached gpbackup 1.12.1 dependencies

kyeap-pivotal:
The gpbackup version 1.12.1 is tested against the latest version of GPDB5
    to ensure the version of gpbackup shipped with GPDB5X is not broken.
    Unfortunately, dep struggles to successfully download gpbackup
    dependencies sometimes so we implement a manual cache workaround. This
    commit unpacks the gpbackup 1.12.1 dependencies so it does not need to be
    downloaded by dep.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -52,26 +52,26 @@ $(GOLANG_LINTER) :
 lint : $(GOLANG_LINTER)
 		golangci-lint run --tests=false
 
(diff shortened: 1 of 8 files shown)


lavjain:
We can get rid of the depend target since no one is using it.

PR #329, opened 2020-01-15
Sync the contents of metadata sql file before closing

lavjain:
when writing to disk during gpbackup

### DIFF
--- a/utils/util.go
+++ b/utils/util.go
@@ -32,7 +32,7 @@ func OpenFileForWrite(filename string) (*os.File, error) {
 }
 
 func WriteToFileAndMakeReadOnly(filename string, contents []byte) error {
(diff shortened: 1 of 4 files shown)


utils/util.go
```diff
@@ -27,6 +27,34 @@ const MINIMUM_GPDB5_VERSION = "5.1.0"
  * General helper functions
  */
 
+func OpenFileForWrite(filename string) (*os.File, error) {
+	return os.OpenFile(filename, os.O_CREATE | os.O_EXCL | os.O_WRONLY, 0644)
```
shivzone: Unnecessary overhead of wrapping this in yet another function. Can as well invoke this directly and allow more granular file mode control

PR #332, opened 2020-01-17
Updates to relation and schema flags

shivzone:
Added --include-schema-file and --include-schema-file options for `gpbackup` and `gprestore`
This would allow us to configure multiple schemas in the file to include or exclude.

All the relation and schema flags --include-table --exclude-table --include-schema --exclude-schema
have been refactored to have a uniform behavior of accepting only one value per flag.
For example if you need to include tables public.t1 and public.t2 you will need to do the following
--include-table public.t1 --include-table public.t2 and not --include-table public.t1, public.t2

Internally use a StringArray (which accepts ',' as part of a string) to capture the flags as opposed to StringSlice. StringSlice would split a value say 'foo,bar' as two separate entries, while StringArray one take it as one entry

Pending: Additional end to end test for the new schema-file options


options/options.go
```diff
@@ -14,42 +14,38 @@ import (
 
 type Options struct {
 	includedRelations         []string
+	excludedRelations		  []string
```
jimmyyih: Although I don't agree with it... it's single spaces (not tabs) between the variable name and its type... if we want to keep consistency.

restore/validate_test.go
```diff
@@ -89,16 +85,19 @@ var _ = Describe("restore/validate tests", func() {
 			Expect(resultRelations).To(ConsistOf(expectedRelations))
 		})
 		It("filters on include relations", func() {
-			cmdFlags.Set(utils.INCLUDE_RELATION, "s1.table1,s2.table2")
-			expectedRelations := []string{"s1.table1", "s2.table2"}
+			//runtime.Breakpoint()
```
jimmyyih: Leftover?

restore/wrappers.go
```diff
@@ -146,8 +146,6 @@ func InitializeFilterLists() {
 }
 
 func BackupConfigurationValidation() {
-	InitializeFilterLists()
```
jimmyyih: Was this really dead code and we were initializing twice or something?  That'd be interesting if true.

shivzone: `restore` was using this function while backup is using a separate mechanism `options.NewOptions(cmdFlags)`. restore has been modified to also using the same approach backup does

shivzone: Just for reference look at the older version of backup https://github.com/greenplum-db/gpbackup/blob/347621ed3eebca2690c53e34ab14c4489c2d6626/backup/backup.go#L96 and more specifically the comment above which hinted about the ticking bomb :)

restore/validate_test.go
shivzone: yes, will remove

PR #334, opened 2020-01-24
Move toc related code and flags out of utils

lavjain:
This PR is continuation of work done in https://github.com/greenplum-db/gpbackup/pull/331 and https://github.com/greenplum-db/gpbackup/pull/332 (utils folder should not contain any business logic related to gpbackup).


shivzone:
On second thoughts i'm not a big fan of having all the packages at the top level. I prefer having bare minimal higher level packages at the top most level similar to the ones we had earlier and move packages such as toc, header etc under a folder say components/

PR #336, opened 2020-02-06
Add --redirect-schema flag to gprestore 

kyeap-pivotal:
Add --redirect-schema flag to gprestore

The --redirect-schema flag allows users to restore data into a different
schema than the original schema(s). Restrictions include...

1. The redirected schema should already be present
2. This feature is to be used in conjunction with --include-table or
   --include-table-file
3. This feature is not compatible with any exclude flags,
   --include-schema, or --include-schema-file
4. If the included data is located in multiple schemas, all data will
   redirected into the new schema
5. If the name of an index contains the schema it is under followed by a
   '.' (dot), redirect schema will fail.

### DIFF
--- a/options/options.go
+++ b/options/options.go
@@ -57,7 +57,7 @@ func NewOptions(initialFlags *pflag.FlagSet) (*Options, error) {
 
 	return &Options{
 		includedRelations:         includedRelations,
-		excludedRelations:		   excludedRelations,
+		excludedRelations:         excludedRelations,
 		includedSchemas:           includedSchemas,
 		excludedSchemas:           excludedSchemas,
 		isLeafPartitionData:       leafPartitionData,
--- a/options/options.go
+++ b/options/options.go
@@ -12,6 +12,9 @@ import (
 	"github.com/spf13/pflag"
 )
 
+// This is meant to be a read only package. Values inside should only be
+// modified by setters, it's method functions, or initialization function.
+// This package is meant to make mocking flags easier.
 type Options struct {
 	includedRelations         []string
 	excludedRelations         []string
@@ -19,6 +22,7 @@ type Options struct {
 	excludedSchemas           []string
 	includedSchemas           []string
 	originalIncludedRelations []string
+	RedirectSchema            string
 }
 
 func NewOptions(initialFlags *pflag.FlagSet) (*Options, error) {
@@ -55,6 +59,14 @@ func NewOptions(initialFlags *pflag.FlagSet) (*Options, error) {
 		return nil, err
 	}
 
+	redirectSchema := ""
+	if initialFlags.Lookup(REDIRECT_SCHEMA) != nil {
+		redirectSchema, err = initialFlags.GetString(REDIRECT_SCHEMA)
+		if err != nil {
+			return nil, err
+		}
+	}
+
 	return &Options{
 		includedRelations:         includedRelations,
 		excludedRelations:         excludedRelations,
@@ -62,6 +74,7 @@ func NewOptions(initialFlags *pflag.FlagSet) (*Options, error) {
 		excludedSchemas:           excludedSchemas,
 		isLeafPartitionData:       leafPartitionData,
 		originalIncludedRelations: includedRelations,
+		RedirectSchema:            redirectSchema,
 	}, nil
 }
 
--- a/restore/restore.go
+++ b/restore/restore.go
@@ -223,21 +223,21 @@ func restorePredata(metadataFilename string) {
 
 		existingSchemas, err := GetExistingSchemas()
 		gplog.FatalOnError(err)
-        existingTableFQNs, err := GetExistingTableFQNs()
+		existingTableFQNs, err := GetExistingTableFQNs()
 		gplog.FatalOnError(err)
 
 		existingTablesMap := make(map[string]Empty)
-		for _, table := range existingTableFQNs{
+		for _, table := range existingTableFQNs {
 			existingTablesMap[table] = Empty{}
 		}
 
-		var schemasToCreate [] string
-		var tableFQNsToCreate [] string
-		var schemasExcludedByUserInput [] string
-		var tablesExcludedByUserInput [] string
+		var schemasToCreate []string
+		var tableFQNsToCreate []string
+		var schemasExcludedByUserInput []string
+		var tablesExcludedByUserInput []string
 		for _, table := range tableFQNsToRestore {
-			schemaName := strings.Split(table,".")[0]
-			if utils.SchemaIsExcludedByUser(inSchemasUserInput, exSchemasUserInput,schemaName){
+			schemaName := strings.Split(table, ".")[0]
+			if utils.SchemaIsExcludedByUser(inSchemasUserInput, exSchemasUserInput, schemaName) {
 				if !utils.Exists(schemasExcludedByUserInput, schemaName) {
 					schemasExcludedByUserInput = append(schemasExcludedByUserInput, schemaName)
 				}
@@ -249,7 +249,7 @@ func restorePredata(metadataFilename string) {
 				if utils.RelationIsExcludedByUser(inRelationsUserInput, exRelationsUserInput, table) {
 					tablesExcludedByUserInput = append(tablesExcludedByUserInput, table)
 				} else {
-					if!utils.Exists(schemasToCreate, schemaName){
+					if !utils.Exists(schemasToCreate, schemaName) {
 						schemasToCreate = append(schemasToCreate, schemaName)
 					}
 					tableFQNsToCreate = append(tableFQNsToCreate, table)
--- a/restore/restore.go
+++ b/restore/restore.go
@@ -56,6 +56,7 @@ func SetFlagDefaults(flagSet *pflag.FlagSet) {
 	flagSet.Bool("version", false, "Print version number and exit")
 	flagSet.Bool(options.QUIET, false, "Suppress non-warning, non-error log messages")
 	flagSet.String(options.REDIRECT_DB, "", "Restore to the specified database instead of the database that was backed up")
+	flagSet.String(options.REDIRECT_SCHEMA, "", "Restore to the specified schema instead of the schema that was backed up")
 	flagSet.Bool(options.WITH_GLOBALS, false, "Restore global metadata")
 	flagSet.String(options.TIMESTAMP, "", "The timestamp to be restored, in the format YYYYMMDDHHMMSS")
 	flagSet.Bool(options.VERBOSE, false, "Print verbose log messages")
@@ -99,7 +100,8 @@ func DoSetup() {
 
 	CreateConnectionPool("postgres")
 
-	_, err := options.NewOptions(cmdFlags)
+	var err error
+	opts, err = options.NewOptions(cmdFlags)
 	gplog.FatalOnError(err)
 
 	segConfig := cluster.MustGetSegmentConfiguration(connectionPool)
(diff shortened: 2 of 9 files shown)


chrishajas:
There's some corner cases here if you want to cover those:
1) Other objects in the schema that reference objects in the same schema: `CREATE VIEW mys.v AS SELECT a.a FROM mys.foo a, mys.foo b WHERE (a.a = b.a);`
2) If the schema and table (or other objects) have the same name (this currently works I think because it only replaces the first instance in `strings.Replace(statement.Statement, oldSchema, newSchema, 1)`, but fixing the previous issue might cause this one to surface)
3) It doesn't look like statistics are covered currently

restore/restore.go
```diff
@@ -274,11 +280,15 @@ func restorePredata(metadataFilename string) {
 		inRelations = inRelationsUserInput
 		exRelations = exRelationsUserInput
 	}
-	
+
 	filters := NewFilters(inSchemas, exSchemas, inRelations, exRelations)
-	schemaStatements := GetRestoreMetadataStatementsFiltered("predata", metadataFilename, []string{"SCHEMA"}, []string{}, filters)
+	schemaStatements := []toc.StatementWithType{}
```
jimmyyih: Use an `if/else block` here instead to make it more clear and obvious that we're making `schemaStatements` an empty object when the `--redirect-schema` flag is being used.  Also, this line is extremely inefficient since the majority of the time (probably 99% of the time), it's going to be replaced in the below `if` block.

restore/restore.go
```diff
@@ -274,11 +280,15 @@ func restorePredata(metadataFilename string) {
...
+	if opts.RedirectSchema == "" {
+		schemaStatements = GetRestoreMetadataStatementsFiltered("predata", metadataFilename, []string{"SCHEMA"}, []string{}, filters)
+	}
 	statements := GetRestoreMetadataStatementsFiltered("predata", metadataFilename, []string{}, []string{"SCHEMA"}, filters)
 
+	editStatementsRedirectSchema(statements, opts.RedirectSchema)
```
jimmyyih: If you decide to go with my above suggestion to use an `if/else` block, you could move the `statements` declaration above the `schemaStatements` declaration and insert this function into the `if/else` where `opts.RedirectSchema` is set.  Then you could remove the 2nd argument to the function to make this more clean.

roicos:
> 3\. This feature is not compatible with any exclude flags

Do we validate it and do we document it somewhere?

restore/restore.go
```diff
@@ -293,6 +303,20 @@ func restorePredata(metadataFilename string) {
...
+func editStatementsRedirectSchema(statements []toc.StatementWithType, redirectSchema string) {
+	if redirectSchema == "" {
+		return
+	}
+
+	for i, statement := range statements {
+		oldSchema := fmt.Sprintf(`%s.`, statement.Schema)
+		newSchema := fmt.Sprintf(`%s.`, redirectSchema)
+		statements[i].Statement = strings.Replace(statement.Statement, oldSchema, newSchema, 1)
+		statements[i].ReferenceObject = strings.Replace(statement.ReferenceObject, oldSchema, newSchema, 1)
```
jimmyyih: I may not have this correct since I only looked for ~3 minutes but predata wouldn't have `ReferenceObject` data.  This is only attached to postdata entries such as indexes.  With this function only being called in `restorePredata`, it seems like this PR is missing logic to redirect the schema for supported postdata entries.

restore/validate.go
```diff
@@ -244,9 +253,15 @@ func ValidateFlagCombinations(flags *pflag.FlagSet) {
...
+	options.CheckExclusiveFlags(flags,
+		options.EXCLUDE_SCHEMA, options.EXCLUDE_SCHEMA_FILE, options.EXCLUDE_RELATION, options.EXCLUDE_RELATION_FILE,
+		options.INCLUDE_SCHEMA, options.INCLUDE_SCHEMA_FILE, options.INCLUDE_RELATION, options.INCLUDE_RELATION_FILE)
 	options.CheckExclusiveFlags(flags, options.METADATA_ONLY, options.DATA_ONLY)
 	options.CheckExclusiveFlags(flags, options.PLUGIN_CONFIG, options.BACKUP_DIR)
+	options.CheckExclusiveFlags(flags, options.REDIRECT_SCHEMA, options.EXCLUDE_SCHEMA, options.EXCLUDE_SCHEMA_FILE, options.EXCLUDE_RELATION, options.EXCLUDE_RELATION_FILE)
+	if flags.Changed(options.REDIRECT_SCHEMA) &&
+		!(flags.Changed(options.INCLUDE_RELATION) || flags.Changed(options.INCLUDE_RELATION_FILE) ||
+			flags.Changed(options.INCLUDE_SCHEMA) || flags.Changed(options.INCLUDE_SCHEMA_FILE)) {
+		gplog.Fatal(errors.Errorf("Cannot use redirect-schema without exactly one include filter"), "")
```
jimmyyih: Mention redirect-schema as a flag like the other error messages (preferably as `--redirect-schema flag` which is more direct).

options/options.go
```diff
@@ -12,13 +12,17 @@ import (
...
+// modified by setters, it's method functions, or initialization function.
+// This package is meant to make mocking flags easier.
 type Options struct {
 	includedRelations         []string
 	excludedRelations         []string
 	isLeafPartitionData       bool
 	excludedSchemas           []string
 	includedSchemas           []string
 	originalIncludedRelations []string
+	RedirectSchema            string
```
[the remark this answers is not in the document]
jimmyyih: It technically is camel case.  The 'R' is uppercase because Golang decided to make public/private variables and functions operate on the first character being uppercase or lowercase.  If anything, you'd be asking to make this variable private by lowercasing it and adding a getter function attached to the Options struct which sounds reasonable.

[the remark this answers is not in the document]
shivzone: i didn't know about the upper/lower case convention was related to public vs private.
Since golang isn't too OOP centric i wouldn't bang on the table for getters/setters

kyeap-pivotal:
> Do we validate [schema redirects and excludes are not used together] and do we document it somewhere? Seems like right now we only validate at least one --include flag and no --exclude flags should be provided

Yup its validated in restore/validate.go function ValidateFlagCombinations(...)

```
options.CheckExclusiveFlags(flags, options.REDIRECT_SCHEMA, options.EXCLUDE_SCHEMA, options.EXCLUDE_SCHEMA_FILE, options.EXCLUDE_RELATION, options.EXCLUDE_RELATION_FILE)
```

restore/restore.go
kyeap-pivotal: Making the RedirectSchema a explicit function argument removes hidden dependency on the variable. I would avoid reaching into a global so deep into the code, when it can be easily prevented.

restore/restore.go
kyeap-pivotal: Changed to variable declaration.

options/options.go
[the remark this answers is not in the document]
kyeap-pivotal: There seems to be little benefit (other than maybe debugging?) for making the variable private and using getters and setters here. Field kept public.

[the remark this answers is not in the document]
lavjain: I feel that redirectSchema should belong to `global_variables.go` (inside the restore module).

restore/restore.go
```diff
@@ -61,6 +61,7 @@ func SetFlagDefaults(flagSet *pflag.FlagSet) {
 	flagSet.Bool(options.VERBOSE, false, "Print verbose log messages")
 	flagSet.Bool(options.WITH_STATS, false, "Restore query plan statistics")
 	flagSet.Bool(options.LEAF_PARTITION_DATA, false, "For partition tables, create one data file per leaf partition instead of one data file for the whole table")
+	flagSet.String(options.REDIRECT_SCHEMA, "", "Restore to the specified schema instead of the schema that was backed up")
```
lavjain: Let us move the above line after` REDIRECT_DB` (line 58) to maintain sorting order

restore/restore.go
```diff
@@ -99,7 +100,8 @@ func DoSetup() {
 
 	CreateConnectionPool("postgres")
 
-	_, err := options.NewOptions(cmdFlags)
+	var err error
+	opts, err = options.NewOptions(cmdFlags)
```
lavjain: These can be collapsed:
`opts, err := options.NewOptions(cmdFlags)`

restore/restore_internal_test.go
```diff
@@ -0,0 +1,66 @@
...
+)
+
+var _ = Describe("restore internet tests", func() {
+	Describe("editStatementsRedirectStatements", func() {
+		It("does not alter schemas if no redirect was specified", func() {
+			statements := []toc.StatementWithType{
+				{ // simple table
+					Schema: "foo", Name: "bar", ObjectType: "TABLE",
+					Statement: "\n\nCREATE TABLE foo.bar (\n\ti integer\n) DISTRIBUTED BY (i);\n",
+				},
```
lavjain: We can insert some data to the table above and analyze to verify that the statistics are also restored.

kyeap-pivotal: The internal tests are purely unit tests no setup required other than the function call. I put statistics and an index on a table in the end_to_end tests.

restore/restore.go
```diff
@@ -309,11 +308,14 @@ func editStatementsRedirectSchema(statements []toc.StatementWithType, redirectSc
...
-		statements[i].Statement = strings.Replace(statement.Statement, oldSchema, newSchema, 1)
-		statements[i].ReferenceObject = strings.Replace(statement.ReferenceObject, oldSchema, newSchema, 1)
+		oldSchema := fmt.Sprintf("%s.", statement.Schema)
+		newSchema := fmt.Sprintf("%s.", redirectSchema)
 		statements[i].Schema = redirectSchema
+		statements[i].Statement = strings.Replace(statement.Statement, oldSchema, newSchema, -1)
+		// only postdata will have a reference object
+		if statement.ReferenceObject != "" {
+			statements[i].ReferenceObject = strings.Replace(statement.ReferenceObject, oldSchema, newSchema, -1)
+		}
```
jimmyyih: Why was this changed to do unlimited string replaces?  We should only be doing 1 string replace on the `Statement` object, not unlimited.  If the reason is functions or views, it was agreed that those would not be in scope at all.

jimmyyih: I would actually argue here that postdata restore operations should be skipped if we're using `--redirect-schema` for now anyways (we'll add support for it later) since most of them would require doing multiple string replaces.

kyeap-pivotal: That was pretty much the exact reason I changed it.

kyeap-pivotal: Not support --redirect-schema will not support --include-schema or --include-schema-file. No longer able to redirect schema restore functions or views.

kyeap-pivotal:
Done with PR fixes. Fixes were amended into the last commit.

restore/restore.go
```diff
@@ -293,6 +303,23 @@ func restorePredata(metadataFilename string) {
 	}
 }
 
+func editStatementsRedirectSchema(statements []toc.StatementWithType, redirectSchema string) {
+	if redirectSchema == "" {
```
jimmyyih: You've made `opts.RedirectSchema` a global object.  We shouldn't have to pass it into this function unless we're making edits too it.

restore/restore_internal_test.go
```diff
@@ -0,0 +1,58 @@
+package restore
+
+import (
+	"github.com/greenplum-db/gpbackup/toc"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("restore internet tests", func() {
```
jimmyyih: What is an internet test?

restore/restore.go
kyeap-pivotal: Yes we have the global, but imo we shouldn't access the global directly for a couple of reasons.
1. Testing is easier less confusing if we don't rely on the global. 
1. This core piece of business logic stays isolated with no possibility of external influence.

jimmyyih:
Make sure to edit the git commit message now that we're not including the `--include-schema` option.

PR #340, opened 2020-02-15
Handle backup of child partitions with different schema than the root

jimmyyih:
It is possible to have child partitions existing in different schemas
than their root partition's schema. Without logic to handle this
scenario, gpbackup was creating metadata that would put those child
partitions back onto the same root partition schema. Fix this issue by
keeping track of which child partitions have had their schema modified
and dump ALTER TABLE SET SCHEMA statements for them.

One big thing to note is that this is a bit complex when involving the
--leaf-partition-data flag and/or the schema/table filtering flags. If
a filter is given which would exclude the schema of the child
partition, we will make an exception to the filter and add the CREATE
SCHEMA statement(s) needed to guarantee table/catalog integrity for
the partition table DDL.

### DIFF
--- a/backup/backup.go
+++ b/backup/backup.go
@@ -258,7 +258,7 @@ func backupPredata(metadataFile *utils.FileWithByteCount, tables []Table, tableO
 	funcInfoMap := GetFunctionOidToInfoMap(connectionPool)
 
 	if !tableOnly {
-		BackupSchemas(metadataFile)
+		BackupSchemas(metadataFile, CreateAlteredPartitionSchemaSet(tables))
 		if len(MustGetFlagStringArray(options.INCLUDE_SCHEMA)) == 0 && connectionPool.Version.AtLeast("5") {
 			BackupExtensions(metadataFile)
 		}
--- a/backup/queries_table_defs.go
+++ b/backup/queries_table_defs.go
@@ -42,6 +42,20 @@ func (t Table) GetMetadataEntry() (string, toc.MetadataEntry) {
 		}
 }
 
+/*
+ * Extract all the unique schemas out from a Table array.
+ */
+func CreateAlteredPartitionSchemaSet(tables []Table) map[string]bool {
+	partitionAlteredSchemas := make(map[string]bool)
+	for _, table := range tables {
+		for _, alteredPartitionRelation := range table.PartitionAlteredSchemas {
+			partitionAlteredSchemas[alteredPartitionRelation.NewSchema] = true
+		}
+	}
+
+	return partitionAlteredSchemas
+}
+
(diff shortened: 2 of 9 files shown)


jimmyyih:
There may or may not be some unexplored edge cases left which I will explore on Tuesday but opening this PR now to expedite review.

@lavjain Please carefully review the SQL query I have added that searches for the partitions with altered schemas.  I didn't want to use the pg_partitions view so I constructed my own query.

backup/queries_table_defs.go
```diff
@@ -321,6 +338,46 @@ func GetPartitionDetails(connectionPool *dbconn.DBConn) (map[uint32]string, map[
...
+		quote_ident(pgn2.nspname) AS oldschema,
+		quote_ident(pgn.nspname) AS newschema,
+		quote_ident(pgc.relname) AS name
+	FROM pg_catalog.pg_partition_rule pgpr
+		LEFT JOIN pg_catalog.pg_partition pgp ON pgp.oid = pgpr.paroid
+		LEFT JOIN pg_catalog.pg_class pgc ON pgpr.parchildrelid = pgc.oid
+		LEFT JOIN pg_catalog.pg_class pgc2 ON pgp.parrelid = pgc2.oid
+		LEFT JOIN pg_catalog.pg_namespace pgn ON pgc.relnamespace = pgn.oid
+		LEFT JOIN pg_catalog.pg_namespace pgn2 ON pgc2.relnamespace = pgn2.oid
+	WHERE pgc.relnamespace != pgc2.relnamespace`)
```
lavjain: The Query looks good except that we do not need to use LEFT JOIN. Regular joins would work just fine and be more performant.

backup/backup.go
```diff
@@ -258,7 +258,7 @@ func backupPredata(metadataFile *utils.FileWithByteCount, tables []Table, tableO
 	funcInfoMap := GetFunctionOidToInfoMap(connectionPool)
 
 	if !tableOnly {
-		BackupSchemas(metadataFile)
+		BackupSchemas(metadataFile, CreateAlteredPartitionSchemaSet(tables))
```
lavjain: `CreateAlteredPartitionSchemaSet` is encroaching into several function calls. Maybe fine for now but the logic would need to be refactored.

backup/queries_table_defs.go
jimmyyih: Thanks, changed them all to regular JOIN.

backup/backup.go
jimmyyih: Fully agree.  I would like to refactor this part after the next release.  This was the cleanest approach I could think of without changing vital parts of the architecture.

PR #342, opened 2020-02-16
Update go modules to the latest stable version

lavjain:


kyeap-pivotal:
There were changes to the go.mod and the go.sum file recently. A rebase would be a good idea.

PR #348, opened 2020-02-22
Enable special character usage in filtered restore

kyeap-pivotal:
Enable special character usage in filtered restore

Enable gprestore usage of special character for the flags --include-table and
--include-table-file from the command line in the same way as
gpbackup.  Refactored code to use the options package more instead of
the MustGetFlag*(...) functions.

### DIFF
--- a/integration/options_integration_test.go
+++ b/integration/options_integration_test.go
@@ -13,61 +13,30 @@ import (
 
 var _ = Describe("Options Integration", func() {
 	Describe("QuoteTablesNames", func() {
-		It("returns unchanged when fqn has no special characters", func() {
+		It("quotes identifiers as expected", func() {
 			tableList := []string{
-				`foo.bar`,
-			}
-
-			resultFQNs, err := options.QuoteTableNames(connectionPool, tableList)
-			Expect(err).ToNot(HaveOccurred())
-			Expect(tableList).To(Equal(resultFQNs))
-		})
-		It("adds quote to single fqn when fqn has special characters", func() {
-			tableList := []string{
-				`FOO.bar`,
-			}
-			expected := []string{
-				`"FOO".bar`,
-			}
-
-			resultFQNs, err := options.QuoteTableNames(connectionPool, tableList)
-			Expect(err).ToNot(HaveOccurred())
-			Expect(expected).To(Equal(resultFQNs))
-		})
-		It("adds quotes as necessary to multiple entries", func() {
-			tableList := []string{
-				`foo.BAR`,
-				`FOO.BAR`,
-				`bim.2`,
-				`public.foo_bar`,
+				`foo.bar`,  // no special characters
+				`foo.BAR`,  // capital characters
+				`foo.'bar`, // make sure that single quotes are escaped before string is fed to quote_ident
+				`foo.2`,    // numbers
+				`foo._bar`, // underscore
 				`foo ~#$%^&*()_-+[]{}><\|;:/?!,.bar`,
-				"tab\t.bar", // important to use double quotes to allow \t to become tab
-				"tab\n.bar", // important to use double quotes to allow \t to become tab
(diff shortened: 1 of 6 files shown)


options/options.go
```diff
@@ -141,18 +141,25 @@ type FqnStruct struct {
...
+		escapedTables = append(escapedTables, utils.EscapeSingleQuotes(v))
+	}
+
+	fqnSlice, err := SeparateSchemaAndTable(escapedTables)
 	if err != nil {
 		return nil, err
 	}
 	result := make([]string, 0)
 
+	const QuoteIdent = `SELECT quote_ident('%s') AS schemaname, quote_ident('%s') AS tablename`
```
jimmyyih: Why is a `const` needed here?  Just wondering.

Makefile
```diff
@@ -64,9 +65,14 @@ coverage :
...
 
 build :
-		$(GO_ENV) go build -tags '$(BACKUP)' $(GOFLAGS) -o $(BIN_DIR)/$(BACKUP) -ldflags $(BACKUP_VERSION_STR)
-		$(GO_ENV) go build -tags '$(RESTORE)' $(GOFLAGS) -o $(BIN_DIR)/$(RESTORE) -ldflags $(RESTORE_VERSION_STR)
-		$(GO_ENV) go build -tags '$(HELPER)' $(GOFLAGS) -o $(BIN_DIR)/$(HELPER) -ldflags $(HELPER_VERSION_STR)
+		$(GO_ENV) go build -tags '$(BACKUP)' -o $(BIN_DIR)/$(BACKUP) -ldflags $(BACKUP_VERSION_STR)
+		$(GO_ENV) go build -tags '$(RESTORE)' -o $(BIN_DIR)/$(RESTORE) -ldflags $(RESTORE_VERSION_STR)
+		$(GO_ENV) go build -tags '$(HELPER)' -o $(BIN_DIR)/$(HELPER) -ldflags $(HELPER_VERSION_STR)
+
+debug :
```
jimmyyih: Thank you.  I've been thinking of doing this for a while now but I kept putting it off.

integration/options_integration_test.go
```diff
@@ -13,61 +13,30 @@ import (
...
-			Expect(err).ToNot(HaveOccurred())
-			Expect(expected).To(Equal(resultFQNs))
-		})
-		It("adds quotes as necessary to multiple entries", func() {
-			tableList := []string{
-				`foo.BAR`,
-				`FOO.BAR`,
-				`bim.2`,
-				`public.foo_bar`,
+				`foo.bar`,  // no special characters
```
jimmyyih: Should we add one more entry here where there is a `.` in the schema?

kyeap-pivotal: Support for filtered table gpbackups where schemas or tables where there is a `.` in the fqn is disabled in gpbackup.

We can write a test in to verify it will exit if there is a `.` in the schema or table.

options/options.go
kyeap-pivotal: It used to be declared in package namespace and shouldn’t be modified. I didn’t see a reason for it since this was this was only function that was using it so I moved it into the function.

jimmyyih: Let's rename the variable then to something like `quoteIdentTableFQNQuery` (just a normal query string so can lose out on the `const` declaration too) so it doesn't get confused with `utils.QuoteIdent()`.

As far as it being used as a public package variable, I would only think plugins or gpcopy could be possible consumers.  I highly doubt any plugins use them and I just checked gpcopy which doesn't use it.  We're safe to move it into the function.

Makefile
kyeap-pivotal: Cherry-picked and merged into master.

integration/options_integration_test.go
kyeap-pivotal: There is unit testing of the options package for failure if there is a `.` in the schema of table.

PR #352, opened 2020-03-07
Refactor sequences, text search, operators, types and other objects to use CreateStatements

lavjain:


backup/dependencies_test.go
```diff
@@ -132,7 +132,8 @@ var _ = Describe("backup/dependencies tests", func() {
 			constraints := []backup.Constraint{
 				{Name: "check_constraint", ConDef: "CHECK (VALUE > 2)", OwningObject: "public.domain"},
 			}
-			backup.PrintDependentObjectStatements(backupfile, tocfile, objects, metadataMap, constraints, funcInfoMap)
+			backup.SetConstaints(constraints)
+			backup.PrintDependentObjectStatements(backupfile, tocfile, objects, metadataMap, funcInfoMap)
```
kyeap-pivotal: If this is the reason for the new `SetConstraints` function, it would be better just to pass it in. The reason for it would be the same as why metadataMap is passed in although its a global. This is a fairly important function that contains the business logic to print SQL statements. By passing in `constraints` in we can ensure changes to the global will not affect the function as it is running, because the function would get a local copy. Easier to unit test. Easier to debug if something goes wrong.

lavjain:
Holding off on this PR for now. Will pull out the concepts into smaller chunks going forward.

PR #356, opened 2020-03-20
Add --without-globals to gpbackup

kyeap-pivotal:
Add --without-globals to gpbackup

Currently we always backup globals on full database backups.
This flag gives users the options to disable backup of globals when
backing up their database.

Co-authored-by: Adam Berlin <aberlin@pivotal.io>

### DIFF
--- a/backup/backup.go
+++ b/backup/backup.go
@@ -125,12 +125,13 @@ func DoBackup() {
 
 	backupSessionGUC(metadataFile)
 	if !MustGetFlagBool(options.DATA_ONLY) {
-		tableOnlyBackup := true
-		if len(MustGetFlagStringArray(options.INCLUDE_RELATION)) == 0 {
-			tableOnlyBackup = false
-			backupGlobal(metadataFile)
+		isFullBackup := len(MustGetFlagStringArray(options.INCLUDE_RELATION)) == 0
+		if isFullBackup && !MustGetFlagBool(options.WITHOUT_GLOBALS) {
+			backupGlobals(metadataFile)
 		}
-		backupPredata(metadataFile, metadataTables, tableOnlyBackup)
+
+		isFilteredBackup := !isFullBackup
+		backupPredata(metadataFile, metadataTables, isFilteredBackup)
 		backupPostdata(metadataFile)
 	}
 
@@ -178,7 +179,7 @@ func DoBackup() {
 	gplog.FatalOnError(err)
 }
 
-func backupGlobal(metadataFile *utils.FileWithByteCount) {
+func backupGlobals(metadataFile *utils.FileWithByteCount) {
 	gplog.Info("Writing global database metadata")
 
 	backupResourceQueues(metadataFile)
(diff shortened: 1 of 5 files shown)


backup/backup.go
```diff
@@ -165,14 +166,14 @@ func DoBackup() {
...
-		if len(MustGetFlagStringArray(options.INCLUDE_RELATION)) == 0 {
-			tableOnlyBackup = false
-			backupGlobal(metadataFile)
+		isFullBackup := len(MustGetFlagStringArray(options.INCLUDE_RELATION)) == 0
+		if isFullBackup && !MustGetFlagBool(options.WITHOUT_GLOBALS) {
+			backupGlobals(metadataFile)
 		}
-		backupPredata(metadataFile, metadataTables, tableOnlyBackup)
+
+		isFilteredBackup := len(MustGetFlagStringArray(options.INCLUDE_RELATION)) > 0
```
lavjain: Does `options.EXCLUDE_RELATION` makes it filtered?

lavjain: Ideally, `isFullBackup := !isFilteredBackup`

backup/backup.go
```diff
@@ -165,14 +166,14 @@ func DoBackup() {
...
 	metadataFile := utils.NewFileWithByteCountFromFile(metadataFilename)
 
-	backupSessionGUC(metadataFile)
 	if !MustGetFlagBool(options.DATA_ONLY) {
-		tableOnlyBackup := true
-		if len(MustGetFlagStringArray(options.INCLUDE_RELATION)) == 0 {
-			tableOnlyBackup = false
-			backupGlobal(metadataFile)
+		isFullBackup := len(MustGetFlagStringArray(options.INCLUDE_RELATION)) == 0
+		if isFullBackup && !MustGetFlagBool(options.WITHOUT_GLOBALS) {
```
shivzone: Some how `if (!(WITH_GLOBALS))` doesn't seem as intuitive as `if (WITH_GLOBALS)`

options/flag.go
```diff
@@ -44,6 +44,7 @@ const (
 	TIMESTAMP             = "timestamp"
 	WITH_GLOBALS          = "with-globals"
 	REDIRECT_SCHEMA       = "redirect-schema"
+	WITHOUT_GLOBALS       = "without-globals"
```
shivzone: We can simply reuse with_globals instead of introducing a new flag

kyeap-pivotal: with-globals is a flag that only exists in gprestore. We can reuse the flag internally. But I would rather keep the naming consistent.

shivzone: yes i was suggesting using the flag with-globals for gpbackup as well . I didn't get you "keep the naming consistent" ?

shivzone: @kyeap-pivotal what's the resolution on this ?

backup/backup.go
kyeap-pivotal: I also thought about the same thing, but when I think about from a user standpoint it doesn't seem intuitive to exclude all globals if a user chose to exclude a single table.

PR #358, opened 2020-03-20
gprestore truncate flag for included tables

shivzone:
Added a `--truncate` flag for the purpose of truncating tables
provided in the `--include-table` or `--include-table-file` prior to
restoring data. This works only with gprestore with `--data-only` mode and
`--incremental-restore` mode is not supported.

We don't expect `include-table` specifically in the scenario of restoring with truncate. Truncating serially should be fine.

### DIFF
--- a/end_to_end/end_to_end_suite_test.go
+++ b/end_to_end/end_to_end_suite_test.go
@@ -203,9 +203,9 @@ func forceMetadataFileDownloadFromPlugin(conn *dbconn.DBConn, timestamp string) 
 	fpInfo := filepath.NewFilePathInfo(backupCluster, "", timestamp, filepath.GetSegPrefix(conn))
 	remoteOutput := backupCluster.GenerateAndExecuteCommand(
 		fmt.Sprintf("Removing backups on all segments for "+
-		"timestamp %s", timestamp), func(contentID int) string {
-		return fmt.Sprintf("rm -rf %s", fpInfo.GetDirForContent(contentID))
-	}, cluster.ON_SEGMENTS_AND_MASTER)
+			"timestamp %s", timestamp), func(contentID int) string {
+			return fmt.Sprintf("rm -rf %s", fpInfo.GetDirForContent(contentID))
+		}, cluster.ON_SEGMENTS_AND_MASTER)
 	if remoteOutput.NumErrors != 0 {
 		Fail(fmt.Sprintf("Failed to remove backup directory for timestamp %s", timestamp))
 	}
@@ -329,8 +329,8 @@ var _ = Describe("backup end to end integration tests", func() {
 		} else {
 			remoteOutput := backupCluster.GenerateAndExecuteCommand(
 				"Creating filespace test directories on all hosts", func(contentID int) string {
-				return fmt.Sprintf("mkdir -p /tmp/test_dir && mkdir -p /tmp/test_dir1 && mkdir -p /tmp/test_dir2")
-			}, cluster.ON_HOSTS_AND_MASTER)
+					return fmt.Sprintf("mkdir -p /tmp/test_dir && mkdir -p /tmp/test_dir1 && mkdir -p /tmp/test_dir2")
+				}, cluster.ON_HOSTS_AND_MASTER)
 			if remoteOutput.NumErrors != 0 {
 				Fail("Could not create filespace test directory on 1 or more hosts")
 			}
@@ -360,8 +360,8 @@ var _ = Describe("backup end to end integration tests", func() {
 				"-c", "DROP TABLESPACE test_tablespace").Run()
 			remoteOutput := backupCluster.GenerateAndExecuteCommand(
 				"Removing /tmp/test_dir* directories on all hosts", func(contentID int) string {
-				return fmt.Sprintf("rm -rf /tmp/test_dir*")
-			}, cluster.ON_HOSTS_AND_MASTER)
+					return fmt.Sprintf("rm -rf /tmp/test_dir*")
+				}, cluster.ON_HOSTS_AND_MASTER)
 			if remoteOutput.NumErrors != 0 {
 				Fail("Could not remove /tmp/testdir* directories on 1 or more hosts")
 			}
@@ -584,7 +584,7 @@ var _ = Describe("backup end to end integration tests", func() {
 				assertRelationsCreated(restoreConn, TOTAL_RELATIONS_AFTER_EXCLUDE)
 				assertDataRestored(restoreConn, map[string]int{
 					"schema2.foo3": 100,
-					"public.foo": 40000,
+					"public.foo":   40000,
 					"public.holds": 50000,
 					"public.sales": 13})
 			})
@@ -601,7 +601,7 @@ var _ = Describe("backup end to end integration tests", func() {
 				assertRelationsCreated(restoreConn, 8)
 				assertDataRestored(restoreConn, map[string]int{
 					"schema2.foo3": 100,
-					"public.foo": 40000,
+					"public.foo":   40000,
 					"public.holds": 50000})
 
 				_ = os.Remove("/tmp/exclude-tables.txt")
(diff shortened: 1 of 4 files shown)


restore/restore.go
```diff
@@ -363,10 +364,11 @@ func restoreData() {
 
 	gucStatements := setGUCsForConnection(nil, 0)
 	for timestamp, entries := range filteredDataEntries {
-		gplog.Verbose("Restoring data from backup with timestamp: %s", timestamp)
-		if MustGetFlagBool(options.INCREMENTAL) {
+		if MustGetFlagBool(options.INCREMENTAL) || MustGetFlagBool(options.TRUNCATE) {
+			gplog.Verbose("Truncating tables prior to restoring backup with timestamp: %s", timestamp)
```
lavjain: Looks like we are truncating all tables at once instead of one by one as discussed.

shivzone: Truncate is a rather expensive operation and it wouldn't make sense for us to run this alongside the data restore `goroutines`
Based on discussion with @oakbarrett the `truncate` feature is requested only for users who want to restore a very finite set of tables so we felt treating it the same way as we do `incremental restore` makes sense.

restore/validate.go
```diff
@@ -258,8 +258,18 @@ func ValidateFlagCombinations(flags *pflag.FlagSet) {
...
+	options.CheckExclusiveFlags(flags,
+		options.TRUNCATE, options.REDIRECT_SCHEMA, options.EXCLUDE_SCHEMA, options.EXCLUDE_SCHEMA_FILE,
+		options.EXCLUDE_RELATION, options.EXCLUDE_RELATION_FILE, options.INCLUDE_SCHEMA, options.INCLUDE_SCHEMA_FILE)
+	if flags.Changed(options.REDIRECT_SCHEMA) &&
+		!(flags.Changed(options.INCLUDE_RELATION) || flags.Changed(options.INCLUDE_RELATION_FILE)) {
 		gplog.Fatal(errors.Errorf("Cannot use --redirect-schema without --include-table or --include-table-file"), "")
 	}
+	options.CheckExclusiveFlags(flags,
+		options.TRUNCATE, options.METADATA_ONLY, options.INCREMENTAL, options.REDIRECT_SCHEMA)
+	if flags.Changed(options.TRUNCATE) &&
```
lavjain: The validations here need looks complex and needs simplification.

shivzone: As per the feature spec
`--truncate` flag shouldn't function with `--exclude`, `--include-schema` `--metadata-only` and `--incremental` flags

`--truncate` flag requires `--include-table` or `--include-table-file` flags and can run only with `--data-only` restore

The above validation reflects that

shivzone: If your suggestion is to re-haul our entire flag Validation code to make it easier for us to add rules we can handle that as a separate story

lavjain: The above validation also says that `--truncate` cannot be used with `--redirect-schema` which does not sound correct.

shivzone: That was the specified in the feature request. @oakbarrett can you please confirm that we don't need to support `--redirect-schema` with the new `--truncate` flag

end_to_end/end_to_end_suite_test.go
```diff
@@ -1513,6 +1513,158 @@ var _ = Describe("backup end to end integration tests", func() {
...
+				assertDataRestored(restoreConn, map[string]int{
+					"public.sales": 13, "public.foo": 40000})
+
+				gprestore(gprestorePath, restoreHelperPath, timestamp,
+					"--redirect-db", "restoredb",
+					"--include-table", "public.foo",
+					"--include-table", "public.sales",
+					"--data-only")
+				assertDataRestored(restoreConn, map[string]int{
+					"public.sales": 26, "public.foo": 80000})
```
jimmyyih: What is this section testing?  Seems unnecessary since it's not testing the truncate feature.

end_to_end/end_to_end_suite_test.go
```diff
@@ -1513,6 +1513,158 @@ var _ = Describe("backup end to end integration tests", func() {
...
+				assertDataRestored(restoreConn, map[string]int{
+					"public.sales": 13})
+
+				gprestore(gprestorePath, restoreHelperPath, timestamp,
+					"--redirect-db", "restoredb",
+					"--backup-dir", backupDir,
+					"--include-table-file", "/tmp/include-tables.txt",
+					"--data-only")
+				assertDataRestored(restoreConn, map[string]int{
+					"public.sales": 26})
```
jimmyyih: Same here, this doesn't look useful.

end_to_end/end_to_end_suite_test.go
```diff
@@ -1513,6 +1513,158 @@ var _ = Describe("backup end to end integration tests", func() {
...
+					"--truncate", "--data-only")
+				assertDataRestored(restoreConn, map[string]int{
+					"public.sales": 1, "public.sales_1_prt_jan17": 1})
+
+				gprestore(gprestorePath, restoreHelperPath, timestamp,
+					"--redirect-db", "restoredb",
+					"--include-table", "public.sales_1_prt_jan17",
+					"--data-only")
+				assertDataRestored(restoreConn, map[string]int{
+					"public.sales": 2, "public.sales_1_prt_jan17": 2})
```
jimmyyih: And here.

restore/validate.go
jimmyyih: Looks fine to me but I do heavily agree that this part of the code looks very confusing.

shivzone: I agree, but that is true for flag validation in general or are you guys specifically finding the `--truncate` flag validation complicated ? If so, any suggestions to simplify this ? Maybe split the checks into more specific ones ?
@kyeap-pivotal is looking at refactoring validation, maybe we can look at simplifying flag validation as part of that effort.

jimmyyih: In general.

end_to_end/end_to_end_suite_test.go
shivzone: It is primarily to make sure there is no regression in behavior without --truncate, but I get your point since there are other tests for that. What would make more sense is to restore followed by `data-only` restore to double the table data and finally restore with truncate to bring it back to the original size. This way we are certain that restore with truncate actually does something as opposed to the current test where even if restore with truncate was a no-op the test would pass.

end_to_end/end_to_end_suite_test.go
shivzone: Updated

end_to_end/end_to_end_suite_test.go
shivzone: Changed

end_to_end/end_to_end_suite_test.go
```diff
@@ -1513,6 +1513,159 @@ var _ = Describe("backup end to end integration tests", func() {
...
+					"INSERT into sales values(1, '2017-01-01', 109.99)")
+
+				gprestore(gprestorePath, restoreHelperPath, timestamp,
+					"--redirect-db", "restoredb",
+					"--include-table", "public.sales",
+					"--truncate", "--data-only")
+				assertDataRestored(restoreConn, map[string]int{
+					"public.sales": 13})
+			})
+			FIt("Repeat runs gpbackup and gprestore with truncate and include-table flags", func() {
```
jimmyyih: Need to get rid of the focus.

lavjain:
@shivzone Your dev pipeline is still red

restore/validate.go
lavjain: According to the tracker story, `--redirect-schema` must be supported with `--truncate` flag

shivzone:
> @shivzone Your dev pipeline is still red

@lavjain  Resolved. Issues were related to 1. using the wrong db connection 2. an issue that uncovered a an issue with our naming convention of our  with a restore report file. The tests have been fixed to deal with this limitation

end_to_end/end_to_end_suite_test.go
shivzone: done

restore/validate.go
shivzone: Nope as per the story `--redirect-schema` is not supported. Only `--redirect-db` is supported

PR #362, opened 2020-03-28
Improve readability for backupPredata by moving version checks into specific underlying functions

lavjain:
No logic changes


shivzone:
Please add in the description of this story (for future reference) that this PR is primarily about moving the gpdb version specific checks away from `backupGlobals` and `backupPredata` into underlying specific functions.
Also, i think we should retain the log upon termination

lavjain:
@shivzone Thanks. I am not clear about your comment on retaining the log

shivzone:
@lavjain ignore the comment i removed it

PR #370, opened 2020-04-08
Split end to end tests by common test cases

kyeap-pivotal:
End to end tests have been split into the following files. Tests have
been divided into the following categories.
- Filtered  (include/excludes)
- Incremental
- Plugin
- Special characters
- Everything else (original file)

A great deal of tests and functions were rearranged and moved. The git changes might not be too useful to read. Main things to look out for is if we agree with the types of "buckets" laid out for tests to fall into.


lavjain:
This is an excellent first step. Please also move all sql scripts to `sql` directory or to the `resources` directory.

PR #371, opened 2020-04-09
Do not allow go to implicitly update go.mod

kyeap-pivotal:
- Enable `-mod=readonly` build flag. When this flag is not specified and
  go encounters a package not listed as a dependency during compile time,
  the default behavior is to find and use the latest version of the package
  available from the internet and automatically update go.mod to include
  the missing dependency. Enabling this build flag is useful during
  automatic testing for detecting packages not explicitly stated as a
  dependency during compile time.
- Update makefile rule `clean`. Will help clean up executables and caching
  creates by go when compiling a project
- Ran `go mod tidy` to clean up mod files


jimmyyih:
What does `implicit automatic updating of go.mod` mean?

Makefile
```diff
@@ -71,19 +71,19 @@ coverage :
 		@./show_coverage.sh
 
 build :
-		$(GO_ENV) go build -tags '$(BACKUP)' -o $(BIN_DIR)/$(BACKUP) -ldflags $(BACKUP_VERSION_STR)
-		$(GO_ENV) go build -tags '$(RESTORE)' -o $(BIN_DIR)/$(RESTORE) -ldflags $(RESTORE_VERSION_STR)
-		$(GO_ENV) go build -tags '$(HELPER)' -o $(BIN_DIR)/$(HELPER) -ldflags $(HELPER_VERSION_STR)
+		$(GO_ENV) go build -mod=readonly -tags '$(BACKUP)' -o $(BIN_DIR)/$(BACKUP) -ldflags $(BACKUP_VERSION_STR)
```
lavjain, line 74: Set `GOFLAGS=$(GOFLAGS) -mod=readonly` and use GOFLAGS instead of modifying each line

go.mod
```diff
@@ -6,9 +6,12 @@ require (
 	github.com/DATA-DOG/go-sqlmock v1.4.1
 	github.com/blang/semver v3.5.1+incompatible
 	github.com/blang/vfs v1.0.0
+	github.com/fatih/color v1.9.0 // indirect
+	github.com/gofrs/uuid v3.2.0+incompatible // indirect
```
lavjain, line 10: The above lines are not required on macOS

go.mod
```diff
@@ -6,9 +6,12 @@ require (
...
 	github.com/greenplum-db/gp-common-go-libs v0.0.0-20200220000734-974a164cfd33
 	github.com/jackc/pgx v3.6.2+incompatible
 	github.com/lib/pq v1.3.0
+	github.com/mattn/go-colorable v0.1.6 // indirect
```
lavjain, line 14: same here

kyeap-pivotal:
> What does `implicit automatic updating of go.mod` mean?

By default, go has permission to update go.mod when building a project if it is missing dependencies. This can result in a scenario where a dev pushes changes that use a new package, but forgets to update go.mod. 

Right now if that happens, the automatic testing will happily pass because go will automatically find the dependency it needs and modify the go.mod file. With this change the automatic tests will fail because we do not allow the behavior.

go.mod
kyeap-pivotal: Can you explain a bit on why theses statements aren't needed for macOS? These we automatically added by `go mod tidy`.

shivzone:
So the expectation is from now on we manually update go.mod for any new dependency ?

go.mod
lavjain: Please ignore my comment as I just ran `make build` and `make unit` (not `go mod tidy`)

Makefile
```diff
@@ -10,6 +10,7 @@ RESTORE=gprestore
 HELPER=gpbackup_helper
 BIN_DIR=$(shell echo $${GOPATH:-~/go} | awk -F':' '{ print $$1 "/bin"}')
 GINKGO_FLAGS := -r -keepGoing -randomizeSuites -randomizeAllSpecs -noisySkippings=false
+GOFLAGS:=$(GOFLAGS) -mod=readonly
```
shivzone, line 13: Since you are setting `GOFLAGS` plesase set it with `-mod=readonly -tags -ldflags` and remove them from the build aciton

Makefile
```diff
@@ -71,19 +72,19 @@ coverage :
...
-		$(GO_ENV) go build -tags '$(HELPER)' $(DEBUG) -o $(BIN_DIR)/$(HELPER) -ldflags $(HELPER_VERSION_STR)
+		$(GO_ENV) go build $(GOFLAGS) -tags '$(BACKUP)' $(DEBUG) -o $(BIN_DIR)/$(BACKUP) -ldflags $(BACKUP_VERSION_STR)
+		$(GO_ENV) go build $(GOFLAGS) -tags '$(RESTORE)' $(DEBUG) -o $(BIN_DIR)/$(RESTORE) -ldflags $(RESTORE_VERSION_STR)
+		$(GO_ENV) go build $(GOFLAGS) -tags '$(HELPER)' $(DEBUG) -o $(BIN_DIR)/$(HELPER) -ldflags $(HELPER_VERSION_STR)
 
 build_linux :
-		env GOOS=linux GOARCH=amd64 go build -tags '$(BACKUP)' $(GOFLAGS) -o $(BACKUP) -ldflags $(BACKUP_VERSION_STR)
-		env GOOS=linux GOARCH=amd64 go build -tags '$(RESTORE)' $(GOFLAGS) -o $(RESTORE) -ldflags $(RESTORE_VERSION_STR)
-		env GOOS=linux GOARCH=amd64 go build -tags '$(HELPER)' $(GOFLAGS) -o $(HELPER) -ldflags $(HELPER_VERSION_STR)
+		env GOOS=linux GOARCH=amd64 go build $(GOFLAGS) -tags '$(BACKUP)' $(GOFLAGS) -o $(BACKUP) -ldflags $(BACKUP_VERSION_STR)
```
shivzone, line 85: `GOFLAGS` dupcliate here and the next 2

Makefile
kyeap-pivotal: Oh I see GOFLAGS is a golang flag. It wasn't my intention to use the same flag let me change the variable name. Also the parameters for -ldflag is different with each build, can't be consolidated. I'll break it out to make it a little more clear.

kyeap-pivotal:
> So the expectation is from now on we manually update go.mod for any new dependency ?

The expectation is not to let go find a dependency for us during build time.`-mod=readonly` will restrict that. Packages brought in through `go get` (which is how we should be getting packages anyway) will still automatically update the go.mod file.

kyeap-pivotal:
Updated commit message to more clear on what `implicit automatic updating of go.mod` meant.

lavjain:
Double quotes may not be required with `"-X $(BACKUP_VERSION_STR)"`

PR #372, opened 2020-04-09
Sets gp_autostats_mode to none when restore data

roicos:
- By default gp_autostats_mode GUC is set to
   on_no_stats, that means that ANALYSE will
   run every time we insert data. It can slow
   down restore, so we need to set it to none
   before restore data to desable ANALYZE while
   restoring
 - Turns out that autostat covered --redirect-schema
   bug and we had false-positive on --redirect-schema
   end_to_end test while actual statistics was
   not redirected. It happened because of two things:
   1. we had filtered all the statements for restore
   statistics when we had --include-table filter.
   This is fixed in the toc func shouldIncludeStatement.
   2. the way we modify statements for redirect is
   a Strings.Replace and we only replaced the first match
   while in the statistics statement we have several queries.
   The way we modified it only replaced schema for one query, 
   but because insert query never run we did not see it fails.
   Currently replacement is changed to global. 
   Another story could be created for farther improvements.

### DIFF
--- a/end_to_end/end_to_end_suite_test.go
+++ b/end_to_end/end_to_end_suite_test.go
@@ -138,7 +138,7 @@ func buildOldBinaries(version string) (string, string) {
 
 func assertDataRestored(conn *dbconn.DBConn, tableToTupleCount map[string]int) {
 	for tableName, expectedNumTuples := range tableToTupleCount {
-		actualTupleCount := dbconn.MustSelectString(conn, fmt.Sprintf("SELECT count(*) AS string from %s", tableName))
+		actualTupleCount := dbconn.MustSelectString(conn, fmt.Sprintf("SELECT count(*) AS string FROM %s", tableName))
 		if strconv.Itoa(expectedNumTuples) != actualTupleCount {
 			Fail(fmt.Sprintf("Expected:\n\t%s rows to have been restored into table %s\nActual:\n\t%s rows were restored", strconv.Itoa(expectedNumTuples), tableName, actualTupleCount))
 		}
@@ -712,6 +712,7 @@ var _ = Describe("backup and restore end to end tests", func() {
 	})
 	Describe("Redirect Schema", func() {
 		It("runs gprestore with --redirect-schema to redirect data back to the original database which still contain the original tables", func() {
+			Skip("TODO: fix restore statistics on redirect")
 			skipIfOldBackupVersionBefore("1.17.0")
 			testhelper.AssertQueryRuns(backupConn,
 				"DROP SCHEMA IF EXISTS schema3 CASCADE; CREATE SCHEMA schema3;")
@@ -740,7 +741,7 @@ var _ = Describe("backup and restore end to end tests", func() {
 			Expect(actualIndexCount).To(Equal("1"))
 
 			actualStatisticCount := dbconn.MustSelectString(backupConn,
-				`SELECT count(*) AS string FROM pg_stats WHERE schemaname='schema3' AND tablename='foo3';`)
+				`SELECT count(*) FROM pg_statistic WHERE starelid='schema3.foo3'::regclass::oid;`)
 			Expect(actualStatisticCount).To(Equal("1"))
 		})
 		It("runs gprestore with --redirect-schema and multiple included schemas", func() {
@@ -1018,10 +1019,22 @@ var _ = Describe("backup and restore end to end tests", func() {
 		Expect(string(output)).To(ContainSubstring("Query planner statistics restore complete"))
 		assertDataRestored(restoreConn, publicSchemaTupleCounts)
 		assertDataRestored(restoreConn, schema2TupleCounts)
+
+		backupStatisticCount := dbconn.MustSelectString(backupConn,
+			`SELECT count(*) AS string FROM pg_statistic;`)
+		restoredStatisticsCount := dbconn.MustSelectString(restoreConn,
+			`SELECT count(*) AS string FROM pg_statistic;`)
+		Expect(backupStatisticCount).To(Equal(restoredStatisticsCount))
+
+		restoredTablesAnalyzed := dbconn.MustSelectString(restoreConn,
+			`SELECT count(*) FROM pg_stat_last_operation WHERE objid IN ('public.foo'::regclass::oid, 'public.holds'::regclass::oid, 'public.sales'::regclass::oid, 'schema2.returns'::regclass::oid, 'schema2.foo2'::regclass::oid, 'schema2.foo3'::regclass::oid, 'schema2.ao1'::regclass::oid, 'schema2.ao2'::regclass::oid) AND staactionname='ANALYZE';`)
+		Expect(restoredTablesAnalyzed).To(Equal("0"))
 	})
 	It("restores statistics only for tables specified in --include-table flag when runs gprestore with with-stats flag", func() {
 		testhelper.AssertQueryRuns(backupConn,
 			"CREATE TABLE public.table_to_include_with_stats(i int)")
+		testhelper.AssertQueryRuns(backupConn,
+			"INSERT INTO public.table_to_include_with_stats SELECT generate_series(0,9);")
 		defer testhelper.AssertQueryRuns(backupConn,
 			"DROP TABLE public.table_to_include_with_stats")
 		timestamp := gpbackup(gpbackupPath, backupHelperPath,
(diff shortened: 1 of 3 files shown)


restore/restore.go
```diff
@@ -282,7 +282,7 @@ func editStatementsRedirectSchema(statements []toc.StatementWithType, redirectSc
 		oldSchema := fmt.Sprintf("%s.", statement.Schema)
 		newSchema := fmt.Sprintf("%s.", redirectSchema)
 		statements[i].Schema = redirectSchema
-		statements[i].Statement = strings.Replace(statement.Statement, oldSchema, newSchema, 1)
+		statements[i].Statement = strings.Replace(statement.Statement, oldSchema, newSchema, -1)
```
jimmyyih, line 285: Although this works in the most basic (and probably majority) of cases, this change introduces too many edge case errors.

A very likely edge case for example could involve user-defined types in the same schema:
```
CREATE SCHEMA fooschema;
CREATE TYPE fooschema.footype AS ENUM ('a', 'b', 'c');
CREATE TABLE fooschema.footable (a fooschema.footype, b int);
```
When using `--redirect-schema`, it should only change `fooschema.footable` to `newschema.footable` and not touch anything else.

Going through the scenario of restoring statistics with `--redirect-schema`, the SQL statements are all combined into one big string.  **Would a better solution be to split that one big statistics string into individual SQL statements?**

Another thing that I noticed while going through the code execution is an incorrect SQL statement in the statistics file:
```
UPDATE pg_class
SET
        relpages = 1::int,
        reltuples = 10.000000::real
WHERE relname = 'footable'
AND relnamespace = 16425;
```
The `relnamespace` cannot be assumed to be a snapshot of the namespace's Oid at that moment in time.  We'll have to create a bug story for that one.

lavjain:
We should only replace the first occurrence during token replacement and split import statistics into multiple statements.

end_to_end/end_to_end_suite_test.go
```diff
@@ -740,7 +741,7 @@ var _ = Describe("backup and restore end to end tests", func() {
 			Expect(actualIndexCount).To(Equal("1"))
 
 			actualStatisticCount := dbconn.MustSelectString(backupConn,
-				`SELECT count(*) AS string FROM pg_stats WHERE schemaname='schema3' AND tablename='foo3';`)
+				`Select count(*) from pg_statistic WHERE starelid='schema3.foo3'::regclass::oid;`)
```
jimmyyih, line 744: CAPS all the SQL keywords.  For this one, upper `Select` and `from`.

end_to_end/end_to_end_suite_test.go
```diff
@@ -1018,10 +1019,42 @@ var _ = Describe("backup and restore end to end tests", func() {
...
 		assertDataRestored(restoreConn, publicSchemaTupleCounts)
 		assertDataRestored(restoreConn, schema2TupleCounts)
+
+		backupStatisticCount := dbconn.MustSelectString(backupConn,
+			`SELECT count(*) AS string FROM pg_statistic;`)
+		restoredStatisticsCount := dbconn.MustSelectString(restoreConn,
+			`SELECT count(*) AS string FROM pg_statistic;`)
+		Expect(backupStatisticCount).To(Equal(restoredStatisticsCount))
+	})
+	It("disables autostat when restore --with-stats", func(){
```
jimmyyih, line 1029: Instead of creating a new test, could you just add on to the stats test (`It("runs gpbackup and gprestore with with-stats flag"`) above?  Disabling autostats is a subset of --with-stats feature so no need to create a new test specifically for it.  If you're wondering on which test tables to use, these are the ones available:
```
publicSchemaTupleCounts = map[string]int{
        "public.foo":   40000,
        "public.holds": 50000,
        "public.sales": 13,
}
schema2TupleCounts = map[string]int{
        "schema2.returns": 6,
        "schema2.foo2":    0,
        "schema2.foo3":    100,
        "schema2.ao1":     1000,
        "schema2.ao2":     1000,
}
```

end_to_end/end_to_end_suite_test.go
```diff
@@ -1018,10 +1019,42 @@ var _ = Describe("backup and restore end to end tests", func() {
...
+			"--with-stats",
+			"--backup-dir", backupDir)
+
+		output := gprestore(gprestorePath, restoreHelperPath, timestamp,
+			"--redirect-db", "restoredb",
+			"--with-stats",
+			"--backup-dir", backupDir)
+
+		Expect(string(output)).To(ContainSubstring("Query planner statistics restore complete"))
+		testhelper.AssertQueryRuns(restoreConn, "SET SESSION gp_autostats_mode='none'")
```
jimmyyih, line 1046: Are you expecting this to check if gprestore ran the `SET SESSION` statement?  If so, that's not what the `testhelper.AssertQueryRuns()` function does.  It merely runs the command you put in and checks for an error.

end_to_end/end_to_end_suite_test.go
```diff
@@ -1018,10 +1019,42 @@ var _ = Describe("backup and restore end to end tests", func() {
...
+
+		restoredTablesAnalyzed := dbconn.MustSelectString(restoreConn,
+			`select count(*) from pg_stat_last_operation where objid='public.table_to_analyze'::regclass::oid and staactionname='ANALYZE';`)
```
jimmyyih, line 1049: Upper `select`, `from`, `where`, and `and`.

end_to_end/end_to_end_suite_test.go
```diff
@@ -1039,7 +1072,7 @@ var _ = Describe("backup and restore end to end tests", func() {
 			"--include-table", "public.table_to_include_with_stats")
 
 		rawCount := dbconn.MustSelectString(restoreConn,
-			"SELECT COUNT(*) FROM pg_stat_all_tables WHERE schemaname='public'")
+			"Select count(*) from pg_statistic WHERE starelid = 'public.table_to_include_with_stats'::regclass::oid;")
```
jimmyyih, line 1075: Upper `Select` and `from`.

end_to_end/end_to_end_suite_test.go
```diff
@@ -1039,7 +1072,7 @@ var _ = Describe("backup and restore end to end tests", func() {
 			"--include-table", "public.table_to_include_with_stats")
 
 		rawCount := dbconn.MustSelectString(restoreConn,
-			"SELECT COUNT(*) FROM pg_stat_all_tables WHERE schemaname='public'")
```
jimmyyih, line 1042: Good catch.  This didn't really make sense.

toc/toc.go
```diff
@@ -160,8 +160,8 @@ func shouldIncludeStatement(entry MetadataEntry, objectSet *utils.FilterSet, sch
 	shouldIncludeObject := objectSet.MatchesFilter(entry.ObjectType)
 	shouldIncludeSchema := schemaSet.MatchesFilter(entry.Schema)
 	relationFQN := utils.MakeFQN(entry.Schema, entry.Name)
-	shouldIncludeRelation := (relationSet.IsExclude && entry.ObjectType != "TABLE" && entry.ObjectType != "VIEW" && entry.ObjectType != "MATERIALIZED VIEW" && entry.ObjectType != "SEQUENCE" && entry.ReferenceObject == "") ||
-		((entry.ObjectType == "TABLE" || entry.ObjectType == "VIEW" || entry.ObjectType == "MATERIALIZED VIEW" || entry.ObjectType == "SEQUENCE") && relationSet.MatchesFilter(relationFQN) && entry.ReferenceObject == "") || // Relations should match the filter
+	shouldIncludeRelation := (relationSet.IsExclude && entry.ObjectType != "TABLE" && entry.ObjectType != "VIEW" && entry.ObjectType != "MATERIALIZED VIEW" && entry.ObjectType != "SEQUENCE" && entry.ObjectType != "STATISTICS" && entry.ReferenceObject == "") ||
```
jimmyyih, line 163: Could you explain what this change does?  It's kinda hard to follow.  Is it that when `--include-table` or `--exclude-table` during gprestore, it would not restore the statistics because they were not in this giant boolean check?

roicos: Exactly. We had a bug here either. It was it was caught when debugging --redirect-schema --with-stats because we only run --redirect-schema with --include-table. We didn't know about it because the test for --with-stats and --include-table combination was incorrect. This story revealed the whole tangle of bugs.

jimmyyih: > This story revealed the whole tangle of bugs.

Yes, it definitely did.

PR #375, opened 2020-04-15
Create savepoint before getting view definition to avoid locking

lavjain:
When querying the view definition using pg_get_viewdef(), this function
obtains dependency locks that are not released until the transaction is
committed at the end of gpbackup session. This blocks other sessions
from commands that need AccessExclusiveLock (e.g. TRUNCATE)

Creating a SAVEPOINT allows the locks to be released once the savepoint
is released or rolled back.

### DIFF
--- a/backup/queries_relations.go
+++ b/backup/queries_relations.go
@@ -288,6 +288,14 @@ func (v View) FQN() string {
 // This function retrieves both regular views and materialized views.
 // Materialized views were introduced in GPDB 7 and backported to GPDB 6.2.
 func GetAllViews(connectionPool *dbconn.DBConn) (regularViews []View, materializedViews []MaterializedView) {
+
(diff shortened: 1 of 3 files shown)


backup/queries_relations.go
```diff
@@ -288,6 +288,8 @@ func (v View) FQN() string {
 // This function retrieves both regular views and materialized views.
 // Materialized views were introduced in GPDB 7 and backported to GPDB 6.2.
 func GetAllViews(connectionPool *dbconn.DBConn) (regularViews []View, materializedViews []MaterializedView) {
+	connectionPool.MustExec("SAVEPOINT gpbackup_get_views")
```
jimmyyih, line 291: Could you add a comment above here to explain why we need to use subtransactions so that new people looking at it would quickly understand the need.  I don't want someone reverting the change simply because they think it's faster not to use subtransactions.

PR #381, opened 2020-04-22
Fix pg_class statistics dump

jimmyyih:
The relnamespace part of the UPDATE query cannot be the schema's
actual oid. The UPDATE query would always report 0 rows updated for
the majority of cases and only work if restoring to a database with
that exact schema oid for that exact schema name. This was not caught
in production because gp_autostats_mode GUC default setting would
create the statistics for pg_class at the end of COPY FROM
SEGMENT. With a recent commit turning off autostats during gprestore
--with-stats, the pg_class statistics no longer get filled
accidentally. This commit fixes the issue by adding a subquery to get
the proper schema oid for the UPDATE query.

Reference Commit:
https://github.com/greenplum-db/gpbackup/commit/0bf5340e4fc8b70c02361e5024819828e5eb6323

### DIFF
--- a/backup/statistics.go
+++ b/backup/statistics.go
@@ -37,14 +37,12 @@ func GenerateTupleStatisticsQuery(table Table, tupleStat TupleStatistic) string 
 SET
 	relpages = %d::int,
 	reltuples = %f::real
-WHERE relname = '%s'
-AND relnamespace = %d;`
+WHERE oid = '%s'::regclass::oid;`
 	return fmt.Sprintf(
 		tupleQuery,
 		tupleStat.RelPages,
 		tupleStat.RelTuples,
-		utils.EscapeSingleQuotes(tupleStat.Table),
-		table.SchemaOid)
+		utils.EscapeSingleQuotes(table.FQN()))
 }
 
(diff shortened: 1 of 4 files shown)


backup/statistics.go
```diff
@@ -32,19 +32,19 @@ func PrintStatisticsStatementsForTable(statisticsFile *utils.FileWithByteCount,
...
 
-func GenerateTupleStatisticsQuery(table Table, tupleStat TupleStatistic) string {
+func GenerateTupleStatisticsQuery(tupleStat TupleStatistic) string {
 	tupleQuery := `UPDATE pg_class
 SET
 	relpages = %d::int,
 	reltuples = %f::real
 WHERE relname = '%s'
-AND relnamespace = %d;`
+AND relnamespace = (SELECT oid FROM pg_namespace WHERE nspname = '%s');`
```
chrishajas, line 41: Might be able to do something similar to `update pg_class set ... where oid= 'public.foo::regclass` if you want.

jimmyyih: Oh duh, I always forget you can add the schema like that to regclass.  My mind keeps them separate as regclass and regnamespace.  Thanks, that's way better.

PR #382, opened 2020-04-23
CopyPluginConfigToAllHosts generates unique file names

roicos:
- as a part of this function we generate tmp
   config file on master before copy it to all
   hosts. If we make several calls of this
   function in parallel, we would race condition
   creating/removing the file with the same name.
   To allow parallel calls we change the name
   of tmp config file to unique using timestamp.

### DIFF
--- a/utils/plugin.go
+++ b/utils/plugin.go
@@ -7,6 +7,7 @@ import (
 	path "path/filepath"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/blang/semver"
 	"github.com/greenplum-db/gp-common-go-libs/cluster"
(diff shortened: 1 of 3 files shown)


utils/plugin.go
```diff
@@ -342,10 +343,12 @@ func (plugin *PluginConfig) DeletePluginConfigWhenEncrypting(c *cluster.Cluster)
...
 }
 
+// Creates a valid segment-specific plugin configuration file with unique name
 func (plugin *PluginConfig) createHostPluginConfig(contentIDForSegmentOnHost int,
-	c *cluster.Cluster) (segmentSpecificConfigFilepath string) {
+	c *cluster.Cluster) (segmentSpecificConfigFile string) {
 	// copy "general" config file to temp, and add segment-specific PGPORT value
-	segmentSpecificConfigFile := plugin.ConfigPath + "_" + strconv.Itoa(contentIDForSegmentOnHost)
+
+	segmentSpecificConfigFile = plugin.ConfigPath + "_" + strconv.FormatInt(time.Now().UnixNano(), 10) + "_" + strconv.Itoa(contentIDForSegmentOnHost)
```
jimmyyih, line 351: Oh wow, named returns is new too me.  Interesting that the debug message below would basically show a blank filename in the old implementation.  Can we change it to not use a named return to make the code easier to look at for people that don't know this Golang mechanic (also, there's already a `return segmentSpecificConfigFile` at the bottom anyways)?

I think changing it to this would be ideal:
```
	c *cluster.Cluster) string {
	// copy "general" config file to temp, and add segment-specific PGPORT value

	segmentSpecificConfigFile := plugin.ConfigPath + "_" + strconv.FormatInt(time.Now().UnixNano(), 10) + "_" + strconv.Itoa(contentIDForSegmentOnHost)
```

shivzone:
Why don't we use this opportunity to do what is mentioned in comments of the function `CopyPluginConfigToAllHosts`
`At some point in the future, the plugin MAY be able to get PGPORT as an environmental var, at which time the code to write *specific* config files per segment can be removed`

utils/plugin.go
roicos: And it did show the empty line in the old implementation. Fair enough, we don't need it here.

roicos:
> Why don't we use this opportunity to do what is mentioned in comments of the function `CopyPluginConfigToAllHosts`
> `At some point in the future, the plugin MAY be able to get PGPORT as an environmental var, at which time the code to write *specific* config files per segment can be removed`

I'll create a separate story for that.

PR #385, opened 2020-05-05
Incremental restore should filter out existing external tables

jimmyyih:
Incremental restore was not properly filtering out existing external
tables. This was because the query for finding existing tables did not
include external tables in the relkind and relstorage query
filters. Fixed by adding them to the filters. Note that 'f' for
foreign tables only matters for GPDB 6+ but we shouldn't need a GPDB
version check on this since this is catalog and 'f' is new starting
from GPDB 6+.

Commit Reference:
https://github.com/greenplum-db/gpbackup/commit/c8ab236ce88b1c94

### DIFF
--- a/restore/wrappers.go
+++ b/restore/wrappers.go
@@ -340,11 +340,13 @@ func RestoreSchemas(schemaStatements []toc.StatementWithType, progressBar utils.
 func GetExistingTableFQNs() ([]string, error) {
 	existingTableFQNs := make([]string, 0)
 
+	// Note that 'f' for foreign tables only matters for GPDB 6+ but we shouldn't need a GPDB
+	// version check on this since this is catalog and 'f' is new starting from GPDB 6+.
 	query := `SELECT quote_ident(n.nspname) || '.' || quote_ident(c.relname)
 			  FROM pg_catalog.pg_class c
 				LEFT JOIN pg_catalog.pg_namespace n ON n.oid = c.relnamespace
-			  WHERE c.relkind IN ('r','')
-			  AND c.relstorage IN ('h', 'a', 'c','')
+			  WHERE c.relkind IN ('r', 'f')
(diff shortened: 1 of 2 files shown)


restore/wrappers.go
```diff
@@ -343,8 +343,8 @@ func GetExistingTableFQNs() ([]string, error) {
 	query := `SELECT quote_ident(n.nspname) || '.' || quote_ident(c.relname)
 			  FROM pg_catalog.pg_class c
 				LEFT JOIN pg_catalog.pg_namespace n ON n.oid = c.relnamespace
-			  WHERE c.relkind IN ('r','')
-			  AND c.relstorage IN ('h', 'a', 'c','')
+			  WHERE c.relkind IN ('r', 'f')
+			  AND c.relstorage IN ('h', 'a', 'c', 'x', 'f')
```
shivzone, line 347: what kind of tables is 'x' ?

restore/wrappers.go
```diff
@@ -343,8 +343,8 @@ func GetExistingTableFQNs() ([]string, error) {
 	query := `SELECT quote_ident(n.nspname) || '.' || quote_ident(c.relname)
 			  FROM pg_catalog.pg_class c
 				LEFT JOIN pg_catalog.pg_namespace n ON n.oid = c.relnamespace
-			  WHERE c.relkind IN ('r','')
-			  AND c.relstorage IN ('h', 'a', 'c','')
+			  WHERE c.relkind IN ('r', 'f')
```
shivzone, line 346: Can we make an inline code comment about the gpdb6 relevance of 'f' just so that we don't unnecessarily add version checks in the future

restore/wrappers.go
jimmyyih: 'x' is for external tables.  'f' is for foreign tables.  In GPDB 6, external tables are partially foreign tables. In GPDB 7, external tables are fully foreign tables.

PR #388, opened 2020-05-14
Fix function and protocol filtering

kyeap-pivotal:
Fixes a regression where a backup taken using the flag `--include-table`
does not correctly filter out functions and protocols.

gpbackup reference regression commits:
functions: 5db5e23b
protocols: abdf8cf2

### DIFF
--- a/end_to_end/filtered_test.go
+++ b/end_to_end/filtered_test.go
@@ -7,6 +7,7 @@ import (
 	"github.com/greenplum-db/gp-common-go-libs/testhelper"
 	"github.com/greenplum-db/gpbackup/utils"
 	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
 )
 
 var _ = Describe("End to End Filtered tests", func() {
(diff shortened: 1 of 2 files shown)


lavjain:
We may also need to include `--include-table-file, --include-schema, --include-schema-file` flags.

end_to_end/filtered_test.go
```diff
@@ -103,6 +104,30 @@ PARTITION BY LIST (gender)
...
+			testhelper.AssertQueryRuns(backupConn,
+				`CREATE FUNCTION public.read_from_s3() RETURNS integer
+AS '$libdir/gps3ext.so', 's3_import'
+LANGUAGE c STABLE NO SQL;`)
+			defer testhelper.AssertQueryRuns(backupConn,
+				"DROP FUNCTION public.read_from_s3()")
+			testhelper.AssertQueryRuns(backupConn,
+				`CREATE PROTOCOL s3_protocol (readfunc = public.read_from_s3);`)
+			defer testhelper.AssertQueryRuns(backupConn,
+				"DROP PROTOCOL s3_protocol")
```
jimmyyih, line 121: Is there a reason you're using the S3 module?  If no specific reason, just use simple stuff to reduce unnecessary dependencies.

For example:
```
CREATE FUNCTION foofunc() RETURNS int AS 'SELECT 1' LANGUAGE SQL;
CREATE PROTOCOL s3_protocol (readfunc = foofunc);
```

kyeap-pivotal: I tried a couple of times to create a simplified function and protocol. But failed. So I copied this function and protocol from another test. I'll change to your simplified objects. Thanks.

kyeap-pivotal:
> We may also need to include `--include-table-file, --include-schema, --include-schema-file` flags.

The --include-table-file follows the same logic path at --include-table. The test case above covers us. `--include-schema` and `--include-schema-file` filters are built directly into the SQL query and use different logic. Schema filters are also working as intended.

PR #389, opened 2020-05-14
Publish gpbackup ubuntu debian package to S3 and pivnet

lavjain:
The `.deb` file is included in the gppkgs tarball that we create in the `gpbackup` pipeline. This file will also be uploaded to pivnet by the `gpbackup-release` pipeline along with other gppkgs.

### DIFF
--- a/ci/scripts/build-gppkgs.bash
+++ b/ci/scripts/build-gppkgs.bash
@@ -51,6 +51,8 @@ if [[ ${OS} == "RHEL" || ${OS} == "SLES" ]]; then
     build_rpm_rhel
 elif [[ ${OS} == "ubuntu" ]]; then
     build_deb_ubuntu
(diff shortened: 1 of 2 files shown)


ci/scripts/build-gppkgs.bash
```diff
@@ -81,3 +81,8 @@ echo "Successfully built gppkg"
 chown gpadmin:gpadmin *.gppkg
 ls -l *.gppkg
 mv *.gppkg gppkgs/
+
+# Publish the debian package also (if ubuntu)
+if [[ ${OS} == "ubuntu" ]]; then
+    cp ${PKG_FILES} gppkgs/
```
kyeap-pivotal, line 87: Possible to combine the two ubuntu conditionals? Put line 87 after 53?

PR #390, opened 2020-05-15
gprestore ability to discard or seek or scan subset data

shivzone:
Enhanced `gprestore` to optimize the way we scan the backup data during restore specifically in the scenario when we have an uncompressed single file backup with table/schema filters

- Introduced `RestoreReader` which encapsulates both reader types `io.ReadSeeker` and `bufio.Reader`. We will use the appropriate reader based on the `ReaderType`. The reader can be one one of the following types
- reader that can directly seek during restore (restore from a filesystem)
- reader that can read subset data based on the target table offsets from `toc` (when we use plugin)
-reader that read entire data and based on offsets from toc, data will need to be discarded.

The functions to move to a specific reader offset or copy data from the reader are now functions of the `RestoreReader`

- Introduced a new function to the plugin interface named `restore_data_subset` which is optimized for reading subset data. In order to make use of this the plugin config file needs to set `restore_subset: "on"`
This takes in an offset file containing all the start and end byte offsets of the relevant `oids` that need to be restored. 

- ddboost plugin tests have been added to plugin_test_bench.sh
- ddboost scale tests have been newly introduced in plugin_test_scale.sh
- s3_plugin doesn't support the new api

### DIFF
--- a/helper/restore_helper.go
+++ b/helper/restore_helper.go
@@ -2,11 +2,14 @@ package helper
 
 import (
 	"bufio"
+	"bytes"
 	"compress/gzip"
+	"encoding/binary"
 	"fmt"
 	"io"
 	"os"
 	"os/exec"
+	"path/filepath"
 	"strings"
 
 	"github.com/greenplum-db/gpbackup/toc"
@@ -17,9 +20,17 @@ import (
 /*
  * Restore specific functions
  */
+type ReaderType int
+
+type RestoreReader struct {
+	bufReader		*bufio.Reader
+	isSubsetReader	bool
+}
 
 func doRestoreAgent() error {
-	tocEntries := toc.NewSegmentTOC(*tocFile).DataEntries
+	segmentTOC := toc.NewSegmentTOC(*tocFile)
+	tocEntries := segmentTOC.DataEntries
+
 	var lastByte uint64
 	var bytesRead int64
 	var start uint64
@@ -33,7 +44,7 @@ func doRestoreAgent() error {
 		return err
 	}
 
-	reader, err := getRestoreDataReader()
+	reader, err := getRestoreDataReader(segmentTOC)
 	if err != nil {
 		return err
 	}
@@ -70,7 +81,7 @@ func doRestoreAgent() error {
 		}
 
 		log(fmt.Sprintf("Data Reader - Start Byte: %d; End Byte: %d; Last Byte: %d", start, end, lastByte))
-		numDiscarded, err = reader.Discard(int(start - lastByte))
+		numDiscarded, err = reader.bufReader.Discard(int(start - lastByte))
 		if err != nil {
 			// Always hard quit if data reader has issues
 			_ = utils.RemoveFileIfExists(currentPipe)
@@ -79,7 +90,7 @@ func doRestoreAgent() error {
 		log(fmt.Sprintf("Data Reader discarded %d bytes", numDiscarded))
 
 		log(fmt.Sprintf("Restoring table with oid %d", oid))
-		bytesRead, err = io.CopyN(writer, reader, int64(end-start))
+		bytesRead, err = io.CopyN(writer, reader.bufReader, int64(end-start))
 		if err != nil {
 			// In case COPY FROM or copyN fails in the middle of a load. We
 			// need to update the lastByte with the amount of bytes that was
@@ -120,11 +131,14 @@ func doRestoreAgent() error {
 	return lastError
 }
 
-func getRestoreDataReader() (*bufio.Reader, error) {
+func getRestoreDataReader(tocEntries *toc.SegmentTOC) (*RestoreReader, error) {
+	var restoreReader *RestoreReader
 	var readHandle io.Reader
+	var isSubsetReader bool
 	var err error
+
 	if *pluginConfigFile != "" {
-		readHandle, err = startRestorePluginCommand()
+		readHandle, isSubsetReader, err = startRestorePluginCommand(tocEntries)
 	} else {
 		readHandle, err = os.Open(*dataFile)
 	}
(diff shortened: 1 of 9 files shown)


helper/restore_helper.go
```diff
@@ -160,21 +176,34 @@ func getRestorePipeWriter(currentPipe string) (*bufio.Writer, *os.File, error) {
...
-func startRestorePluginCommand() (io.Reader, error) {
+func startRestorePluginCommand(tocEntries *toc.SegmentTOC) (io.Reader, bool, error) {
 	pluginConfig, err := utils.ReadPluginConfig(*pluginConfigFile)
 	if err != nil {
-		return nil, err
+		return nil, false, err
+	}
+	cmdStr := ""
+	isSubset := false
+	if strings.HasPrefix(pluginConfig.ExecutablePath, "gpbackup_ddboost_plugin") && *isFilter && *compressionLevel > 0 {
```
jimmyyih: Thought we were going to make this a general feature and not hardcode the plugin names into the gprestore code.  Was there anything wrong with having it as a plugin config option?

jimmyyih: Okay, nvm.  Looking at separate commits is not good here.

helper/restore_helper.go
```diff
@@ -17,14 +20,62 @@ import (
...
+const (
+	SEEKABLE ReaderType = "seekable"	// reader which supports seek
+	NONSEEKABLE			= "discard"		// reader which is not seekable
+	SUBSET				= "subset"		// reader which has filtered data
+)
+type RestoreReader struct {
+	bufReader	*bufio.Reader
+	reader		io.ReadSeeker
+	readerType	ReaderType
+}
```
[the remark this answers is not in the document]
shivzone: Added

shivzone:
Yea this PR will stay open until the tests are done and perf is evaluated

helper/restore_helper.go
```diff
@@ -17,14 +20,70 @@ import (
...
+			// Always hard quit if data reader has issues
+			_ = utils.RemoveFileIfExists(currentPipe)
+			return err
+		}
+		log(fmt.Sprintf("Data Reader seeked forward to %d byte offset", seekPosition))
+	case NONSEEKABLE:
+		numDiscarded, err := r.bufReader.Discard(int(pos))
+		if err != nil {
+			// Always hard quit if data reader has issues
+			_ = utils.RemoveFileIfExists(currentPipe)
```
kyeap-pivotal, line 57: nit: lines 49 and 57 ignores errors if the files fail to get cleaned up.

helper/restore_helper.go
```diff
@@ -160,21 +240,39 @@ func getRestorePipeWriter(currentPipe string) (*bufio.Writer, *os.File, error) {
...
+	if pluginConfig.CanRestoreSubset() && *isFilter && !strings.HasSuffix(*dataFile, ".gz") {
+		offsetsFile, _ := ioutil.TempFile("/tmp", "gprestore_offsets_")
+		defer func() {
+			offsetsFile.Close()
+			os.Remove(offsetsFile.Name())
+		}()
+		buf := new(bytes.Buffer)
+		for _, entry := range tocEntries.DataEntries {
+			_ = binary.Write(buf, binary.LittleEndian, entry.StartByte)
+			_ = binary.Write(buf, binary.LittleEndian, entry.EndByte)
```
kyeap-pivotal, line 259: Errors writing offsets are not handled.

utils/agent_remote.go
```diff
@@ -122,13 +122,17 @@ func StartGpbackupHelpers(c *cluster.Cluster, fpInfo filepath.FilePathInfo, oper
 	if onErrorContinue {
 		onErrorContinueStr = " --on-error-continue"
 	}
+	isFilterStr := ""
```
kyeap-pivotal, line 125: nit: rename `isFilerterStr` to `filterStr`

helper/restore_helper.go
```diff
@@ -70,16 +130,13 @@ func doRestoreAgent() error {
...
 		if err != nil {
-			// Always hard quit if data reader has issues
-			_ = utils.RemoveFileIfExists(currentPipe)
 			return err
 		}
-		log(fmt.Sprintf("Data Reader discarded %d bytes", numDiscarded))
 
 		log(fmt.Sprintf("Restoring table with oid %d", oid))
-		bytesRead, err = io.CopyN(writer, reader, int64(end-start))
+		bytesRead, err = reader.copyData(int64(end-start))
```
lavjain, line 139: For consistency, it would be better to pass a writer object into `copyData()` function call.

plugins/plugin_test_bench.sh
```diff
@@ -201,6 +203,44 @@ echo "[PASSED] backup_data with no data"
...
+output=`$plugin restore_data_subset $plugin_config $testdatalarge "$testdir/offsets"`
+data_part1=$(echo $data_large | cut -c1-700000)
+data_part2=$(echo $data_large | cut -c900001-900001)
+if [ "$output" != "$data_part1$data_part2" ]; then
+  echo "Failure restore_data_subset of two partitions from large data"
+  exit 1
+fi
+echo "[PASSED] restore_data_subset of two partitions from large data"
+cleanup_test_dir $testdir
+
```
lavjain, line 243: Wonderful :-)

plugins/plugin_test_bench.sh
```diff
@@ -306,18 +346,21 @@ set -e
 # ----------------------------------------------
 # Run test gpbackup and gprestore with plugin
 # ----------------------------------------------
-
 #gpbackup --dbname $test_db --plugin-config $plugin_config $further_options > $log_file
 
-
 test_backup_and_restore_with_plugin() {
     flags=$1
+    restore_filter=$2
```
lavjain, line 353: Minor, maybe add `single_data-file` as $2 and `restore filter` as $3

shivzone: `flags` is supposed to cover all the gpbackup/gprestore flags including single-data-file,compression,etc. That way we don't need to define a new argument for each flag.
The additional argument `restore_filter` is more of a custom test parameter to signal to the test that we want to restore with filters, as the caller would have no clue about what tables to filter.

PR #391, opened 2020-05-20
Error out if target database is missing any objects

lavjain:
during incremental restore unless --on-error-continue

### DIFF
--- a/restore/restore.go
+++ b/restore/restore.go
@@ -201,6 +201,10 @@ func restorePredata(metadataFilename string) {
 		existingTableFQNs, err := GetExistingTableFQNs()
 		gplog.FatalOnError(err)
 
+		existingSchemasMap := make(map[string]Empty)
+		for _, schema := range existingSchemas {
+			existingSchemasMap[schema] = Empty{}
+		}
 		existingTablesMap := make(map[string]Empty)
 		for _, table := range existingTableFQNs {
 			existingTablesMap[table] = Empty{}
@@ -224,7 +228,8 @@ func restorePredata(metadataFilename string) {
 				if utils.RelationIsExcludedByUser(inRelationsUserInput, exRelationsUserInput, table) {
 					tablesExcludedByUserInput = append(tablesExcludedByUserInput, table)
 				} else {
-					if !utils.Exists(schemasToCreate, schemaName) {
+					_, ok := existingSchemasMap[schemaName]
+					if !ok && !utils.Exists(schemasToCreate, schemaName) {
 						schemasToCreate = append(schemasToCreate, schemaName)
 					}
 					tableFQNsToCreate = append(tableFQNsToCreate, table)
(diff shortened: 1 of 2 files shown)


restore/restore.go
```diff
@@ -224,24 +228,30 @@ func restorePredata(metadataFilename string) {
 				if utils.RelationIsExcludedByUser(inRelationsUserInput, exRelationsUserInput, table) {
 					tablesExcludedByUserInput = append(tablesExcludedByUserInput, table)
 				} else {
-					if !utils.Exists(schemasToCreate, schemaName) {
+					_, ok := existingSchemasMap[schemaName]
```
roicos, line 231:
```suggestion
					_, schemaNotExists := !existingSchemasMap[schemaName]
					prefilteredToCreate := !utils.Exists(schemasToCreate, schemaName)
					if schemaNotExists && !prefilteredToCreate {
```

restore/restore.go
```diff
@@ -296,8 +306,8 @@ func restoreData() {
 	restorePlan := backupConfig.RestorePlan
 	restorePlanEntries := make([]history.RestorePlanEntry, 0)
 	if MustGetFlagBool(options.INCREMENTAL) {
-		restorePlanEntries = append(restorePlanEntries, restorePlan[len(backupConfig.RestorePlan)-1])
-
+		restorePlanEntries = append(restorePlanEntries,
```
shivzone, line 309: This change seems unnecessary

restore/restore.go
```diff
@@ -224,24 +228,30 @@ func restorePredata(metadataFilename string) {
 				if utils.RelationIsExcludedByUser(inRelationsUserInput, exRelationsUserInput, table) {
 					tablesExcludedByUserInput = append(tablesExcludedByUserInput, table)
 				} else {
-					if !utils.Exists(schemasToCreate, schemaName) {
+					_, ok := existingSchemasMap[schemaName]
+					if !ok && !utils.Exists(schemasToCreate, schemaName) {
```
shivzone, line 232: Can make the whole thing a one liner to reduce clutter `if _, val := existingSchemasMap[schemaName]; !val && !utils.Exists(schemasToCreate, schemaName)`

restore/restore.go
```diff
@@ -201,6 +201,10 @@ func restorePredata(metadataFilename string) {
 		existingTableFQNs, err := GetExistingTableFQNs()
 		gplog.FatalOnError(err)
 
+		existingSchemasMap := make(map[string]Empty)
```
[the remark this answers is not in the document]
shivzone: utils.exists() is a linear vs constant time map lookup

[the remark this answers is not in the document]
shivzone: In some scenarios we use it for a rather small input like `includeSchemas` which are rather small and using a map is an overkill.

PR #392, opened 2020-05-20
Move truncate operation to before data restore

lavjain:
Remove bulk truncate and truncate tables one at a time (if incremental or truncate flag is specified)

### DIFF
--- a/restore/wrappers.go
+++ b/restore/wrappers.go
@@ -369,6 +369,7 @@ func GetExistingSchemas() ([]string, error) {
 }
 
 func TruncateTable(tableFQN string) error {
+	gplog.Verbose("Truncating table %s prior to restoring data", tableFQN)
 	_, err := connectionPool.Exec(`TRUNCATE ` + tableFQN)
 	return err
 }
--- a/restore/wrappers.go
+++ b/restore/wrappers.go
@@ -368,15 +368,7 @@ func GetExistingSchemas() ([]string, error) {
 	return existingSchemas, err
 }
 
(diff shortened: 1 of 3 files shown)


restore/wrappers.go
```diff
@@ -368,15 +368,7 @@ func GetExistingSchemas() ([]string, error) {
...
-	tableFQNs := make([]string, 0)
-	for _, entry := range entries {
-		tableFQN := utils.MakeFQN(entry.Schema, entry.Name)
-		tableFQNs = append(tableFQNs, tableFQN)
-	}
-	query += strings.Join(tableFQNs, ",")
-	query += ";"
-	_, err := connectionPool.Exec(query)
+func TruncateTable(tableFQN string) error {
+	_, err := connectionPool.Exec(`TRUNCATE ` + tableFQN)
```
jimmyyih, line 372: Would be nice to add a verbose log here saying we are truncating the table since you're removing the only truncate log we had at the old caller.  Would make things easier to debug later as well.

lavjain:
Majority (>95%) of the restores should not see any time delay because of this change. As @jimmyyih pointed out, COPY operation for most tables is significantly higher than TRUNCATE. I agree that the restore times may increase if the customer is restoring thousands of small tables.

Remove bulk truncate would also leave the database in more consistent state as we would be only truncating the tables that are being restored. Therefore, in case of an error (restore completed partially) the customer will not end up with tables that got truncated unnecessarily.

lavjain:
> I would say:
> 
> Truncating table %s prior to restoring data
> 
> im going with the assumption that this will be in the format of schemaname.tablename, correct..

That is correct

PR #396, opened 2020-06-11
Remove undocumented external table option attribute

kyeap-pivotal:
After a changes to external tables in GPDB, it was discovered that the
options attribute on external tables was an undocumented feature. Since
it is undocumented it should be safe to remove. This commit also fixes
some formatting changes to format options attribute that was the result
of commits that changed how external table format options were stored.

GPDB reference commits:
https://github.com/greenplum-db/gpdb/commit/e4b499aa4a16df49241abc9dd70acec5589f1323
https://github.com/greenplum-db/gpdb/commit/eddd3a6e08e81c19b4588a3718cf1cb7197cfaf9

### DIFF
--- a/integration/predata_externals_create_test.go
+++ b/integration/predata_externals_create_test.go
@@ -25,7 +25,7 @@ var _ = Describe("backup integration create statement tests", func() {
 		BeforeEach(func() {
 			extTable = backup.ExternalTableDefinition{Oid: 0, Type: 0, Protocol: backup.FILE, Location: "file://tmp/ext_table_file",
 				ExecLocation: "ALL_SEGMENTS", FormatType: "t", FormatOpts: "delimiter '	' null '\\N' escape '\\'",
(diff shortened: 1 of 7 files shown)


integration/predata_externals_create_test.go
```diff
@@ -101,7 +101,7 @@ var _ = Describe("backup integration create statement tests", func() {
 			extTable.Type = backup.READABLE
 			extTable.Writable = false
 			extTable.FormatType = "b"
-			extTable.FormatOpts = "formatter 'fixedwidth_out' i '20' "
+			extTable.FormatOpts = "formatter 'fixedwidth_out'i '20'"
```
jimmyyih, line 104: Have you tested this against GPDB versions that are not master (7devel)?  I would have expected this to have a version check.

kyeap-pivotal: You're absolutely right. This needs a version check.

Issue #387, opened 2020-05-06
gpbackup on Solaris


japinli:
When I build gpbackup on Solaris, there is an error like following:

``` shell
[root@a686a8c2-73cf-4796-bb62-e7caca687b9f ~/gocode/src/github.com/greenplum-db/gpbackup]# make build
GO111MODULE=on  go build -mod=readonly -tags 'gpbackup' -o /root/gocode/bin/gpbackup -ldflags "-X github.com/greenplum-db/gpbackup/backup.version=1.17.0+dev.72.g4ecd0d9"
GO111MODULE=on  go build -mod=readonly -tags 'gprestore' -o /root/gocode/bin/gprestore -ldflags "-X github.com/greenplum-db/gpbackup/restore.version=1.17.0+dev.72.g4ecd0d9"
GO111MODULE=on  go build -mod=readonly -tags 'gpbackup_helper' -o /root/gocode/bin/gpbackup_helper -ldflags "-X github.com/greenplum-db/gpbackup/helper.version=1.17.0+dev.72.g4ecd0d9"
# github.com/greenplum-db/gpbackup/helper
helper/helper.go:127:9: undefined: syscall.Mkfifo
make: *** [Makefile:77: build] Error 2
```

After some search, I find that there are some system calls are not present on Solaris in syscall in Go [1].  Here is my fix, however, it maybe not suit for Linux.

```diff
diff --git a/go.mod b/go.mod
index 12c6f08..f28b22d 100644
--- a/go.mod
+++ b/go.mod
@@ -20,6 +20,7 @@ require (
        github.com/sergi/go-diff v1.1.0
        github.com/spf13/cobra v0.0.5
        github.com/spf13/pflag v1.0.5
+       golang.org/x/sys v0.0.0-20200223170610-d5e6a3e2c0ae
        golang.org/x/tools v0.0.0-20200214225126-5916a50871fb
        gopkg.in/cheggaaa/pb.v1 v1.0.28
        gopkg.in/yaml.v2 v2.2.8
diff --git a/helper/helper.go b/helper/helper.go
index 93517c2..7604d22 100644
--- a/helper/helper.go
+++ b/helper/helper.go
@@ -14,6 +14,7 @@ import (
        "sync"
        "syscall"

+       "golang.org/x/sys/unix"
        "github.com/greenplum-db/gp-common-go-libs/gplog"
        "github.com/greenplum-db/gp-common-go-libs/operating"
[...]

PR #398, opened 2020-06-19
Minor patches to make gpbackup/gprestore work on Solaris

jimmyyih:
```
commit f6994ad252d3af6ccfcf2bff9aede1c23110989f
    Update test paths to properly use $GOPATH

    The tests in this patch were all hardcoded to assume $GOPATH was
    $HOME/go. That's not a safe assumption. We should instead use $GOPATH
    to be more OSS developer friendly.

commit 73533befa61517195ffc14ffdc3c6ea763d84600
    Fix wrappers unit test cleanup

    We should move out of the directory before removing it. Some operating
    systems will error out on rmdir if there's an active reference.

commit 6fcafd0b258c085c5f207971ae7e7fd8894137fb
    Use x/sys/unix for Mkfifo call

    The syscall standard library is frozen and does not work for certain
    operating systems (e.g. Solaris). The x/sys/unix library is more up to
    date and has more OS support. Switching libraries for the Mkfifo call
    should be okay as there is no regression found while investigating its
    feasability.
```

Fixes https://github.com/greenplum-db/gpbackup/issues/387 which will help @japinli support gpbackup/gprestore on his Solaris platform.

### DIFF
--- a/end_to_end/end_to_end_suite_test.go
+++ b/end_to_end/end_to_end_suite_test.go
@@ -303,8 +303,8 @@ var _ = BeforeSuite(func() {
 	// This is used to run tests from an older gpbackup version to gprestore latest
 	useOldBackupVersion = os.Getenv("OLD_BACKUP_VERSION") != ""
 	pluginConfigPath =
-		fmt.Sprintf("%s/go/src/github.com/greenplum-db/gpbackup/plugins/example_plugin_config.yaml",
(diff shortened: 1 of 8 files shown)


jimmyyih:
> We can possibly have a var for gpbackup location in testutils

I thought about that but some tests didn't match up with others.  Since there's a tiny bit of work to be done there, we can leave it for a future test refactor.

PR #399, opened 2020-06-23
Include failed backups into history file

roicos:
- Previously we only added successful backups in the
   history file. This commit moves WriteBackupHistory
   into deferred DoTeardown() function that executes
   even if there was an error (unless we explicitly
   exited before). 
- It also adds Failed field to reflect the status of backup in the 
  history file and makes restore to ignore backups that failed
  when searching for the backup with specific timestamp
  or latest backup with matching config (for --incremental)
- Validated manually: 
    * if history already has some records without Failed field 
      gpbackup will just treat them as ones that have Failed false
   * on next incremental restore plan will move the record for the table
     updated in failed backup into next successful.

### DIFF
--- a/backup/backup.go
+++ b/backup/backup.go
@@ -155,7 +155,6 @@ func DoBackup() {
 		backupReport.RestorePlan = PopulateRestorePlan(backupSetTables, targetBackupRestorePlan, dataTables)
 		backupData(backupSetTables)
 	}
-
 	if MustGetFlagBool(options.WITH_STATS) {
 		backupStatistics(metadataTables)
 	}
@@ -175,9 +174,6 @@ func DoBackup() {
 		_ = utils.CopyFile(pluginConfigFlag, globalFPInfo.GetPluginConfigPath())
 		pluginConfig.MustBackupFile(globalFPInfo.GetPluginConfigPath())
 	}
-
-	err := history.WriteBackupHistory(globalFPInfo.GetBackupHistoryFilePath(), &backupReport.BackupConfig)
-	gplog.FatalOnError(err)
 }
 
 func backupGlobals(metadataFile *utils.FileWithByteCount) {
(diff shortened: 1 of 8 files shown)


backup/wrappers.go
```diff
@@ -98,6 +99,7 @@ func NewBackupConfig(dbName string, dbVersion string, backupVersion string, plug
 		Timestamp:             timestamp,
 		WithoutGlobals:        MustGetFlagBool(options.WITHOUT_GLOBALS),
 		WithStatistics:        MustGetFlagBool(options.WITH_STATS),
+		Failed:                true,
```
jimmyyih, line 102: Adding a `Failed` field here doesn't seem right.  The backup config file consists of mostly gpbackup arg settings (with a few exceptions when you look at the actual generated file).  Why do we need a `Failed` field in the backup config file when we could use the backup report file `backup status`?  The `backup status` in the backup report file is much more verbose too with a `backup error` field and could be used in backup parsing applications such as gpbackup_manager.

backup/backup.go
```diff
@@ -360,13 +356,23 @@ func DoTeardown() {
...
 			return
 		}
+		historyFilename := globalFPInfo.GetBackupHistoryFilePath()
 		reportFilename := globalFPInfo.GetBackupReportFilePath()
 		configFilename := globalFPInfo.GetConfigFilePath()
 
 		time.Sleep(time.Second) // We sleep for 1 second to ensure multiple backups do not start within the same second.
 
 		if backupReport != nil {
+			statusStr := history.BackupStatusFailed
```
jimmyyih, line 366:
```
   365:			if backupReport != nil {
=> 366:				statusStr := history.BackupStatusFailed
   367:				if !backupFailed {
   368:					statusStr = history.BackupStatusSucceed
   369:				}
   370:				backupReport.BackupConfig.Status = statusStr
   371:				backupReport.ConstructBackupParamsString()
(dlv) p backupReport.BackupConfig.Status
"failed"
```
The `backupReport.BackupConfig.Status` is already by default `BackupStatusFailed` and this is the only location we set `BackupStatusSucceed`.  Would it be better to change this to a more simpler logic?
```
if !backupFailed {
	backupReport.BackupConfig.Status = history.BackupStatusSucceed
}
```

history/history.go
```diff
@@ -18,6 +18,11 @@ type RestorePlanEntry struct {
 	TableFQNs []string
 }
 
+const (
+	BackupStatusSucceed = "succeed"
+	BackupStatusFailed  = "failed"
+)
+
```
jimmyyih, line 25:
```
if errMsg != "" {
        reportInfo = append(reportInfo,
                LineInfo{},
                LineInfo{Key: "backup status:", Value: "Failure"},
                LineInfo{Key: "backup error:", Value: errMsg})
} else {
        reportInfo = append(reportInfo,
                LineInfo{},
                LineInfo{Key: "backup status:", Value: "Success"})
}
```
Since we're trying to duplicate this backup status between the backup config file and the report file, the status values should match.  We should be using "Success" and "Failure" since it has already been established in the report file.  We should also be refactoring the above code snippet to use your new consts too for consistency.

PR #400, opened 2020-06-24
Filter out system extensions in extension queries

kyeap-pivotal:
As part of the removal of pg_exttable in GPDB master, the external table's
execution logic was moved into an external table fdw extension. A few of
Gpbackup's integration tests broke because it was not filtering out
system extensions.

GPDB reference commit:
https://github.com/greenplum-db/gpdb/commit/d86f32e5bead35167dc4ea0e2053becfc2ed6b00

### DIFF
--- a/backup/queries_functions.go
+++ b/backup/queries_functions.go
@@ -719,12 +719,13 @@ func (e Extension) FQN() string {
 func GetExtensions(connectionPool *dbconn.DBConn) []Extension {
 	results := make([]Extension, 0)
 
-	query := `
+	query := fmt.Sprintf(`
 	SELECT e.oid,
 		quote_ident(extname) AS name,
 		quote_ident(n.nspname) AS schema
 	FROM pg_extension e
-		JOIN pg_namespace n ON e.extnamespace = n.oid`
+		JOIN pg_namespace n ON e.extnamespace = n.oid
+	WHERE e.oid >= %d`, FIRST_NORMAL_OBJECT_ID)
 	err := connectionPool.Select(&results, query)
 	gplog.FatalOnError(err)
 	return results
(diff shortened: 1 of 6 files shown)


backup/queries_functions.go
```diff
@@ -719,12 +719,13 @@ func (e Extension) FQN() string {
...
 
-	query := `
+	query := fmt.Sprintf(`
 	SELECT e.oid,
 		quote_ident(extname) AS name,
 		quote_ident(n.nspname) AS schema
 	FROM pg_extension e
-		JOIN pg_namespace n ON e.extnamespace = n.oid`
+		JOIN pg_namespace n ON e.extnamespace = n.oid
+	WHERE e.oid >= %d`, FIRST_NORMAL_OBJECT_ID)
```
jimmyyih, line 728: Wow, that only works because they put the creation of the extension in gp_toolkit:
```
--- a/src/backend/catalog/gp_toolkit.sql
+++ b/src/backend/catalog/gp_toolkit.sql
@@ -180,6 +180,8 @@ AS
 GRANT SELECT ON TABLE gp_toolkit.__gp_number_of_segments TO public;


+CREATE EXTENSION gp_exttable_fdw;
```

I guess that works but pg_extension already had plpgsql so why wouldn't this change create another diff issue now that plpgsql is being excluded from the results?
```
postgres=# select oid, extname from pg_extension;
  oid  |     extname
-------+-----------------
 12562 | plpgsql
 13040 | gp_exttable_fdw
(2 rows)
```

kyeap-pivotal: As discovered by us earlier. The plpgsql extension is being dropped in the test suite, which is why further diff happened when the system extension filter is put in place. I have removed the this drop statement as it is no longer needed.

PR #401, opened 2020-06-26
ddboost perf and refactored plugin perf workflow

shivzone:
ddboost_plugin_perf is the pipeline job to benchmark ddboost plugin (cluster on aws t2.2xlarge )
s3_plugin_perf is the the pipeline job to benchmark job s3 plugin (cluster on gcp n1-standard-8 )

both perf jobs use a common data setup script setup-perf.bash to setup tpch database with a configurable scale factor.
both perf jobs  use a common scale testing script plugin_test_scale.sh to benchmark backup and restore against various scenarios/flags.

s3_plugin_test job modified to only run regular tests and not any performance related ones.

TODO: Will need to migrate the ddve to gcp so that we can use identical instance types for ddboost and s3 perf jobs.


plugins/plugin_test_scale.sh
```diff
@@ -89,39 +82,39 @@ test_backup_and_restore_with_plugin() {
...
-
-echo
-echo "DISABLED restore_subset"
-plugin_config_temp=$(mktemp)
-cat $plugin_config > $plugin_config_temp
-echo "  restore_subset: \"off\"" >> $plugin_config_temp
-test_backup_and_restore_with_plugin "$plugin_config_temp" "--single-data-file --no-compression" "restore-filter"
+test_db=tpchdb
+restore_filter="--include-table public.lineitem_100"
+#test_backup_and_restore_with_plugin "$plugin_config" "" "--jobs=4"
```
jimmyyih, line 93: Should this line be removed?

shivzone: I wanted to have the scale test benchmark full restore, but they take forever on t2.2xlarge.

PR #403, opened 2020-06-30
Skips empty lines in files provided for filtering

roicos:
- Files provided in include/exclude-table/schema-file
   flags when generated automatically may contain
   empty lines. When we parse them to set options
   and update initial user provided include/exclude
   flags we now skip this empty lines


roicos:
Removed trim for now.

options/options.go
```diff
@@ -89,15 +89,18 @@ func setFiltersFromFile(initialFlags *pflag.FlagSet, filterFlag string, filterFi
...
 		if err != nil {
 			return nil, err
 		}
 		// copy any values for flag filterFileFlag into global flag for filterFlag
-		for _, fqn := range filters {
-			err = initialFlags.Set(filterFlag, fqn) //This appends to the slice underlying the flag.
-			if err != nil {
-				return nil, err
+		for _, fqn := range filterLines {
+			if len(strings.TrimSpace(fqn)) > 0 {
```
jimmyyih, line 98: Do we need this `strings.TrimSpace()` as well?  IMO, we should error out if you see a filterLine with only spaces.  Take my previous scenario as an example but with a bit extra:
```
In psql session...

CREATE TABLE public."  " (a int);
CREATE TABLE public.foobar (a int);
\! echo "  \npublic.foobar" > /tmp/include_file
\! gpbackup --dbname postgres --include-table-file=/tmp/include_file
```

The above example should error out saying the table "  " is not fully qualified.  Since this new code is trimming the spaces, only `public.foobar` will be backed up when the user meant to restore both `public.foobar` and `public."  "` but forgot to fully qualify.  IMO, we should only be skipping over blank lines.

roicos: Then we will see "not fully qualified" error on each line empty line with spaces. But if we prefer this to skipping blank lines and only address empty lines, updated commit with no trimming on comparison.

shivzone: This is a very exceptional case for a user to have public." " and have only " " in the file.
I would personally prefer us just ignoring lines which are either empty only have \s+ in the line

shivzone: Given that this is quite rare, i'm fine with the existing approach too

PR #405, opened 2020-07-03
Enforce incremental restore to require --data-only

shivzone:
Incremental restore only restores data incrementally and doesn't restore
any metadata (predata/postdata). Validation check has been added to
enforce restore with --incremental to only work with --data-only.

Identifying the missing relations/schemas is now a separate function `verifyIncrementalState` outside of `restorePreData`


restore/restore.go
```diff
@@ -141,7 +146,7 @@ func DoRestore() {
 		restoreData()
 	}
 
-	if !isDataOnly && !MustGetFlagBool(options.INCREMENTAL) {
+	if !isDataOnly && !isIncremental {
```
roicos, line 149: We have !isDataOnly && !isIncremental check for postdata but only !isDataOnly for predata. Shouldn't we make these two "if" conditions the same?

shivzone: Agreed. Its updated now

PR #406, opened 2020-07-08
Adds validation for options field in plugin-config

roicos:
If .yml file is not formatted correctly (including additional fields present on the root level) 
the user will get the message "plugin config file is formatted incorrectly".
If options field is missing or empty - the user will get the message
"options field is missing or empty in config file".

### DIFF
--- a/utils/plugin.go
+++ b/utils/plugin.go
@@ -44,12 +44,15 @@ func ReadPluginConfig(configFile string) (*PluginConfig, error) {
 	if err != nil {
 		return nil, err
 	}
-	err = yaml.Unmarshal(contents, config)
+	err = yaml.UnmarshalStrict(contents, config)
 	if err != nil {
-		return nil, err
+		return nil, errors.New("plugin config file is formatted incorrectly")
 	}
 	if config.ExecutablePath == "" {
 		return nil, errors.New("executablepath is required in config file")
+	}
+	if config.Options == nil {
+		config.Options = make(map[string]string)
 	}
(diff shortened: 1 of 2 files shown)


utils/plugin.go
```diff
@@ -51,6 +51,9 @@ func ReadPluginConfig(configFile string) (*PluginConfig, error) {
 	if config.ExecutablePath == "" {
 		return nil, errors.New("executablepath is required in config file")
 	}
+	if config.Options == nil {
```
shivzone, line 54: What happens if there is an options field but no values in it ?

jimmyyih:
According to the plugin documentation, only `executablepath` is a required yaml field for the plugin configuration.  Are we changing this to now require the `options` field?

The documentation snippet from https://github.com/greenplum-db/gpbackup/blob/master/plugins/README.md:
The executablepath is a required parameter and must point to the absolute path of the executable on each host. Additional parameters may be specified under the options key as required by the specific plugin. Refer to the documentation for the plugin you are using for additional required paramters.

IMO, the answer is probably yes in that we should have `options` as a required field even if it would be empty.  If so, we'll need to update the plugin documentation too (possibly in this PR or create a story).

utils/plugin.go
roicos: It will be discovered on ddboost plugin side, but the plugin will give the exact field name that is not there, not mysterious null map error.

roicos:
> According to the plugin documentation, only `executablepath` is a required yaml field for the plugin configuration. Are we changing this to now require the `options` field?
> 
Left the options field optional, only error out on incorrectly formatted yml.

PR #408, opened 2020-08-04
Better synchronize Concourse resources across pipeline jobs

jimmyyih:
When multiple commits come in around a specific window of time, a race
condition is created where the jobs after `build_binaries` could be
using a more updated resource (e.g. gpbackup HEAD, s3 plugin HEAD,
etc). This can cause intermittent failures... but when it doesn't
cause a failure, it's a bigger cause for concern since we're not
testing the correct things. Fix this by adding `passed` parameter to
resources that need to be synchronized across jobs.


jimmyyih:
Will update pipeline configurations after merging these changes.

PR #412, opened 2020-08-12
Optimize writing history file

roicos:
- If we are not doing search in history but only
   add the last backup, there is no need to load
   tho whole file in the memory just to add one entry.
   Also there is no need to re-sort the history every
   time we write it as the latest entry always goes
   in the beginning of the file.

Tested on ~100 Mb gpbackup_history file
 `gpbackup —dbname postgres` command
Time between Post-data metadata backup complete and Email report

Timing from the end of backupPostdata to the beginning of emailReport
400Mb, master
1) 2 min 14.642 sec
2) 2 min 21.368 sec
3) 2 min 28.188 sec

400Mb, fix branch
1) Time: 2.033 sec
2) Time: 2.090 sec
3) Time: 2.157 sec

Timing for writing backup history only
without fix
1050mb 10m13.70491183s
with fix
1050mb 1.550556176s

### DIFF
--- a/history/history.go
+++ b/history/history.go
@@ -1,8 +1,10 @@
 package history
 
 import (
+	"io"
 	"io/ioutil"
 	"os"
+	"path/filepath"
 	"sort"
 	"time"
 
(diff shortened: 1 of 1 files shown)


history/history.go
```diff
@@ -101,28 +102,67 @@ func CurrentTimestamp() string {
...
+		lock := lockHistoryFile()
+		oldHistoryFile, err := os.Open(historyFilePath)
+		if err != nil {
+			return err
+		}
+		_, err = oldHistoryFile.Seek(15, 0)
+		if err != nil {
+			return err
+		}
+		_, err = io.Copy(tmpFile, oldHistoryFile)
```
jimmyyih, line 137: Nice, `io.Copy()` definitely seems to be a good choice under our circumstances.  However, I think `io.CopyBuffer()` might be better since `io.Copy()` defaults to 32KB buffer.

`io.Copy()` code:
```
if buf == nil {
	size := 32 * 1024
...
	buf = make([]byte, size)
}
```

Maybe we could supply our own bigger buffer (512KB maybe? 🤷‍♂️ ) and use `io.CopyBuffer()`?  I would suggest trying it out and seeing if there's a difference (test against small file, large file, and very large file).

PR #413, opened 2020-08-26
Update dependency gp-common-go-lib for pgx udpate

kyeap-vmware:
Update gp-common-go-lib to fix postgres connection initialization issue
that occurs when there is a folder or a file that is the same name at
the computer's hostname in the current directory when gpabckup runs.


kyeap-vmware:
Sister PR against gp-common-go-lib here: https://github.com/greenplum-db/gp-common-go-libs/pull/58

PR #416, opened 2020-08-29
Fix gprestore --truncate-table performance bottleneck when using --jobs

jimmyyih:
When using --truncate-table and --jobs together for gprestore, the
workers were all using the same database connection to execute the
TRUNCATE statements. This causes an unintentional performance
bottleneck since each worker would be waiting on that single database
connection to execute TRUNCATE. Also, that single database connection
is dedicated to its own worker so you could have an extreme bottleneck
scenario as such:
1. Worker A is running COPY FROM SEGMENT on database connection 1
2. Workers B and C are trying to run TRUNCATE on database connection 1
3. Workers B and C are queued until Worker A finishes its COPY FROM
   SEGMENT
4. Workers B and C have their own COPY FROM SEGMENT commands waiting
   on their TRUNCATE commands to finish before running on their
   dedicated database connections 2 and 3

Fix the issue by attaching connection id to the TRUNCATE statements
when executing.

### DIFF
--- a/restore/wrappers.go
+++ b/restore/wrappers.go
@@ -368,8 +368,8 @@ func GetExistingSchemas() ([]string, error) {
 	return existingSchemas, err
 }
 
-func TruncateTable(tableFQN string) error {
+func TruncateTable(tableFQN string, whichConn int) error {
 	gplog.Verbose("Truncating table %s prior to restoring data", tableFQN)
-	_, err := connectionPool.Exec(`TRUNCATE ` + tableFQN)
+	_, err := connectionPool.Exec(`TRUNCATE ` + tableFQN, whichConn)
 	return err
 }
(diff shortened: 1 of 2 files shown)


restore/wrappers.go
```diff
@@ -368,8 +368,8 @@ func GetExistingSchemas() ([]string, error) {
 	return existingSchemas, err
 }
 
-func TruncateTable(tableFQN string) error {
+func TruncateTable(tableFQN string, whichConn int) error {
 	gplog.Verbose("Truncating table %s prior to restoring data", tableFQN)
-	_, err := connectionPool.Exec(`TRUNCATE ` + tableFQN)
+	_, err := connectionPool.Exec(`TRUNCATE ` + tableFQN, whichConn)
```
shivzone, line 373: I wish the Exec function picked a random connection object instead of defaulting to connection index 0

jimmyyih: Yeah, the Exec function has bit us twice now.  I think it's okay though and we just need to keep it in mind.  Whomever remembers it really needs to call it out during PR reviews.

PR #417, opened 2020-09-10
Fix uncaught error in scale tests

jimmyyih:
In our scale tests, there are failures outputted when running the
scale_db1.sql file. The large hdb."order" partition table is not able
to be created because the default vmem protect limit is reached and
the query gets canceled. Simply double the vmem protect limit default
on the master segment to allow the master segment to process the large
CREATE TABLE statement.

### DIFF
--- a/ci/scripts/scale-tests.bash
+++ b/ci/scripts/scale-tests.bash
@@ -15,6 +15,11 @@ cat <<SCRIPT > /tmp/run_tests.bash
 
 source env.sh
 
+# Double the vmem protect limit default on the master segment to
+# prevent query cancels on large table creations (e.g. scale_db1.sql)
+gpconfig -c gp_vmem_protect_limit -v 16384 --masteronly
+gpstop -air
+
 # only install if not installed already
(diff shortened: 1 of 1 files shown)


shivzone:
Nice catch. Can we move this to setup-perf.bash script coz thats where we handle the cluster bootstrapping and the test script is mostly focussed on the actual test.

jimmyyih:
I tried not to put it in the test script but all other scripts didn't make much sense.

```
  - task: gen_cluster
    file: ccp_src/ci/tasks/gen_cluster.yml
  - task: gpinitsystem
    file: ccp_src/ci/tasks/gpinitsystem.yml
  - task: setup-centos-env
    file: gpbackup/ci/tasks/setup-centos-env.yml
  - task: scale-tests
    file: gpbackup/ci/tasks/scale-tests.yml
```

The only other task suitable would be `setup-centos-env.yml` but that's a common task for a lot of other gpbackup tests when this fix is mainly targeting the scale test specific scale_db1.sql file.

shivzone (approved):

PR #418, opened 2020-09-15
Restore sequence values during data only restores

jimmyyih:
The sequence setval commands need to be run during data only restores
since they are arguably the data of the sequence relations and can
affect user tables containing columns that reference those sequence
relations. To do this, extract the sequence setval commands from the
metadata file and run them before the data restores. This follows the
same flow of a full restore where predata sequence metadata would be
executed before data restore. Since this applies to data only
restores, technically incremental restores are improved by this as
well.

### DIFF
--- a/restore/restore.go
+++ b/restore/restore.go
@@ -4,6 +4,7 @@ import (
 	"bufio"
 	"fmt"
 	"os"
+	"regexp"
 	"runtime/debug"
 	"strings"
 	"sync"
@@ -138,6 +139,11 @@ func DoRestore() {
 
 	if !isDataOnly && !isIncremental {
 		restorePredata(metadataFilename)
+	} else if isDataOnly {
+		// The sequence setval commands need to be run during data only restores since
+		// they are arguably the data of the sequence relations and can affect user tables
+		// containing columns that reference those sequence relations.
+		restoreSequenceValues(metadataFilename)
 	}
 
 	totalTablesRestored := 0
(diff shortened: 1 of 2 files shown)


restore/restore.go
```diff
@@ -138,6 +139,11 @@ func DoRestore() {
 
 	if !isDataOnly && !isIncremental {
 		restorePredata(metadataFilename)
+	} else if isDataOnly {
```
shivzone, line 142: For sake of readability can we move this block above the previous one, so the flow will look like
```
if isIncremental {} 
else if isDataOnly {}
else {}

**Review by shivzone — APPROVED** [review]

File: restore/restore.go (hunk 1 above)
Comment (jimmyyih) [reply restore/restore.go:142]: It would have to look something like this then for combining:
```
if isDataOnly {
    if isIncremental {
        verifyIncrementalState()
    }
    restoreSequenceValues()
} else {
    restorePredata()
}
```

Although I'd rather keep the incremental verify out of this conditional block for atomicity of operations.  The incremental and data only flags for gprestore are tied pretty closely (`--data-only` and `--incremental --data-only`).  Not sure why the `!isDataOnly && !isIncremental` is there as it could be simplified to just `!isDataOnly`.  There's actually a couple instances of this so it'd probably be better to do in a separate small refactor PR if that's okay.

shivzone: Makes sense

shivzone (approved):

PR #421, opened 2020-10-06
Update pgx v4.9.0

kyeap-vmware:
Clean up blocked queries using pg_cancel_backend

Remove the original fix that cleaned up blocked query session that were
leaking if gpbackup was interrupted when waiting for a lock. With the
update to pgx v4.9.0 this fix has broken. Canceling a pgx query via
Golang's context closes the postgres connection. This means we can no
longer queue a ROLLBACK to ensure the cancel query has finished on
postgres side.  Exposing the pgconn channel that is notified when a
context cancel is finished is proving difficult because of all the
wrappers around it, dbconn(sqlx(sql(pgx(pgconn))))). We shall cancel
blocked queries in the following commit with pg_cancel_backend.

Make gpbackup postgres connection session application_names unique with
respect to timestamp. This helps differentiate gpbackup connections when
more than one instance of gpbackup is running at the same time.

Update gp-common-go-lib to fix postgres connection initialization issue
that occurs when there is a folder or a file that is the same name at
the computer's hostname in the current directory when gpabckup runs.
Update pgx to v4.9.0.


restore/data.go
```diff
@@ -15,7 +15,7 @@ import (
 	"github.com/greenplum-db/gpbackup/options"
 	"github.com/greenplum-db/gpbackup/toc"
 	"github.com/greenplum-db/gpbackup/utils"
-	"github.com/jackc/pgx"
+	"github.com/jackc/pgconn"
```
shivzone, line 18: Just to understand this, do we expect any change related to pgError in the future once gp-common-go-lib is updated ? If yes, please add a TODO note otherwise feel free to ignore.

backup/backup.go
```diff
@@ -430,19 +432,48 @@ func DoCleanup(backupFailed bool) {
...
+
+	// Query for all blocked queries
+	pids := make([]int64, 0)
+	findBlockedQuery := fmt.Sprintf("SELECT procpid from pg_stat_activity WHERE application_name='gpbackup_%s' AND waiting='t' AND  waiting_reason='lock';", timestamp)
+	if conn.Version.AtLeast("6") {
+		findBlockedQuery = fmt.Sprintf("SELECT pid from pg_stat_activity WHERE application_name='gpbackup_%s' AND waiting='t' AND  waiting_reason='lock';", timestamp)
+	}
+	err := conn.Select(&pids, findBlockedQuery)
+	gplog.FatalOnError(err)
+
```
shivzone, line 453: Would be good to have a verblose level log message indicating that we are cancelling `n` number of blocked queries (if pids is not Nil)

shivzone:
Nice effort and over all looks good !

restore/data.go
kyeap-vmware: We are not expecting any changes to pgError. The location of the definition was just moved to modularize pgx v4. So we also change to use the same definition pgx v4.9.0 uses.

backup/wrappers.go
```diff
@@ -37,13 +37,13 @@ func SetLoggerVerbosity() {
...
 
-func initializeConnectionPool() {
+func initializeConnectionPool(timestamp string) {
 	connectionPool = dbconn.NewDBConnFromEnvironment(MustGetFlagString(options.DBNAME))
 	connectionPool.MustConnect(MustGetFlagInt(options.JOBS))
 	utils.ValidateGPDBVersionCompatibility(connectionPool)
 	InitializeMetadataParams(connectionPool)
 	for connNum := 0; connNum < connectionPool.NumConns; connNum++ {
-		connectionPool.MustExec("SET application_name TO 'gpbackup'", connNum)
+		connectionPool.MustExec(fmt.Sprintf("SET application_name TO 'gpbackup_%s'", timestamp), connNum)
```
shivzone, line 46: For consistency sake, can we set the application name to use the timestamp as well during `gprestore` that way we also future proof ourselves if we need to deal with a similar scenario during restore.

kyeap-vmware: This is a valid point and we should change gprestore name to be unique, but the unique identifier will be different. While its not possible to run multiple gpbackups with the same timestamp its possible to run multiple gprestores with the same timestamp. Maybe we could use timestamp when gprestore was run? Fine for the program, but slight unintuitive for anyone trying to debug a stuck gprestore. I'm leaning toward putting the change for unique gprestore application name to be in another PR.

backup/backup.go
kyeap-vmware: PR updated

backup/wrappers.go
shivzone: i would use both similar to how we name our gprestore artifacts. Sure you can do that in a separate PR

shivzone (approved):

backup/wrappers.go
```diff
@@ -143,9 +143,11 @@ func createBackupLockFile(timestamp string) {
...
 
 func createBackupDirectoriesOnAllHosts() {
-	remoteOutput := globalCluster.GenerateAndExecuteCommand("Creating backup directories", func(contentID int) string {
-		return fmt.Sprintf("mkdir -p %s", globalFPInfo.GetDirForContent(contentID))
-	}, cluster.ON_SEGMENTS_AND_MASTER)
+	remoteOutput := globalCluster.GenerateAndExecuteCommand("Creating backup directories",
+		cluster.ON_SEGMENTS|cluster.INCLUDE_MASTER,
+		func(contentID int) string {
+			return fmt.Sprintf("mkdir -p %s", globalFPInfo.GetDirForContent(contentID))
+		})
```
jimmyyih, line 150: That's pretty harsh for gp-common-go-libs to change the function signature like this but I guess it's for the best.

backup/wrappers.go
jimmyyih: Yes, just add a story in the icebox.

backup/backup.go
```diff
@@ -430,19 +432,48 @@ func DoCleanup(backupFailed bool) {
...
+// Cancel blocked gpbackup queries waiting for locks.
+func cancelBlockedQueries(timestamp string) {
+	conn := dbconn.NewDBConnFromEnvironment(MustGetFlagString(options.DBNAME))
+	conn.MustConnect(1)
+
+	// Query for all blocked queries
+	pids := make([]int64, 0)
+	findBlockedQuery := fmt.Sprintf("SELECT procpid from pg_stat_activity WHERE application_name='gpbackup_%s' AND waiting='t' AND  waiting_reason='lock';", timestamp)
+	if conn.Version.AtLeast("6") {
+		findBlockedQuery = fmt.Sprintf("SELECT pid from pg_stat_activity WHERE application_name='gpbackup_%s' AND waiting='t' AND  waiting_reason='lock';", timestamp)
```
jimmyyih, line 449: There's an extra space char in `AND  waiting_reason`.  Same down below as well in the `blockedQueryCount`.

backup/backup.go
```diff
@@ -430,19 +432,48 @@ func DoCleanup(backupFailed bool) {
...
-
+		cancelBlockedQueries(globalFPInfo.Timestamp)
 		connectionPool.Close()
 	}
 }
 
+// Cancel blocked gpbackup queries waiting for locks.
+func cancelBlockedQueries(timestamp string) {
+	conn := dbconn.NewDBConnFromEnvironment(MustGetFlagString(options.DBNAME))
+	conn.MustConnect(1)
```
jimmyyih, line 443: Need to defer a connection close on `conn` here as good practice.

jimmyyih (approved):

PR #422, opened 2020-10-14
Run analyze against root partitions on GPDB 4.3

kyeap-vmware:
Bug fix for --run-analyze on GPDB 4.3. During analyze, GPDB 5+ has leaf
partition stats merged up to the root automatically. Against GPDB 4.3,
we must extract the root partitions from the leaf partition info and run
ANALYZE only on the root partitions.

Authored-by: Kevin Yeap <kyeap@vmware.com>

### DIFF
--- a/restore/restore.go
+++ b/restore/restore.go
@@ -440,11 +440,46 @@ func runAnalyze(filteredDataEntries map[string][]toc.MasterDataEntry) {
 			analyzeCommand := fmt.Sprintf("ANALYZE %s", tableFQN)
 
 			newAnalyzeStatement := toc.StatementWithType{
-				Schema: tableSchema,
-				Name: entry.Name,
+				Schema:    tableSchema,
+				Name:      entry.Name,
 				Statement: analyzeCommand,
 			}
 			analyzeStatements = append(analyzeStatements, newAnalyzeStatement)
+		}
+	}
+
+	// Only GPDB 5+ has leaf partition stats merged up to the root
+	// automatically. Against GPDB 4.3, we must extract the root partitions
+	// from the leaf partition info and run ANALYZE ROOTPARTITION on the root
+	// partitions. These particular ANALYZE ROOTPARTITION statements should run
+	// last so add them to the end of the analyzeStatements list.
+	if connectionPool.Version.Is("4") {
+		// Create root partition set
+		partitionRootSet := map[toc.StatementWithType]struct{}{}
+		for _, dataEntries := range filteredDataEntries {
+			for _, entry := range dataEntries {
+				if entry.PartitionRoot != "" {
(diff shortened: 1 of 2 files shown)


restore/restore.go
```diff
@@ -440,14 +440,48 @@ func runAnalyze(filteredDataEntries map[string][]toc.MasterDataEntry) {
...
+						Name:      entry.PartitionRoot,
+						Statement: analyzeCommand,
+					}
+
+					if _, ok := partitionRootSet[rootStatement]; !ok {
+						partitionRootSet[rootStatement] = struct{}{}
+					}
+				}
+			}
+		}
```
jimmyyih, line 478: Looks kinda weird to double loop on `filteredDataEntries` again just to extract the root partitions... but I understand the thought behind it (the `partitionRootSet` is what I'm thinking).  Just to prevent any bad refactors in the future, can you add to the comment something like this: `These particular ANALYZE statements should run last so add them to the end of the analyzeStatements list.`.

restore/restore.go
```diff
@@ -440,14 +440,48 @@ func runAnalyze(filteredDataEntries map[string][]toc.MasterDataEntry) {
...
+		partitionRootSet := map[toc.StatementWithType]struct{}{}
+		for _, dataEntries := range filteredDataEntries {
+			for _, entry := range dataEntries {
+				if entry.PartitionRoot != "" {
+					tableSchema := entry.Schema
+					if opts.RedirectSchema != "" {
+						tableSchema = opts.RedirectSchema
+					}
+					rootFQN := utils.MakeFQN(tableSchema, entry.PartitionRoot)
+					analyzeCommand := fmt.Sprintf("ANALYZE %s", rootFQN)
```
jimmyyih, line 466: Change `ANALYZE` to `ANALYZE ROOTPARTITION` as suggested by our internal query processing colleagues.  Modify the comment and commit message to reflect this as well.

restore/restore.go
```diff
@@ -440,14 +440,49 @@ func runAnalyze(filteredDataEntries map[string][]toc.MasterDataEntry) {
...
+				Name:      entry.Name,
 				Statement: analyzeCommand,
 			}
 			analyzeStatements = append(analyzeStatements, newAnalyzeStatement)
 		}
 	}
 
+	// Only GPDB 5+ has leaf partition stats merged up to the root
+	// automatically. Against GPDB 4.3, we must extract the root partitions
+	// from the leaf partition info and run ANALYZE only on the root
```
jimmyyih, line 453: Update the comment to reflect the new command.  (e.g. s/ANALYZE only/ANALYZE ROOTPARTITION/)

Update the commit message to reflect the same as well.

jimmyyih (approved):

PR #427, opened 2020-10-23
Master partition logic    Prevent partition queries from running in gpdb7

kyeap-vmware:
In GPDB 7+, partition tables have been heavily revamped to match the
upstream Postgres implementation. Due to this, many existing gpbackup
partitioning logic will simply error out due to missing catalog
functions, catalog views, and catalog tables. As a first step in
preparation for upgrading the gpbackup to support gpdb7, prevent
any queries that use the following deprecated code.

Catalog functions removed:
pg_get_partition_def
pg_get_partition_template_def

Catalog tables removed:
pg_partitions
pg_partition_columns
pg_partition_templates
pg_partition
pg_partition_rule
pg_partition_encoding

### DIFF
--- a/backup/queries_postdata.go
+++ b/backup/queries_postdata.go
@@ -80,14 +80,15 @@ func (i IndexDefinition) FQN() string {
  * e.g. comments on implicitly created indexes
  */
 func GetIndexes(connectionPool *dbconn.DBConn) []IndexDefinition {
-	resultIndexes := make([]IndexDefinition, 0)
+
+	var query string
 	if connectionPool.Version.Before("6") {
 		indexNameSet := ConstructImplicitIndexNames(connectionPool)
 		implicitIndexStr := ""
 		if indexNameSet != "" {
 			implicitIndexStr = fmt.Sprintf("OR n.nspname || '.' || ic.relname IN (%s)", indexNameSet)
 		}
-		query := fmt.Sprintf(`
+		query = fmt.Sprintf(`
 	SELECT DISTINCT i.indexrelid AS oid,
 		quote_ident(ic.relname) AS name,
 		quote_ident(n.nspname) AS owningschema,
@@ -109,12 +110,16 @@ func GetIndexes(connectionPool *dbconn.DBConn) []IndexDefinition {
 		AND NOT EXISTS (SELECT 1 FROM pg_partition_rule r WHERE r.parchildrelid = c.oid)
 		AND %s
 	ORDER BY name`,
-	implicitIndexStr, relationAndSchemaFilterClause(), ExtensionFilterClause("c"))
+			implicitIndexStr, relationAndSchemaFilterClause(), ExtensionFilterClause("c"))
 
-		err := connectionPool.Select(&resultIndexes, query)
-		gplog.FatalOnError(err)
 	} else {
-		query := fmt.Sprintf(`
+		// TODO: fix for gpdb7 partitioning
+		partitionRuleExcludeClause := ""
+		if connectionPool.Version.Before("7") {
+			partitionRuleExcludeClause = "AND NOT EXISTS (SELECT 1 FROM pg_partition_rule r WHERE r.parchildrelid = c.oid)"
+		}
+
+		query = fmt.Sprintf(`
 	SELECT DISTINCT i.indexrelid AS oid,
 		quote_ident(ic.relname) AS name,
 		quote_ident(n.nspname) AS owningschema,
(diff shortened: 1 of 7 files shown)


backup/queries_postdata.go
```diff
@@ -80,6 +80,11 @@ func (i IndexDefinition) FQN() string {
  * e.g. comments on implicitly created indexes
  */
 func GetIndexes(connectionPool *dbconn.DBConn) []IndexDefinition {
+	// TODO: fix for gpdb7 partitioning
+	if connectionPool.Version.AtLeast("7") {
+		return make([]IndexDefinition, 0)
+	}
+
```
jimmyyih, line 87: So this one is just the filter on pg_partition_rule (`AND NOT EXISTS (SELECT 1 FROM pg_partition_rule r WHERE r.parchildrelid = c.oid))?

Could we separate that out of the queries into this new GPDB version check instead of returning an empty object?

Example:
```
partitionRuleExcludeClause := ""
if connectionPool.Version.AtLeast("7") {
    partitionRuleExcludeClause = "AND NOT EXISTS (SELECT 1 FROM pg_partition_rule r WHERE r.parchildrelid = c.oid)"
}
```

backup/queries_relations.go
```diff
@@ -75,6 +75,11 @@ func (r Relation) GetUniqueID() UniqueID {
  * it is currently much simpler than the include case.
  */
 func getUserTableRelations(connectionPool *dbconn.DBConn) []Relation {
+	// TODO: fix for gpdb7 partitioning
+	if connectionPool.Version.AtLeast("7") {
+		return []Relation{}
+	}
+
```
jimmyyih, line 82: For this function, it seems the only offender is the `childPartitionFilter`.  I think you can just add the version check to the conditional check and we should be good here.

Example:
```
if !MustGetFlagBool(options.LEAD_PARTITION_DATA) && connectionPool.Version.Before("7")
```

Now that I think about it... I don't think the --leaf-partition-data flag would even work as intended against GPDB 7 with the new partitioning logic.  Essentially, --leaf-partition-data would always be true and cannot be turned off when running against GPDB 7.  That will be interesting to think about as we get the partitioning logic working.

backup/queries_shared.go
```diff
@@ -109,7 +109,17 @@ func (c Constraint) FQN() string {
 }
 
 func GetConstraints(connectionPool *dbconn.DBConn, includeTables ...Relation) []Constraint {
-	// ConIsLocal should always return true from GetConstraints because we filter out constraints that are inherited using the INHERITS clause, or inherited from a parent partition table. This field only accurately reflects constraints in GPDB6+ because check constraints on parent tables must propogate to children. For GPDB versions 5 or lower, this field will default to false.
+	// TODO: fix for gpdb7 partitioning
+	if connectionPool.Version.AtLeast("7") {
+		return []Constraint{}
+	}
```
jimmyyih, line 115: Oh... this one is a bit weirder.  For this one, I guess it's okay to leave the TODO item for now until we can think up a new query.  I'll take care of that and create a new story for this.

backup/queries_table_defs.go
```diff
@@ -195,6 +200,11 @@ var storageTypeCodes = map[string]string{
 }
 
 func GetColumnDefinitions(connectionPool *dbconn.DBConn) map[uint32][]ColumnDefinition {
+	// TODO: fix for gpdb7 partitioning
+	if connectionPool.Version.AtLeast("7") {
+		return make(map[uint32][]ColumnDefinition)
+	}
+
```
jimmyyih, line 207: Similar for here, we only need to filter out the pg_partition_rule part with something like such:
```
partitionRuleExcludeClause := ""
if connectionPool.Version.AtLeast("7") {
    partitionRuleExcludeClause = "AND NOT EXISTS (SELECT 1 FROM 
			(SELECT parchildrelid FROM pg_partition_rule EXCEPT SELECT reloid FROM pg_exttable)
			par WHERE par.parchildrelid = c.oid)"
}
```

backup/queries_table_defs.go
```diff
@@ -311,6 +321,11 @@ func GetTableReplicaIdentity(connectionPool *dbconn.DBConn) map[uint32]string {
 }
 
 func GetPartitionDetails(connectionPool *dbconn.DBConn) (map[uint32]string, map[uint32]string) {
+	// TODO: fix for gpdb7 partitioning
```
jimmyyih, line 324: We won't be touching this function ever again for GPDB 7+ so the TODO is not necessary.

backup/queries_table_defs.go
```diff
@@ -137,6 +137,11 @@ type PartitionLevelInfo struct {
 }
 
 func GetPartitionTableMap(connectionPool *dbconn.DBConn) map[uint32]PartitionLevelInfo {
+	// TODO: fix for gpdb7 partitioning
```
jimmyyih, line 140: I don't think we'll be touching this function again for GPDB 7+ but this one has some uncertainty to warrant the TODO.

backup/queries_table_defs.go
```diff
@@ -351,6 +366,11 @@ type AlteredPartitionRelation struct {
  * them.
  */
 func GetPartitionAlteredSchema(connectionPool *dbconn.DBConn) map[uint32][]AlteredPartitionRelation {
+	// TODO: fix for gpdb7 partitioning
```
jimmyyih, line 369: We won't be touching this function ever again for GPDB 7+ so the TODO is not necessary.

backup/queries_table_defs.go
```diff
@@ -195,6 +200,11 @@ var storageTypeCodes = map[string]string{
 }
 
 func GetColumnDefinitions(connectionPool *dbconn.DBConn) map[uint32][]ColumnDefinition {
+	// TODO: fix for gpdb7 partitioning
+	if connectionPool.Version.AtLeast("7") {
+		return make(map[uint32][]ColumnDefinition)
+	}
```
jimmyyih, line 206: Now that the pg_partition_rule exclusion is being handled, we no longer need this version check at the top.

jimmyyih (approved):

PR #430, opened 2020-10-29
Support GPDB 7 append optimized table DDL

kyeap-vmware:
In GPDB 7+, the usual way of creating append optimized tables has been
deprecated. We need to add a GPDB version check and use the new DDL
syntax for append optimized tables.


integration/predata_relations_create_test.go
```diff
@@ -109,6 +109,17 @@ var _ = Describe("backup integration create statement tests", func() {
 			resultTable := backup.ConstructDefinitionsForTables(connectionPool, []backup.Relation{testTable.Relation})[0]
 			structmatcher.ExpectStructsToMatchExcluding(testTable.TableDefinition, resultTable.TableDefinition, "ColumnDefs.Oid", "ExtTableDef")
 		})
+		It("creates a basic gpdb7 append-optimized table", func() {
```
jimmyyih, line 112: Should future-proof this as "GPDB 7+ append-optimized table".

jimmyyih (approved):

PR #432, opened 2020-10-30
Update GPDB 7 incremental logic for ao tables

kyeap-vmware:
In GPDB 7+, the pg_class.relstorage column no longer exists to tell us
if a table is heap or append-only. Instead, we now have the upstream
Postgres implementation of pluggable storage which introduces table (and
other relation type) access methods exposed in catalog table pg_am which
is referenced in the new pg_class.relam column.

### DIFF
--- a/backup/queries_incremental.go
+++ b/backup/queries_incremental.go
@@ -34,22 +34,44 @@ func getAllModCounts(connectionPool *dbconn.DBConn) map[string]int64 {
 }
 
 func getAOSegTableFQNs(connectionPool *dbconn.DBConn) map[string]string {
-	query := fmt.Sprintf(`
-	SELECT seg.aotablefqn,
-		'pg_aoseg.' || quote_ident(aoseg_c.relname) AS aosegtablefqn
-	FROM pg_class aoseg_c
-		JOIN (SELECT pg_ao.relid AS aooid,
-				pg_ao.segrelid,
-				aotables.aotablefqn
-			FROM pg_appendonly pg_ao
-				JOIN (SELECT c.oid,
-						quote_ident(n.nspname)|| '.' || quote_ident(c.relname) AS aotablefqn
-					FROM pg_class c
-						JOIN pg_namespace n ON c.relnamespace = n.oid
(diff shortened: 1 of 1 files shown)


backup/queries_incremental.go
```diff
@@ -86,24 +108,47 @@ func getModCount(connectionPool *dbconn.DBConn, aosegtablefqn string) int64 {
...
+					WHERE c.relstorage IN ('ao', 'co')
+					AND %s
+				) aotables
+			JOIN ( SELECT lo.objid,
+						MAX(lo.statime) AS lastddltimestamp
+					FROM pg_stat_last_operation lo
+					WHERE lo.staactionname IN ('CREATE', 'ALTER', 'TRUNCATE')
+					GROUP BY lo.objid
+				) lastop
+			ON aotables.aooid = lastop.objid`, relationAndSchemaFilterClause())
```
jimmyyih, line 150: The tabbing seems to be a little bit off compared to the other query sections above.

jimmyyih (approved):

backup/queries_incremental.go
kyeap-vmware: @jimmyyih check the new commit. Too much?

backup/queries_incremental.go
```diff
@@ -86,24 +108,73 @@ func getModCount(connectionPool *dbconn.DBConn, aosegtablefqn string) int64 {
...
+            MAX(lo.statime) AS lastddltimestamp
+         FROM
+            pg_stat_last_operation lo
+         WHERE
+            lo.staactionname IN ( 'CREATE', 'ALTER', 'TRUNCATE')
+         GROUP BY
+            lo.objid
+      ) lastop
+      ON aotables.aooid = lastop.objid`, relationAndSchemaFilterClause())
+	}
```
jimmyyih, line 177: The conditional here is setting `query` as the same query string for both conditional results.  I guess if it's not easy to quickly spot this, it may be better after all to have one single query string with `%s` replacements.

kyeap-vmware: oh oops. my mistake. must have miscopied while formatting.

PR #437, opened 2020-11-02
Query for sequence info using new catalog table

roicos:
- In GPDB 7+, sequence relations have been changed
   to no longer store all metadata in the sequence
   relation itself. Instead, there is now a catalog
   table pg_sequence which contains the majority of
   the metadata that we require.

### DIFF
--- a/backup/queries_relations.go
+++ b/backup/queries_relations.go
@@ -154,6 +154,7 @@ type Sequence struct {
 	OwningTableSchema string
 	OwningTable       string
 	OwningColumn      string
+	IsIdentity        bool
 	Definition        SequenceDefinition
 }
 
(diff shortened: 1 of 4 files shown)


backup/queries_relations.go
```diff
@@ -183,7 +184,34 @@ type SequenceDefinition struct {
...
+	FROM pg_class c
+		JOIN pg_namespace n ON n.oid = c.relnamespace
+		LEFT JOIN pg_depend d ON c.oid = d.objid AND d.deptype in ('a', 'i')
+		LEFT JOIN pg_class t ON t.oid = d.refobjid
+		LEFT JOIN pg_namespace m ON m.oid = t.relnamespace
+		LEFT JOIN pg_attribute a ON a.attrelid = d.refobjid AND a.attnum = d.refobjsubid
+	WHERE c.relkind = 'S'
+	AND %s
+	AND %s
+	ORDER BY n.nspname, c.relname`,
```
jimmyyih, line 211: The tab formatting seems to be a bit off.  Could you make sure this is tabbed similarly to the original query below?

backup/queries_relations.go
```diff
@@ -235,21 +263,36 @@ func GetAllSequences(connectionPool *dbconn.DBConn) []Sequence {
...
+		SELECT s.seqstart AS startval,
+		   r.last_value AS lastval,
+		   s.seqincrement AS increment,
+		   s.seqmax AS maxval,
+		   s.seqmin AS minval,
+		   s.seqcache AS cacheval,
+		   s.seqcycle AS iscycled,
+		   r.is_called AS iscalled
+		FROM %s r
+		 JOIN pg_sequence s ON s.seqrelid = '%s'::regclass::oid;`, seqName, seqName)
```
jimmyyih, line 278: Same thing here with the tabbing.

jimmyyih (approved):

PR #438, opened 2020-11-05
Gpbackup supports stored procedures

roicos:
- In GPDB 7+, users can now create stored procedures
   in a SQL-standard way. It's a user-defined function
   but with no return value and can do commit/rollback.
   Because of that, we only need to make changes in our
   gpbackup function handling.

### DIFF
--- a/backup/predata_functions.go
+++ b/backup/predata_functions.go
@@ -16,8 +16,12 @@ import (
 func PrintCreateFunctionStatement(metadataFile *utils.FileWithByteCount, toc *toc.TOC, funcDef Function, funcMetadata ObjectMetadata) {
 	start := metadataFile.ByteCount
 	funcFQN := utils.MakeFQN(funcDef.Schema, funcDef.Name)
-	metadataFile.MustPrintf("\n\nCREATE FUNCTION %s(%s) RETURNS ", funcFQN, funcDef.Arguments.String)
-	metadataFile.MustPrintf("%s AS", funcDef.ResultType.String)
+
+	if connectionPool.Version.AtLeast("7") && funcDef.Kind == "p" {
+		metadataFile.MustPrintf("\n\nCREATE PROCEDURE %s(%s) AS", funcFQN, funcDef.Arguments.String)
+	} else {
+		metadataFile.MustPrintf("\n\nCREATE FUNCTION %s(%s) RETURNS %s AS", funcFQN, funcDef.Arguments.String, funcDef.ResultType.String)
+	}
 	PrintFunctionBodyOrPath(metadataFile, funcDef)
 	metadataFile.MustPrintf("LANGUAGE %s", funcDef.Language)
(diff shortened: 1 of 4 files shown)


backup/predata_functions.go
```diff
@@ -56,32 +65,37 @@ func PrintFunctionModifiers(metadataFile *utils.FileWithByteCount, funcDef Funct
...
-	}
-	if funcDef.IsLeakProof {
-		metadataFile.MustPrintf(" LEAKPROOF")
-	}
 	if funcDef.IsSecurityDefiner {
 		metadataFile.MustPrintf(" SECURITY DEFINER")
 	}
+
+	if connectionPool.Version.AtLeast("7") && funcDef.Kind == "p" {
+		// TODO: TRANSFORM
```
jimmyyih, line 80: TRANSFORM is valid for both functions and procedures so it's not necessary to have an if/else here.

backup/predata_functions.go
```diff
@@ -16,12 +16,21 @@ import (
...
+		PrintFunctionModifiers(metadataFile, funcDef)
+		metadataFile.MustPrintln(";")
+	} else {
+		metadataFile.MustPrintf("\n\nCREATE FUNCTION %s(%s) RETURNS ", funcFQN, funcDef.Arguments.String)
+		metadataFile.MustPrintf("%s AS", funcDef.ResultType.String)
+		PrintFunctionBodyOrPath(metadataFile, funcDef)
+		metadataFile.MustPrintf("LANGUAGE %s", funcDef.Language)
+		PrintFunctionModifiers(metadataFile, funcDef)
+		metadataFile.MustPrintln(";")
+	}
```
jimmyyih, line 33: To prevent duplicated code, it's fine to just have `LANGUAGE %s` for procedures where it is already for functions.  Something similar to what was mentioned in the story should work fine:
```
if connectionPool.Version.AtLeast("7") && funcDef.Kind == "p" {
 	metadataFile.MustPrintf("\n\nCREATE PROCEDURE %s(%s) AS", funcFQN, funcDef.Arguments.String)
} else {
 	metadataFile.MustPrintf("\n\nCREATE FUNCTION %s(%s) RETURNS %s AS", funcFQN, funcDef.Arguments.String, funcDef.ResultType.String)
}
PrintFunctionBodyOrPath(metadataFile, funcDef)
metadataFile.MustPrintf("LANGUAGE %s", funcDef.Language)
PrintFunctionModifiers(metadataFile, funcDef)
metadataFile.MustPrintln(";")
```

backup/predata_functions.go
jimmyyih: Ah sorry, I wasn't clear enough.  The if condition for the TRANSFORM wasn't necessary but the else section that you had was okay to have.  Either way, I don't think it would have logically prevented anything except give clarity in the code on what's possible for functions and procedures... so we can merge this PR in as it is right now.

jimmyyih (approved):

PR #440, opened 2020-11-09
Planner support

roicos:
- In GPDB 7+, user-defined functions can be given extra
   information to help the planner. This is done with new
   SQL syntax SUPPORT <support function> added to the
   CREATE FUNCTION command. We need to make changes in
   Function struct, query to get function info and
   print statement to create function.

### DIFF
--- a/backup/predata_functions.go
+++ b/backup/predata_functions.go
@@ -88,6 +88,9 @@ func PrintFunctionModifiers(metadataFile *utils.FileWithByteCount, funcDef Funct
 	}
 	if connectionPool.Version.AtLeast("7") {
 		// TODO: TRANSFORM
+		if funcDef.PlannerSupport != "-" {
+			metadataFile.MustPrintf("\nSUPPORT %s", funcDef.PlannerSupport)
+		}
 	}
 	// Default cost is 1 for C and internal functions or 100 for functions in other languages
 	isInternalOrC := funcDef.Language == "c" || funcDef.Language == "internal"
--- a/integration/predata_functions_queries_test.go
+++ b/integration/predata_functions_queries_test.go
@@ -2,6 +2,7 @@ package integration
 
 import (
 	"database/sql"
+
 	"github.com/greenplum-db/gp-common-go-libs/structmatcher"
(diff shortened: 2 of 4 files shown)


backup/predata_functions.go
```diff
@@ -97,6 +97,11 @@ func PrintFunctionModifiers(metadataFile *utils.FileWithByteCount, funcDef Funct
 	if funcDef.ReturnsSet && funcDef.NumRows != 0 && funcDef.NumRows != 1000 {
 		metadataFile.MustPrintf("\nROWS %v", funcDef.NumRows)
 	}
+	if connectionPool.Version.AtLeast("7") {
+		if funcDef.PlannerSupport != "-" {
```
jimmyyih, line 101: Could consolidate this into `if connectionPool.Version.AtLeast("7") && funcDef.PlannerSupport != "-"` or maybe move the `funcDef.PlannerSupport != "-"` to the above GPDB 7+ check containing the TRANSFORM TODO item.

Also... wow, the prosupport default is "-" instead of NULL... that caught me by surprise.

integration/predata_functions_queries_test.go
```diff
@@ -442,7 +430,7 @@ SORTOP = ~>~ );`)
 			resultAggregates := backup.GetAggregates(connectionPool)
 			aggregateDef := backup.Aggregate{
 				Schema: "public", Name: "ascii_max", Arguments: sql.NullString{String: "character", Valid: true},
-				IdentArgs: sql.NullString{String: "character", Valid: true},
+				IdentArgs:          sql.NullString{String: "character", Valid: true},
```
jimmyyih, line 433: Are these refactoring changes automatically coming from Goland IDE?  They're a bit distracting when reviewing the PR.

jimmyyih (approved):
Make sure to merge in https://github.com/greenplum-db/gpbackup/pull/438 first and rebase this PR before merging.

integration/predata_functions_queries_test.go
roicos: goimports

PR #441, opened 2020-11-11
Attach GPDB 7+ child tables to their root

kyeap-vmware:
In GPDB 7+, we no longer have large DDL commands to create a partition
table and its leaf partitions. Each leaf partition must be created
individually (at the moment) and attached to the root or subroot
partition. To do this, we must query for some simple partition
information and construct the SQL command to attach the partition. This
SQL command should be printed immediately after the CREATE TABLE
statement of the leaf partition.

### DIFF
--- a/backup/predata_relations.go
+++ b/backup/predata_relations.go
@@ -236,6 +236,12 @@ func PrintPostCreateTableStatements(metadataFile *utils.FileWithByteCount, toc *
 				utils.MakeFQN(alteredPartitionRelation.OldSchema, alteredPartitionRelation.Name), alteredPartitionRelation.NewSchema))
 	}
 
+	attachInfo := table.AttachPartitionInfo
+	if (attachInfo != AttachPartitionInfo{}) {
+		statements = append(statements,
+			fmt.Sprintf("ALTER TABLE ONLY %s ATTACH PARTITION %s %s;", attachInfo.Parent, attachInfo.Relname, attachInfo.Expr))
+	}
+
 	PrintStatements(metadataFile, toc, table, statements)
 }
 
--- a/backup/queries_table_defs.go
+++ b/backup/queries_table_defs.go
@@ -74,6 +74,7 @@ type TableDefinition struct {
 	PartitionAlteredSchemas []AlteredPartitionRelation
 	AccessMethodName        string
 	PartitionKeyDef         string
+	AttachPartitionInfo     AttachPartitionInfo
 }
 
 /*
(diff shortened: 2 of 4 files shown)


backup/queries_table_defs.go
```diff
@@ -531,6 +534,38 @@ func GetTableInheritance(connectionPool *dbconn.DBConn, tables []Relation) map[u
...
+
+	query := fmt.Sprintf(`
+	SELECT
+		c.oid,
+		quote_ident(n.nspname) || '.' || quote_ident(c.relname) AS relname,
+		pg_partition_root(c.oid) AS parent,
+		pg_get_expr(relpartbound, c.oid) AS expr
+	FROM pg_class c
+		JOIN pg_namespace n ON c.relnamespace = n.oid
+	WHERE relispartition = 't'`)
```
jimmyyih, line 557: Let's use something like this for now to solve your parent FQN issue:
```
SELECT
        c.oid,
        quote_ident(n.nspname) || '.' || quote_ident(c.relname) AS relname,
        quote_ident(rn.nspname) || '.' || quote_ident(rc.relname) AS parent,
        pg_get_expr(c.relpartbound, c.oid) AS expr
FROM pg_class c
        JOIN pg_namespace n ON c.relnamespace = n.oid
        JOIN pg_class rc ON pg_partition_root(c.oid) = rc.oid
        JOIN pg_namespace rn ON rc.relnamespace = rn.oid
WHERE c.relispartition = 't';
```

integration/predata_table_defs_queries_test.go
```diff
@@ -878,4 +878,22 @@ SET SUBPARTITION TEMPLATE
 			Expect(result[oid]).To(ConsistOf(expectedAlteredPartitions))
 		})
 	})
+	Describe("GetAttachedPartitionInfo", func() {
+		It("Returns a map of table oid to attach artition info", func() {
```
jimmyyih, line 882: s/artition/partition/

integration/predata_relations_create_test.go
```diff
@@ -416,6 +416,36 @@ SET SUBPARTITION TEMPLATE ` + `
 			resultTable := backup.ConstructDefinitionsForTables(connectionPool, []backup.Relation{testTable.Relation})[0]
 			Expect(resultTable.ReplicaIdentity).To(Equal("f"))
 		})
+		It("prints a GPDB 7+ ALTER statement to ATTACH a (child) table to it's root", func() {
```
jimmyyih, line 419: No need to have child in parentheses.

backup/predata_relations.go
```diff
@@ -233,6 +233,12 @@ func PrintPostCreateTableStatements(metadataFile *utils.FileWithByteCount, toc *
 				utils.MakeFQN(alteredPartitionRelation.OldSchema, alteredPartitionRelation.Name), alteredPartitionRelation.NewSchema))
 	}
 
+	api := table.AttachPartitionInfo
+	if (api != AttachPartitionInfo{}) {
+		statements = append(statements,
+			fmt.Sprintf("ALTER TABLE ONLY %s ATTACH PARTITION %s %s;", api.Parent, api.Relname, api.Expr))
```
jimmyyih, line 239: Variable name `api` could give the wrong impression... especially when grepping for actual API stuff.  How about `attachInfo` since ATTACH is pretty specific to partitions?

jimmyyih (approved):

PR #442, opened 2020-11-12
Dump the integer type of a sequence relation

roicos:
- In GPDB 7+, you can change the integer type of a sequence
   relation from bigint to smallint or integer by specifying
   CREATE SEQUENCE AS <integer type>. We must dump the
   AS <integer type> clause if the sequence is of type
   smallint or integer.

### DIFF
--- a/backup/predata_relations.go
+++ b/backup/predata_relations.go
@@ -254,6 +254,18 @@ func PrintCreateSequenceStatements(metadataFile *utils.FileWithByteCount,
 		start := metadataFile.ByteCount
 		definition := sequence.Definition
 		metadataFile.MustPrintln("\n\nCREATE SEQUENCE", sequence.FQN())
+		if connectionPool.Version.AtLeast("7") {
+			if definition.Type != "bigint"{
+				metadataFile.MustPrintln("\tAS", definition.Type)
+			}
+			if definition.Type == "smallint" {
+				maxVal = int64(math.MaxInt16)
+				minVal = int64(math.MinInt16)
+			} else if definition.Type == "integer" {
+				maxVal = int64(math.MaxInt32)
+				minVal = int64(math.MinInt32)
+			}
+		}
 		if connectionPool.Version.AtLeast("6") {
 			metadataFile.MustPrintln("\tSTART WITH", definition.StartVal)
 		} else if !definition.IsCalled {
--- a/backup/predata_relations_other_test.go
+++ b/backup/predata_relations_other_test.go
@@ -191,6 +191,44 @@ SELECT pg_catalog.setval('public.seq_name', 7, true);`, getSeqDefReplace()),
 				`REVOKE ALL ON SEQUENCE public.seq_name FROM PUBLIC;
 GRANT SELECT,USAGE ON SEQUENCE public.seq_name TO testrole WITH GRANT OPTION;`)
 		})
+		It("prints data_type of the sequence", func() {
+			if connectionPool.Version.Before("7") {
+				Skip("Test only applicable to GPDB 7 and above")
+			}
+			seqSmallInt := backup.Sequence{Relation: baseSequence, Definition: backup.SequenceDefinition{LastVal: 10, Type: "smallint", Increment: 2, MaxVal: math.MaxInt16, MinVal: 1, CacheVal: 5, LogCnt: 42, IsCycled: false, IsCalled: true}}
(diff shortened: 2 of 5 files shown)


backup/predata_relations.go
```diff
@@ -254,6 +254,16 @@ func PrintCreateSequenceStatements(metadataFile *utils.FileWithByteCount,
 		start := metadataFile.ByteCount
 		definition := sequence.Definition
 		metadataFile.MustPrintln("\n\nCREATE SEQUENCE", sequence.FQN())
+		if connectionPool.Version.AtLeast("7") {
+			metadataFile.MustPrintln("\tAS", definition.Type)
```
jimmyyih, line 258: `pg_dump` doesn't print this for default data type `bigint`.  We should probably keep that parity by doing the same.

backup/predata_relations_other_test.go
```diff
@@ -191,6 +191,44 @@ SELECT pg_catalog.setval('public.seq_name', 7, true);`, getSeqDefReplace()),
 				`REVOKE ALL ON SEQUENCE public.seq_name FROM PUBLIC;
 GRANT SELECT,USAGE ON SEQUENCE public.seq_name TO testrole WITH GRANT OPTION;`)
 		})
+		It("prints data_type of the sequence", func() {
+			if connectionPool.Version.Before("7.0.0") {
+				Skip("Test only applicable to GPDB 7.0.0 and above")
```
jimmyyih, line 196: Can change 7.0.0 to just 7 here for both instances.  Most greps will be for `Before("7")`.

backup/predata_relations_other_test.go
```diff
@@ -191,6 +191,44 @@ SELECT pg_catalog.setval('public.seq_name', 7, true);`, getSeqDefReplace()),
...
+			}
+			seqSmallInt := backup.Sequence{Relation: baseSequence, Definition: backup.SequenceDefinition{LastVal: 10, Type: "smallint", Increment: 2, MaxVal: math.MaxInt16, MinVal: 1, CacheVal: 5, LogCnt: 42, IsCycled: false, IsCalled: true}}
+			seqInteger := backup.Sequence{Relation: baseSequence, Definition: backup.SequenceDefinition{LastVal: 10, Type: "integer", Increment: 2, MaxVal: math.MaxInt32, MinVal: 1, CacheVal: 5, LogCnt: 42, IsCycled: false, IsCalled: true}}
+			seqBigInt := backup.Sequence{Relation: baseSequence, Definition: backup.SequenceDefinition{LastVal: 10, Type: "smallint", Increment: 2, MaxVal: math.MaxInt64, MinVal: 1, CacheVal: 5, LogCnt: 42, IsCycled: false, IsCalled: true}}
```
jimmyyih, line 200: Type parameter should show `Type: "bigint"`.

backup/predata_relations_other_test.go
```diff
@@ -191,6 +191,44 @@ SELECT pg_catalog.setval('public.seq_name', 7, true);`, getSeqDefReplace()),
...
+
+				fmt.Sprintf(`CREATE SEQUENCE public.seq_name%s AS integer
+INCREMENT BY 2
+NO MAXVALUE
+NO MINVALUE
+CACHE 5;
+
+SELECT pg_catalog.setval('public.seq_name', 10, true);`, getSeqDefReplace()),
+
+				fmt.Sprintf(`CREATE SEQUENCE public.seq_name%s AS bigint
```
jimmyyih, line 222: If we follow `pg_dump` parity, we should not have `AS bigint` for this CREATE SEQUENCE command.

backup/queries_relations.go
```diff
@@ -268,14 +269,20 @@ func GetSequenceDefinition(connectionPool *dbconn.DBConn, seqName string) Sequen
...
 			s.seqincrement AS increment,
 			s.seqmax AS maxval,
 			s.seqmin AS minval,
 			s.seqcache AS cacheval,
 			s.seqcycle AS iscycled,
 			r.is_called AS iscalled
 		FROM %s r
-		JOIN pg_sequence s ON s.seqrelid = '%s'::regclass::oid;`, seqName, seqName)
+		JOIN pg_sequence s ON s.seqrelid = '%s'::regclass::oid
+		JOIN pg_type t on s.seqtypid = t.oid;`, seqName, seqName)
```
jimmyyih, line 285: Let's use `pg_catalog.format_type(s.seqtypid, NULL) AS type` here like in `pg_dump` instead to avoid the JOIN.

jimmyyih (approved):

PR #445, opened 2020-11-12
Fix gprestore GetExistingTableFQNs() function

roicos:
- In the query to get all tables we were not
  filtering properly on pg_class.relkind. It
  should depend on gpdb version.

- Also pg_class.relstorage is removed in gpdb7
  and was not actually used in gpdb versions below
  so it should be removed.

- Also namespace filter was changed to include
  all pg_ and gp_ namespaces.

### DIFF
--- a/restore/wrappers.go
+++ b/restore/wrappers.go
@@ -343,18 +343,23 @@ func RestoreSchemas(schemaStatements []toc.StatementWithType, progressBar utils.
 
 func GetExistingTableFQNs() ([]string, error) {
 	existingTableFQNs := make([]string, 0)
+	var relkindFilter string
 
-	// Note that 'f' for foreign tables only matters for GPDB 6+ but we shouldn't need a GPDB
-	// version check on this since this is catalog and 'f' is new starting from GPDB 6+.
-	query := `SELECT quote_ident(n.nspname) || '.' || quote_ident(c.relname)
+	if connectionPool.Version.Before("6") {
+		relkindFilter = "'r', 'S'"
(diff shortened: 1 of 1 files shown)


restore/wrappers.go
```diff
@@ -346,15 +346,23 @@ func GetExistingTableFQNs() ([]string, error) {
 
 	// Note that 'f' for foreign tables only matters for GPDB 6+ but we shouldn't need a GPDB
 	// version check on this since this is catalog and 'f' is new starting from GPDB 6+.
-	query := `SELECT quote_ident(n.nspname) || '.' || quote_ident(c.relname)
+	var relkindFilter string
+
+	if connectionPool.Version.Before("6") {
+		relkindFilter = "'r', 'S'"
+	} else if connectionPool.Version.AtLeast("6") {
```
jimmyyih, line 353: The next `AtLeast("7")` condition would never hit due to this line.  Can probably change it to `connectionPool.Version.Is("6")`.

restore/wrappers.go
```diff
@@ -346,15 +346,23 @@ func GetExistingTableFQNs() ([]string, error) {
 
 	// Note that 'f' for foreign tables only matters for GPDB 6+ but we shouldn't need a GPDB
 	// version check on this since this is catalog and 'f' is new starting from GPDB 6+.
```
jimmyyih, line 348: This comment can be removed since the relkind is being version checked in this patch.

shivzone (approved):

jimmyyih (approved):

Issue #434, opened 2020-10-30
What operations will gpbackup block


cobolbaby:
When I execute `gpbackup`, I found that the truncate operation of table is blocked until the gpbackup is finished. 

Will the truncate of the temporary table be blocked? What other operations will be blocked? 

If the backup takes a long time, the blocked SQL will have a great impact on online biz.

![2020-10-30 18-18-22屏幕截图](https://user-images.githubusercontent.com/11460848/97693567-52d8cf80-1adc-11eb-994d-b4298d3202c4.png)

cobolbaby:
It seems that I have meet the deadlock.

jimmyyih:
When this happens again, could you please run the following queries and paste the output here?
```
SELECT * FROM pg_stat_activity;
SELECT * FROM pg_locks;
```

Also, could you provide the GPDB version and gpbackup command used?  Tracing the $MASTER_DATA_DIRECTORY/pg_log/ to see the order of events would be good too (e.g. the LOCK TABLE (from gpbackup), TRUNCATE , and COPY (from gpbackup) operations on the tables that are blocked).

jimmyyih:
> Will the truncate of the temporary table be blocked? What other operations will be blocked?

Are the TRUNCATE operations only happening on temp tables?  Are they actually temp tables (e.g. CREATE TEMP TABLE) or staging tables?  I don't think the gpbackup utility touches temp tables.

jimmyyih:
Okay, I’ve RCA’d the issue and created a small repro:
```
1: CREATE TABLE foobar (a int);
1: BEGIN;
1: LOCK TABLE foobar IN ACCESS SHARE MODE;

2: TRUNCATE foobar; -- blocking expected
3: COPY foobar TO '/tmp/tempdir/foobar_<SEGID>' ON SEGMENT; -- blocking unexpected
```
[...]

PR #446, opened 2020-11-13
Backup worker threads to take accessshare locks

shivzone:
Each jobs worker thread now gets its own set of AccessShareLocks on the target relations.
This would guarantee each COPY command to not get blocked by other concurrent requests that required AccessExclusiveLock.

### DIFF
--- a/backup/queries_relations.go
+++ b/backup/queries_relations.go
@@ -9,6 +9,7 @@ import (
 	"database/sql"
 	"fmt"
 	"strings"
+	"sync"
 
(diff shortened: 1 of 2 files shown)


roicos (approved):

jimmyyih (approved):
Could you add this snippet to the bottom of the commit message:
> This mostly fixes https://github.com/greenplum-db/gpbackup/issues/434 reported by github user cobolbaby by substantially reducing the deadlock window to only this beginning locking step. More improvements to come in the future.

PR #450, opened 2020-11-18
Convert CASE unnest() query logic into LEFT JOIN LATERAL for GPDB 7+

jimmyyih:
In GPDB 7+, set-returning functions are no longer allowed inside CASE
and COALESCE statements. Because of this, we must convert our CASE
unnest() query logic into LEFT JOIN LATERAL when running against GPDB
7+ to conserve gpbackup code logic.

One thing to note is that GPDB 6 also has LATERAL joins available but
we opt not to use it when running against GPDB 6 because the CASE
unnest() query logic is much more performant than the LEFT JOIN
LATERAL alternative (~2x performance regression from my small local
testing but probably less than that in an actual production
environment).

Postgres Reference:
https://github.com/postgres/postgres/commit/0436f6bde8848b7135f19dd7f8548b8c2ae89a34

With this PR, gpbackup should no longer error out against GPDB 7 for catalog differences (when using simple bare minimum flags).

### DIFF
--- a/backup/predata_externals.go
+++ b/backup/predata_externals.go
@@ -7,6 +7,7 @@ package backup
  */
 
 import (
+	"database/sql"
 	"fmt"
 	"strings"
 
@@ -32,7 +33,7 @@ type ExternalTableDefinition struct {
 	Oid             uint32
 	Type            int
 	Protocol        int
-	Location        string
+	Location        sql.NullString
 	ExecLocation    string
 	FormatType      string
 	FormatOpts      string
@@ -75,7 +76,7 @@ func DetermineExternalTableCharacteristics(extTableDef ExternalTableDefinition) 
 	isWritable := extTableDef.Writable
 	var tableType int
 	tableProtocol := -1
-	if extTableDef.Location == "" { // EXTERNAL WEB tables may have EXECUTE instead of LOCATION
+	if !extTableDef.Location.Valid { // EXTERNAL WEB tables may have EXECUTE instead of LOCATION
 		tableProtocol = HTTP
 		if isWritable {
 			tableType = WRITABLE_WEB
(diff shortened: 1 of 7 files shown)


roicos (approved):

kyeap-vmware (approved):

backup/predata_externals.go
```diff
@@ -32,7 +33,7 @@ type ExternalTableDefinition struct {
 	Oid             uint32
 	Type            int
 	Protocol        int
-	Location        string
+	Location        sql.NullString
```
shivzone, line 36: +1 sql.NullString

backup/queries_acl.go
```diff
@@ -137,11 +137,21 @@ func GetMetadataForObjectType(connectionPool *dbconn.DBConn, params MetadataQuer
...
 	aclCols := "''"
 	kindCol := "''"
+	aclLateralJoin := ""
 	if params.ACLField != "" {
-		aclCols = fmt.Sprintf(`CASE
-		WHEN %[1]s IS NULL THEN NULL
-		WHEN array_upper(%[1]s, 1) = 0 THEN %[1]s[0]
-		ELSE unnest(%[1]s) END`, params.ACLField)
+		// Cannot use unnest() in CASE statements anymore in GPDB 7+
+		// so convert it to a LEFT JOIN LATERAL.
```
shivzone, line 143: Just so that we don't lose context, might be worth putting a note about potential run time difference between the two approaches.

shivzone (approved):

backup/queries_acl.go
jimmyyih: There's already a note in the commit message but sure, I'll add it to the code comments too just to make sure.

PR #452, opened 2020-11-24
Prevent orphaned gpbackup_helper race condition

kyeap-vmware:
Previously, a well-timed SIGTERM can orphan gpbackup_helpers if cleanup
happens before gpbackup_helpers are started. Now a mutex for starting
and stopping gpbackup_helpers prevents this race condition. Before
gpbackup_helpers are started we now check if a SIGTERM was caught.

### DIFF
--- a/utils/agent_remote.go
+++ b/utils/agent_remote.go
@@ -5,6 +5,7 @@ import (
 	"io"
 	path "path/filepath"
 	"strings"
+	"sync"
 
 	"github.com/greenplum-db/gp-common-go-libs/cluster"
 	"github.com/greenplum-db/gp-common-go-libs/gplog"
@@ -13,6 +14,8 @@ import (
 	"github.com/greenplum-db/gpbackup/filepath"
 	"github.com/pkg/errors"
 )
+
+var helperMutex sync.Mutex
 
 /*
  * Functions to run commands on entire cluster during both backup and restore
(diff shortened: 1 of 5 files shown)


utils/agent_remote.go
```diff
@@ -164,6 +176,9 @@ func CleanUpHelperFilesOnAllHosts(c *cluster.Cluster, fpInfo filepath.FilePathIn
 }
 
 func CleanUpSegmentHelperProcesses(c *cluster.Cluster, fpInfo filepath.FilePathInfo, operation string) {
+	helperMutex.Lock()
```
jimmyyih, line 179: This section could block forever if the main process gets the `helperMutex` lock first and is waiting in the `select {}` pause.  It may be a good idea to explore releasing the lock before infinite pausing.

utils/agent_remote.go
```diff
@@ -111,7 +114,17 @@ func VerifyHelperVersionOnSegments(version string, c *cluster.Cluster) {
...
 }
 
-func StartGpbackupHelpers(c *cluster.Cluster, fpInfo filepath.FilePathInfo, operation string, pluginConfigFile string, compressStr string, onErrorContinue bool, isFilter bool) {
+func StartGpbackupHelpers(c *cluster.Cluster, fpInfo filepath.FilePathInfo, operation string, pluginConfigFile string, compressStr string, onErrorContinue bool, isFilter bool, wasTerminated *bool) {
+	// A mutex lock for cleaning up and starting gpbackup helpers prevents a
+	// race condition that causes gpbackup_helpers to be orphaned if
+	// gpbackup_helper cleanup happens before they are started.
+	helperMutex.Lock()
+	if *wasTerminated {
+		helperMutex.Unlock()
```
jimmyyih, line 123: After thinking about this a bit, placing the extra unlock here is definitely the correct choice.  Moving the termination conditional check out of the locked section could have caused orphan gpbackup_helper processes in an unlikely-to-happen-in-production-but-possible edge case (`wasTerminated` is false on the check but becomes true afterwords and somehow `CleanUpSegmentHelperProcesses()` takes the mutex lock first before `StarGpbackupHelpers()`).

jimmyyih (approved):

jimmyyih:
Note that no test additions are needed since the SIGTERM end_to_end tests already cover this patch.  Those tests were the ones that revealed this issue on the CI where the test job was not completing due to orphaned gpbackup_helper processes existing.

PR #456, opened 2020-12-08
Support for generated columns

shivzone:
In GPDB7+, generated columns are supported which is a special column
that is always computed from other columns. There are
two types of generated columsn - stored and virtual. As of now stored is
the only type supported and this would occupy storage.

Postgres Reference:
postgres/postgres@fc22b66

### DIFF
--- a/backup/predata_relations.go
+++ b/backup/predata_relations.go
@@ -167,7 +167,11 @@ func printColumnDefinitions(metadataFile *utils.FileWithByteCount, columnDefs []
 			line += fmt.Sprintf(" COLLATE %s", column.Collation)
 		}
 		if column.HasDefault {
-			line += fmt.Sprintf(" DEFAULT %s", column.DefaultVal)
+			if column.AttGenerated != ""  {
+				line += fmt.Sprintf(" GENERATED ALWAYS AS %s %s", column.DefaultVal, column.AttGenerated)
+			} else {
+				line += fmt.Sprintf(" DEFAULT %s", column.DefaultVal)
+			}
 		}
 		if column.NotNull {
 			line += " NOT NULL"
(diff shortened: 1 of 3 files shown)


backup/predata_relations.go
```diff
@@ -167,7 +167,11 @@ func printColumnDefinitions(metadataFile *utils.FileWithByteCount, columnDefs []
 			line += fmt.Sprintf(" COLLATE %s", column.Collation)
 		}
 		if column.HasDefault {
-			line += fmt.Sprintf(" DEFAULT %s", column.DefaultVal)
+			if column.AttGenerated != ""  {
+				line += fmt.Sprintf(" GENERATED ALWAYS AS %s %s", column.DefaultVal, column.AttGenerated)
```
jimmyyih, line 171: I think you need parenthesis around the default value as `GENERATED ALWAYS AS (%s) %s`.

From postgres doc:
```
GENERATED ALWAYS AS ( generation_expr ) STORED
```

shivzone: the value itself has the paranthesis, adding the extra wrapping brackets woudl result in `(( generation_expr ))`

jimmyyih: Ah, okay I see.

jimmyyih (approved):
Looks good.  Make sure to squash before merging.

backup/queries_table_defs.go
```diff
@@ -295,6 +302,9 @@ func GetColumnDefinitions(connectionPool *dbconn.DBConn) map[uint32][]ColumnDefi
 	resultMap := make(map[uint32][]ColumnDefinition)
 	for _, result := range results {
 		result.StorageType = storageTypeCodes[result.StorageType]
+		if connectionPool.Version.AtLeast("7") {
+			result.AttGenerated = attGeneratedCodes[result.AttGenerated]
```
roicos, line 306: Should we error out if att is not in the map?

PR #459, opened 2020-12-10
Gpbackup should dump explicitly declared AGGREGATE finalfunc behavior

roicos:
In GPDB 7+, users will be able to explicitly declare the finalfunc behavior 
for user-defined aggregates. To support this we need to add aggkind, 
aggfinalmodify and aggmfinalmodify fields to Aggregate struct,  change 
the query to getAggregates() and the query we dump in PrintCreateAggregateStatement().

### DIFF
--- a/backup/predata_functions.go
+++ b/backup/predata_functions.go
@@ -160,8 +160,14 @@ func PrintCreateAggregateStatement(metadataFile *utils.FileWithByteCount, toc *t
 	if aggDef.SortOperator != "" {
 		metadataFile.MustPrintf(",\n\tSORTOP = %s.\"%s\"", aggDef.SortOperatorSchema, aggDef.SortOperator)
 	}
-	if aggDef.Hypothetical {
-		metadataFile.MustPrintf(",\n\tHYPOTHETICAL")
+	if connectionPool.Version.Before("7") {
+		if aggDef.Hypothetical {
+			metadataFile.MustPrintf(",\n\tHYPOTHETICAL")
+		}
+	} else {
+		if aggDef.Kind == "h" {
+			metadataFile.MustPrintf(",\n\tHYPOTHETICAL")
+		}
 	}
 	if aggDef.MTransitionFunction != 0 {
 		metadataFile.MustPrintf(",\n\tMSFUNC = %s", funcInfoMap[aggDef.MTransitionFunction].QualifiedName)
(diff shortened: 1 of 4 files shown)


backup/predata_functions.go
```diff
@@ -171,6 +180,43 @@ func PrintCreateAggregateStatement(metadataFile *utils.FileWithByteCount, toc *t
...
 	}
+
+	if connectionPool.Version.AtLeast("7") {
+		var defaultFinalModify string
+		if aggDef.Kind == "n" {
+			defaultFinalModify = "r"
+		} else {
+			defaultFinalModify = "w"
+		}
+		if aggDef.Finalmodify == "" {
```
shivzone, line 191: In the c version we were checking for `== '0' `. Just checking if this is the right check

shivzone (approved):

backup/predata_functions.go
roicos: I haven't seen any mentions of '0' in Postgress code.
From what I see there is a default value and it can only be 'r', 's' or 'w'.
```
/* tells whether aggfinalfn modifies transition state */
     char        aggfinalmodify BKI_DEFAULT(r);
```

```
/*
  * Symbolic values for aggfinalmodify and aggmfinalmodify columns.
  * Preferably, finalfns do not modify the transition state value at all,
  * but in some cases that would cost too much performance.  We distinguish
  * "pure read only" and "trashes it arbitrarily" cases, as well as the
  * intermediate case where multiple finalfn calls are allowed but the
  * transfn cannot be applied anymore after the first finalfn call.
  */
 #define AGGMODIFY_READ_ONLY         'r'
 #define AGGMODIFY_SHAREABLE         's'
 #define AGGMODIFY_READ_WRITE        'w'
```
'0' in pg_dump was for old versions. In our case it's a string with the empty value of "", not "0".

shivzone: Got it

backup/predata_functions_test.go
```diff
@@ -540,6 +547,38 @@ $_$`)
...
+			Entry("kind: o or h, aggfinalmodify: s", "o", "s", "", ",\n\tFINALFUNC_MODIFY = SHARABLE"),
+			Entry("kind: o or h, aggfinalmodify: w", "o", "w", "", ""), // default, don't print
+
+			Entry("kind: n, aggmfinalmodify: r", "n", "", "r", ""), // default, don't print
+			Entry("kind: n, aggmfinalmodify: s", "n", "", "s", ",\n\tMFINALFUNC_MODIFY = SHARABLE"),
+			Entry("kind: n, aggmfinalmodify: w", "n", "", "w", ",\n\tMFINALFUNC_MODIFY = READ_WRITE"),
+			Entry("kind: o or h, aggmfinalmodify: r", "o", "", "r", ",\n\tMFINALFUNC_MODIFY = READ_ONLY"),
+			Entry("kind: o or h, aggmfinalmodify: s", "o", "", "s", ",\n\tMFINALFUNC_MODIFY = SHARABLE"),
+			Entry("kind: o or h, aggmfinalmodify: w", "o", "", "w", ""), // default, don't print
+		)
```
jimmyyih, line 581: Interesting, I never looked at what other things ginkgo provided.  Could you give a quick explanation on how this works as a reply to this comment? (No need to write a comment block in the code; I'm just curious)

jimmyyih (approved):

backup/predata_functions_test.go
roicos: With [Table testing extension](https://godoc.org/github.com/onsi/ginkgo/extensions/table) you can write just one test in a function which is second parameter o DescribeTable and then provide sets of arguments as Entries. For each entry your test will run separately, so you will be able to test all combinations you need without writing a separate test for every single one.

PR #465, opened 2020-12-16
ATTACH partition indexes to its parent index

kyeap-vmware:
In GPDB 7+, partition tables have been heavily changed. Along with the
partition tables themselves, the partition indexes are also now
structured differently. We now must attach partition indexes to parent
partition index. Topological sorting of indexes is done to ensure
that a parent index is printed before its child partition indexes.

### DIFF
--- a/backup/postdata.go
+++ b/backup/postdata.go
@@ -22,10 +22,14 @@ func PrintCreateIndexStatements(metadataFile *utils.FileWithByteCount, toc *toc.
 			metadataFile.MustPrintf("\n\n%s;", index.Def.String)
 			toc.AddMetadataEntry(section, entry, start, metadataFile.ByteCount)
 
-			indexFQN := utils.MakeFQN(index.OwningSchema, index.Name)
 			if index.Tablespace != "" {
 				start := metadataFile.ByteCount
-				metadataFile.MustPrintf("\nALTER INDEX %s SET TABLESPACE %s;", indexFQN, index.Tablespace)
+				metadataFile.MustPrintf("\nALTER INDEX %s SET TABLESPACE %s;", index.FQN(), index.Tablespace)
+				toc.AddMetadataEntry(section, entry, start, metadataFile.ByteCount)
+			}
+			if index.ParentIndexFQN != "" && connectionPool.Version.AtLeast("7") {
+				start := metadataFile.ByteCount
+				metadataFile.MustPrintf("\nALTER INDEX %s ATTACH PARTITION %s;", index.ParentIndexFQN, index.FQN())
 				toc.AddMetadataEntry(section, entry, start, metadataFile.ByteCount)
 			}
 			tableFQN := utils.MakeFQN(index.OwningSchema, index.OwningTable)
--- a/backup/postdata_test.go
+++ b/backup/postdata_test.go
@@ -58,6 +58,21 @@ var _ = Describe("backup/postdata tests", func() {
 				"ALTER TABLE public.testtable REPLICA IDENTITY USING INDEX testindex;",
 			)
 		})
+		It("can print an index of a partition that is attached to a parent partition index", func() {
+			testutils.SkipIfBefore7(connectionPool)
+			parentIndex := backup.IndexDefinition{Oid: 2, Name: "parenttestindex", OwningSchema: "public", OwningTable: "parenttesttable", Def: sql.NullString{String: "CREATE INDEX parenttestindex ON public.parenttesttable USING btree(i)", Valid: true}}
+			index.ParentIndexFQN = "public.parenttestindex"
+			indexes := []backup.IndexDefinition{parentIndex, index}
+			backup.PrintCreateIndexStatements(backupfile, tocfile, indexes, emptyMetadataMap)
+			testutils.ExpectEntry(tocfile.PostdataEntries, 0, "public", "public.parenttesttable", "parenttestindex", "INDEX")
+			testutils.ExpectEntry(tocfile.PostdataEntries, 1, "public", "public.testtable", "testindex", "INDEX")
+			testutils.ExpectEntry(tocfile.PostdataEntries, 2, "public", "public.testtable", "testindex", "INDEX")
+			testutils.AssertBufferContents(tocfile.PostdataEntries, buffer,
+				"CREATE INDEX parenttestindex ON public.parenttesttable USING btree(i);",
+				"CREATE INDEX testindex ON public.testtable USING btree(i);",
+				"ALTER INDEX public.parenttestindex ATTACH PARTITION public.testindex;",
+			)
+		})
 	})
 	Context("PrintCreateRuleStatements", func() {
 		rule := backup.RuleDefinition{Oid: 1, Name: "testrule", OwningSchema: "public", OwningTable: "testtable", Def: sql.NullString{String: "CREATE RULE update_notify AS ON UPDATE TO testtable DO NOTIFY testtable;", Valid: true}}
--- a/backup/queries_postdata.go
+++ b/backup/queries_postdata.go
@@ -51,6 +51,8 @@ type IndexDefinition struct {
 	IsClustered        bool
 	SupportsConstraint bool
 	IsReplicaIdentity  bool
+	ParentIndex        uint32
+	ParentIndexFQN     string
 }
 
 func (i IndexDefinition) GetMetadataEntry() (string, toc.MetadataEntry) {
@@ -80,7 +82,6 @@ func (i IndexDefinition) FQN() string {
  * e.g. comments on implicitly created indexes
  */
 func GetIndexes(connectionPool *dbconn.DBConn) []IndexDefinition {
-
 	var query string
 	if connectionPool.Version.Before("6") {
 		indexNameSet := ConstructImplicitIndexNames(connectionPool)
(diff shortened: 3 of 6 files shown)


backup/postdata.go
```diff
@@ -20,10 +20,14 @@ func PrintCreateIndexStatements(metadataFile *utils.FileWithByteCount, toc *toc.
...
 			toc.AddMetadataEntry(section, entry, start, metadataFile.ByteCount)
 
-			indexFQN := utils.MakeFQN(index.OwningSchema, index.Name)
 			if index.Tablespace != "" {
 				start := metadataFile.ByteCount
-				metadataFile.MustPrintf("\nALTER INDEX %s SET TABLESPACE %s;", indexFQN, index.Tablespace)
+				metadataFile.MustPrintf("\nALTER INDEX %s SET TABLESPACE %s;", index.FQN(), index.Tablespace)
+				toc.AddMetadataEntry(section, entry, start, metadataFile.ByteCount)
+			}
+			if index.ParentIndexFQN != "" {
```
jimmyyih, line 28: Although it should be okay, it might be good to add a GPDB 7+ check here just in case and to make it more obvious to others.

backup/postdata_test.go
```diff
@@ -57,6 +57,20 @@ var _ = Describe("backup/postdata tests", func() {
 				"ALTER TABLE public.testtable REPLICA IDENTITY USING INDEX testindex;",
 			)
 		})
+		It("can print an index of a partition that is attached to a parent partition index", func() {
```
jimmyyih, line 60: If you do go with my suggestion of adding the GPDB version check in `PrintCreateIndexStatements()`, you'll have to add a version check to this unit test as well.

backup/queries_table_defs.go
```diff
@@ -604,7 +604,7 @@ func GetAttachPartitionInfo(connectionPool *dbconn.DBConn) map[uint32]AttachPart
 		c.oid,
 		quote_ident(n.nspname) || '.' || quote_ident(c.relname) AS relname,
 		quote_ident(rn.nspname) || '.' || quote_ident(rc.relname) AS parent,
-		pg_get_expr(c.relpartbound, c.oid) AS expr
+		coalesce(pg_get_expr(c.relpartbound, c.oid), '') AS expr
```
jimmyyih, line 607: What scenario would this be blank?

integration/postdata_create_test.go
```diff
@@ -115,6 +115,31 @@ var _ = Describe("backup integration create statement tests", func() {
 			resultIndex := resultIndexes[0]
 			structmatcher.ExpectStructsToMatchExcluding(&resultIndex, &index, "Oid")
 		})
+		It("creates a parition index and attaches it to the parent index", func() {
+			testutils.SkipIfBefore7(connectionPool)
+
+			testhelper.AssertQueryRuns(connectionPool, "CREATE TABLE public.foopart_new (a integer, b integer) PARTITION BY RANGE (b) DISTRIBUTED BY (a)")
+			defer testhelper.AssertQueryRuns(connectionPool, "DROP TABLE public.foopart_new")
+			testhelper.AssertQueryRuns(connectionPool, "CREATE TABLE public.foopart_new_p1 (a integer, b integer) DISTRIBUTED BY (a); ALTER TABLE ONLY public.foopart_new ATTACH PARTITION public.foopart_new_p1 FOR VALUES FROM (0) TO (1);")
+			defer testhelper.AssertQueryRuns(connectionPool, "DROP TABLE public.foopart_new_p1")
```
jimmyyih, line 124: You don't need this DROP TABLE since you already have the DROP TABLE on the root partition.

integration/postdata_create_test.go
```diff
@@ -115,6 +115,31 @@ var _ = Describe("backup integration create statement tests", func() {
...
+			testhelper.AssertQueryRuns(connectionPool, "CREATE INDEX fooidx ON ONLY public.foopart_new USING btree (b)")
```
jimmyyih, line 125: You need to remove the ONLY part or else it will only create the index on the root partition.

integration/postdata_create_test.go
```diff
@@ -115,6 +115,31 @@ var _ = Describe("backup integration create statement tests", func() {
...
+			defer testhelper.AssertQueryRuns(connectionPool, "DROP INDEX public.fooidx")
```
jimmyyih, line 126: You don't need to DROP INDEX.  The DROP TABLE does it automatically.

integration/postdata_queries_test.go
```diff
@@ -187,6 +187,31 @@ PARTITION BY RANGE (date)
 
 			structmatcher.ExpectStructsToMatchExcluding(&index1, &results[0], "Oid")
 		})
+		It("returns a sorted slice of partition indexes ", func() {
+			testutils.SkipIfBefore7(connectionPool)
+			testhelper.AssertQueryRuns(connectionPool, "CREATE TABLE public.foopart_new (a integer, b integer) PARTITION BY RANGE (b) DISTRIBUTED BY (a)")
+			defer testhelper.AssertQueryRuns(connectionPool, "DROP TABLE public.foopart_new")
+			testhelper.AssertQueryRuns(connectionPool, "CREATE TABLE public.foopart_new_p1 (a integer, b integer) DISTRIBUTED BY (a); ALTER TABLE ONLY public.foopart_new ATTACH PARTITION public.foopart_new_p1 FOR VALUES FROM (0) TO (1);")
+			defer testhelper.AssertQueryRuns(connectionPool, "DROP TABLE public.foopart_new_p1")
```
jimmyyih, line 195: You don't need this DROP TABLE since you already have the DROP TABLE on the root partition.

integration/postdata_queries_test.go
```diff
@@ -187,6 +187,31 @@ PARTITION BY RANGE (date)
...
+			testhelper.AssertQueryRuns(connectionPool, "CREATE INDEX fooidx ON ONLY public.foopart_new USING btree (b)")
+			defer testhelper.AssertQueryRuns(connectionPool, "DROP INDEX public.fooidx")
```
jimmyyih, line 197: You don't need to DROP INDEX. The DROP TABLE does it automatically.

backup/queries_postdata.go
```diff
@@ -142,29 +137,106 @@ func GetIndexes(connectionPool *dbconn.DBConn) []IndexDefinition {
...
+			LEFT JOIN pg_constraint con ON i.indexrelid = con.conindid
+			LEFT JOIN pg_catalog.pg_inherits inh ON inh.inhrelid = i.indexrelid
+		WHERE i.indisvalid
+			AND i.indisready
+			AND i.indisprimary = 'f'
+			AND i.indexrelid >= %d
+			AND %s
+		ORDER BY name`,
+			FIRST_NORMAL_OBJECT_ID,
+			ExtensionFilterClause("c"))
```
jimmyyih, line 174: Why is `relationAndSchemaFilterClause()` left out for the GPDB 7 query?

backup/queries_postdata.go
```diff
@@ -142,29 +137,106 @@ func GetIndexes(connectionPool *dbconn.DBConn) []IndexDefinition {
...
+			}
+		}
+
+		// "Pop" indexes found by DFS
+		for i := len(stack) - 1; i >= 0; i-- {
+			indexOid := stack[i]
+			popIndex := indexMap[indexOid]
+			if popIndex.ParentIndex != 0 {
+				// Preprocess parent index FQN for GPDB 7+ partition indexes
+				popIndex.ParentIndexFQN = indexMap[popIndex.ParentIndex].FQN()
```
jimmyyih, line 232: Nice use of this map to avoid doing extra JOINs in the index query. 👍

integration/postdata_create_test.go
roicos: Don't we need CASCADE for this to work?

jimmyyih: CASCADE is not necessary.

Example:
```
postgres=# CREATE TABLE foobar (a int);
CREATE TABLE
postgres=# CREATE INDEX fooidx ON foobar USING btree(a);
CREATE INDEX
postgres=# SELECT count(*) FROM pg_index WHERE indexrelid = 'fooidx'::regclass;
 count
-------
     1
(1 row)

postgres=# DROP TABLE foobar;
DROP TABLE
postgres=# SELECT count(*) FROM pg_index WHERE indexrelid = 'fooidx'::regclass;
ERROR:  relation "fooidx" does not exist
```

Although I can kinda see confusion between what is and what is not automatically dropped during DROP TABLE without CASCADE... which there doesn't seem to be quick documentation on.

backup/queries_table_defs.go
kyeap-vmware: I tried to repro a panic using partition tables, but I can't seem to produce a panic that I thought I saw last month. I'll remove this change.

integration/predata_functions_queries_test.go
```diff
@@ -18,14 +18,17 @@ var _ = Describe("backup integration tests", func() {
...
 		BeforeEach(func() {
 			testutils.SkipIfBefore5(connectionPool)
 			if connectionPool.Version.AtLeast("7") {
 				prokindValue = "f"
 				plannerSupportValue = "-"
+				proparallelValue = "u"
 			} else {
 				prokindValue = ""
 				plannerSupportValue = ""
+				proparallelValue = ""
```
jimmyyih, line 31: Make sure to have these integration test refactors as a separate commit before merging in this PR.

jimmyyih (approved):

PR #472, opened 2021-01-07
Fix some integration tests for GPDB 7 changes

jimmyyih:
Some gpbackup commits that went in did not run against the integration
test suite which resulted in some missing test refactors to introduce
GPDB 7 specific struct changes. Also, proparallel was not being
queried correctly for GetAggregates() because it was accidentally
placed in the GPDB 6 query and not the GPDB 7 query.


roicos (approved):

integration/predata_shared_queries_test.go
```diff
@@ -341,6 +341,10 @@ PARTITION BY RANGE (date)
 	})
 	Describe("GetAccessMethods", func() {
 		It("returns information for user defined access methods", func() {
+			if connectionPool.Version.Before("7") {
```
shivzone, line 344: similar to above comment

integration/predata_shared_create_test.go
```diff
@@ -177,6 +177,10 @@ PARTITION BY RANGE (year)
 	})
 	Describe("PrintAccessMethodStatement", func() {
 		It("creates user defined access method for tables and indexes", func() {
+			if connectionPool.Version.Before("7") {
```
shivzone, line 180: replace block with testutils.SkipIfBefore7(connectionPool)

shivzone (approved):

integration/predata_shared_create_test.go
jimmyyih: Sure.

PR #473, opened 2021-01-07
Transform

roicos:
In GPDB 7+, users are able to define a transform to better convert SQL data types to their procedural language counterparts. We need to support this in gpbackup/gprestore.

### DIFF
--- a/backup/predata_functions.go
+++ b/backup/predata_functions.go
@@ -367,6 +367,40 @@ func PrintCreateLanguageStatements(metadataFile *utils.FileWithByteCount, toc *t
 	}
 }
 
+func PrintCreateTransformStatements(metadataFile *utils.FileWithByteCount, toc *toc.TOC, transforms []Transform, transformMetadata MetadataMap, funcInfoMap map[uint32]FunctionInfo) {
+	for _, transform := range transforms {
+		fromSQLFunc, fromSQLIsDefined := funcInfoMap[transform.FromSQLFunc]
+		toSQLFunc, toSQLIsDefined := funcInfoMap[transform.ToSQLFunc]
+		TypeFQN := fmt.Sprintf("%s.%s", transform.TypeNamespace, transform.TypeName)
+
+		if !fromSQLIsDefined && !toSQLIsDefined {
+			gplog.Warn(fmt.Sprintf("Skipping invalid transform object for type %s and language %s; At least one of FROM and TO functions should be specified.", TypeFQN, transform.LanguageName))
+			continue
+		}
+		start := metadataFile.ByteCount
+		statement := fmt.Sprintf("\n\nCREATE TRANSFORM FOR %s LANGUAGE %s (", TypeFQN, transform.LanguageName)
+		if fromSQLIsDefined {
+			statement += fmt.Sprintf("FROM SQL WITH FUNCTION %s", fromSQLFunc.FQN())
+		} else {
+			gplog.Warn(fmt.Sprintf("No FROM function found for transform object with type %s and language %s\n", TypeFQN, transform.LanguageName))
+		}
+
+		if toSQLIsDefined {
+			if fromSQLIsDefined {
+				statement += ", "
+			}
+			statement += fmt.Sprintf("TO SQL WITH FUNCTION %s", toSQLFunc.FQN())
+		} else {
+			gplog.Warn(fmt.Sprintf("No TO function found for transform object with type %s and language %s\n", TypeFQN, transform.LanguageName))
+		}
+		statement += ");"
+		metadataFile.MustPrintf(statement)
+		section, entry := transform.GetMetadataEntry()
+		toc.AddMetadataEntry(section, entry, start, metadataFile.ByteCount)
+		PrintObjectMetadata(metadataFile, toc, transformMetadata[transform.GetUniqueID()], transform, "")
+	}
+}
+
 func PrintCreateConversionStatements(metadataFile *utils.FileWithByteCount, toc *toc.TOC, conversions []Conversion, conversionMetadata MetadataMap) {
 	for _, conversion := range conversions {
 		start := metadataFile.ByteCount
--- a/backup/queries_functions.go
+++ b/backup/queries_functions.go
@@ -887,6 +887,54 @@ func GetProceduralLanguages(connectionPool *dbconn.DBConn) []ProceduralLanguage 
 	return results
 }
 
+type Transform struct {
+	Oid           uint32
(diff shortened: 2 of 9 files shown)


backup/predata_functions.go
```diff
@@ -365,6 +365,35 @@ func PrintCreateLanguageStatements(metadataFile *utils.FileWithByteCount, toc *t
...
+		if transform.FromSQLFunc == "-" && transform.ToSQLFunc == "-" {
+			gplog.Warn(fmt.Sprintf("Bogus transform definition, at least one of trffromsql and trftosql should be nonzero, object will be not backed up: Transform{Typename: %s, Language: %s}", transform.TypeName, transform.LanguageName))
+			continue
+		}
+		start := metadataFile.ByteCount
+		statement := fmt.Sprintf("\n\nCREATE TRANSFORM FOR %s LANGUAGE %s (", transform.TypeName, transform.LanguageName)
+		if transform.FromSQLFunc != "-" {
+			statement += fmt.Sprintf("FROM SQL WITH FUNCTION %s", transform.FromSQLFunc)
+		} else {
+			gplog.Warn("Bogus value in pg_transform.trffromsql field\n")
```
jimmyyih, line 379: Let's change the message to something like this:
"No FROM function found for transform object with type %s and language %s."

backup/predata_functions.go
```diff
@@ -365,6 +365,35 @@ func PrintCreateLanguageStatements(metadataFile *utils.FileWithByteCount, toc *t
...
+		}
+
+		if transform.ToSQLFunc != "-" {
+			if transform.FromSQLFunc != "-" {
+				statement += ", "
+			}
+			statement += fmt.Sprintf("TO SQL WITH FUNCTION %s", transform.ToSQLFunc)
+		} else {
+			gplog.Warn("Bogus value in pg_transform.trftosql field\n")
```
jimmyyih, line 388: Let's change the message to something like this:
"No TO function found for transform object with type %s and language %s."

backup/predata_functions.go
```diff
@@ -365,6 +365,35 @@ func PrintCreateLanguageStatements(metadataFile *utils.FileWithByteCount, toc *t
...
 }
 
+func PrintCreateTransformsStatements(metadataFile *utils.FileWithByteCount, toc *toc.TOC, transforms []Transform) {
+	for _, transform := range transforms {
+		if transform.FromSQLFunc == "-" && transform.ToSQLFunc == "-" {
+			gplog.Warn(fmt.Sprintf("Bogus transform definition, at least one of trffromsql and trftosql should be nonzero, object will be not backed up: Transform{Typename: %s, Language: %s}", transform.TypeName, transform.LanguageName))
+			continue
+		}
+		start := metadataFile.ByteCount
+		statement := fmt.Sprintf("\n\nCREATE TRANSFORM FOR %s LANGUAGE %s (", transform.TypeName, transform.LanguageName)
```
jimmyyih, line 375: I think the TYPE should be an FQN which you might need to query the type namespace for.

backup/wrappers.go
```diff
@@ -485,6 +485,16 @@ func backupProceduralLanguages(metadataFile *utils.FileWithByteCount,
...
 }
 
+func backupTransforms(metadataFile *utils.FileWithByteCount, funcInfoMap map[uint32]FunctionInfo) {
+	if connectionPool.Version.Before("7") {
+		return
+	}
+	gplog.Verbose("Writing CREATE TRANSFORM statements to metadata file")
+	transforms := GetTransforms(connectionPool)
+	objectCounts["Transforms"] = len(transforms)
+	PrintCreateTransformsStatements(metadataFile, globalTOC, transforms, funcInfoMap)
```
jimmyyih, line 495: Part 2 about the transform comments: Probably need to put in something like `transformMetadata := GetMetadataForObjectType(connectionPool, TYPE_TRANSFORM)` and pass it into the Print function.

backup/predata_functions.go
```diff
@@ -367,6 +367,40 @@ func PrintCreateLanguageStatements(metadataFile *utils.FileWithByteCount, toc *t
...
+			}
+			statement += fmt.Sprintf("TO SQL WITH FUNCTION %s", toSQLFunc.FQN())
+		} else {
+			gplog.Warn(fmt.Sprintf("No TO function found for transform object with type %s and language %s\n", TypeFQN, transform.LanguageName))
+		}
+		statement += ");"
+		metadataFile.MustPrintf(statement)
+		section, entry := transform.GetMetadataEntry()
+		toc.AddMetadataEntry(section, entry, start, metadataFile.ByteCount)
+		PrintObjectMetadata(metadataFile, toc, transformMetadata[transform.GetUniqueID()], transform, "")
```
jimmyyih, line 400: The TRANSFORM comments are now being outputted to the metadata file but they're not correct and make gprestore fail.  Have you tried backuping up and restoring the COMMENT command (`COMMENT ON TRANSFORM FOR pg_catalog.int8 LANGUAGE c IS 'my comment';`) mentioned in an earlier comment?  I think the `PrintObjectMetadata()` function needs to be slightly modified to support TRANSFORM comments which luckily the `transform` argument being passed in here should have everything.

At the moment, I'm getting `COMMENT ON TRANSFORM  IS 'my comment';`.  This is probably because the metadata object print function outputs relation name whereas for TRANSFORM objects, it needs the `FOR %s LANGUAGE %s` similar to the above code.

roicos: Modified transform.FQN() method for that. Transform is the only thing that doesn't have a name field of some kind but just connects type and procedural language.

backup/queries_functions.go
```diff
@@ -887,6 +887,54 @@ func GetProceduralLanguages(connectionPool *dbconn.DBConn) []ProceduralLanguage
...
+			StartByte:       0,
+			EndByte:         0,
+		}
+}
+
+func (trf Transform) GetUniqueID() UniqueID {
+	return UniqueID{ClassID: PG_TRANSFORM_OID, Oid: trf.Oid}
+}
+
+func (trf Transform) FQN() string {
```
roicos, line 915: Here is the change to dump comment on transform correctly.

backup/predata_functions.go
jimmyyih: Yeah, I thought about that solution too but was a bit hesitant to suggest it since it's not truly the FQN and other COMMENT objects might have the same problem in the future that may have a real FQN + extra syntax (e.g. comments on policy objects which we probably don't have yet).  It's fine for now since we can always easily refactor it when doing COMMENT support for other missed objects.

jimmyyih (approved):

PR #474, opened 2021-01-07
Tests for indexes/constraints with INCLUDE clause

roicos:
- In GPDB 7+, we are able to non-key columns to
  a PRIMARY KEY or UNIQUE index/constraint. This
  should supposedly work in current gpbackup code
  where UNIQUE is covered by pg_get_indexdef()
  and PRIMARY KEY is covered by pg_get_constraintdef().
  This commit adds an integration test to make sure
  we have coverage.

### DIFF
--- a/integration/postdata_queries_test.go
+++ b/integration/postdata_queries_test.go
@@ -216,6 +216,19 @@ PARTITION BY RANGE (date)
 			structmatcher.ExpectStructsToMatchExcluding(&index1, &results[1])
 		})
 
+		It("returns a slice for an index with non-key columns included", func() {
+			testutils.SkipIfBefore7(connectionPool)
+			testhelper.AssertQueryRuns(connectionPool, "CREATE TABLE public.table_with_index (a int, b int, c int, d int) DISTRIBUTED BY (a);")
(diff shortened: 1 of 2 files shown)


shivzone (approved):

integration/postdata_queries_test.go
```diff
@@ -193,6 +193,20 @@ PARTITION BY RANGE (date)
 
 			structmatcher.ExpectStructsToMatchExcluding(&index1, &results[0], "Oid")
 		})
+		It("returns a slice for an index with non-key columns included", func() {
+			testutils.SkipIfBefore7(connectionPool)
+			testhelper.AssertQueryRuns(connectionPool, "CREATE TABLE public.table_with_index (a int, b int, c int, d int) DISTRIBUTED BY (a);")
+			defer testhelper.AssertQueryRuns(connectionPool, "DROP TABLE public.table_with_index")
+			testhelper.AssertQueryRuns(connectionPool, "CREATE UNIQUE INDEX table_with_index_idx ON public.table_with_index USING btree (a, b) INCLUDE (c, d);")
+			defer testhelper.AssertQueryRuns(connectionPool, "DROP INDEX public.table_with_index_idx")
```
jimmyyih, line 201: No need to drop the index.  The DROP TABLE will drop it automatically.

PR #475, opened 2021-01-08
Support force row level security on table owner

kyeap-vmware:
In GPDB 7+, users can add row level security to their tables to enforce
row permissions for other users.  On top of that, the table owning user
can force the restriction on themselves.

### DIFF
--- a/backup/queries_table_defs.go
+++ b/backup/queries_table_defs.go
@@ -80,6 +80,7 @@ type TableDefinition struct {
 	AccessMethodName        string
 	PartitionKeyDef         string
 	AttachPartitionInfo     AttachPartitionInfo
+	ForceRowSecurity        bool
 }
 
 /*
@@ -106,6 +107,7 @@ func ConstructDefinitionsForTables(connectionPool *dbconn.DBConn, tableRelations
 	partitionAlteredSchemaMap := GetPartitionAlteredSchema(connectionPool)
 	partitionKeyDefs := GetPartitionKeyDefs(connectionPool)
 	attachPartitionInfo := GetAttachPartitionInfo(connectionPool)
+	forceRowSecurity := GetForceRowSecurity(connectionPool)
 
 	gplog.Verbose("Constructing table definition map")
 	for _, tableRel := range tableRelations {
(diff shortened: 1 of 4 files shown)


backup/queries_table_defs.go
```diff
@@ -644,6 +647,27 @@ func GetAttachPartitionInfo(connectionPool *dbconn.DBConn) map[uint32]AttachPart
...
+	FROM pg_class
+	WHERE relforcerowsecurity = 't'
+		AND oid >= %d`, FIRST_NORMAL_OBJECT_ID)
+
+	results := make([]uint32, 0)
+	err := connectionPool.Select(&results, query)
+	gplog.FatalOnError(err)
+	resultMap := make(map[uint32]bool)
+	for _, oid := range results {
+		resultMap[oid] = true
```
roicos, line 666: In this map we will not have any 'false' values, tables with 'false' value simply will not be in the map. Will it affect line 134 when we construct tableDefinition?

kyeap-vmware: It will not. Since the default is to return zero value if a key is not present. In the case of a boolean map, false will be returned if a key is not found. 

source: https://blog.golang.org/maps
> It can be convenient that a map retrieval yields a zero value when the key is not present.

backup/queries_table_defs.go
```diff
@@ -644,6 +647,27 @@ func GetAttachPartitionInfo(connectionPool *dbconn.DBConn) map[uint32]AttachPart
...
+func GetForceRowSecurity(connectionPool *dbconn.DBConn) map[uint32]bool {
+	if connectionPool.Version.Before("7") {
+		return make(map[uint32]bool, 0)
+	}
+
+	query := fmt.Sprintf(`
+	SELECT oid
+	FROM pg_class
+	WHERE relforcerowsecurity = 't'
+		AND oid >= %d`, FIRST_NORMAL_OBJECT_ID)
```
jimmyyih, line 659: I know I wrote this query in the story but looking at this kinda hurts knowing that many other table queries hit pg_class similarly and can be combined... but that's a refactor for another day.

jimmyyih (approved):

roicos (approved):

PR #477, opened 2021-01-11
`make install` should NOT call build

roicos:
- We have three different build Makefile targets
   (build, debug, and build_linux). After building,
   calling the install target will first run the
   build target... which pretty much makes the build
   targets useless if you want to run the install target.

https://dp.ci.gpdb.pivotal.io/teams/main/pipelines/dev:gpbackup_install_remove_build - prod pipeline is green.


shivzone (approved):
Since you are updating `install` can you also update the README which refers to `install_helper` instead of `install`

jimmyyih (approved):
I'm surprised nothing in the pipeline broke.  Probably because the pipeline jobs use the gppkg install instead?

roicos:
> I'm surprised nothing in the pipeline broke. Probably because the pipeline jobs use the gppkg install instead?

The only place we have `make install` in our ci is al-tests.bash
`make install helper_path=/tmp/\${OLD_BACKUP_VERSION}/gpbackup_helper`
 to run end_to_end.

jimmyyih:
I see... so the only reference left is in the backwards compatibility part.  That makes sense and should be fine for now.

PR #480, opened 2021-01-20
 Support index statistics for expression columns

shivzone:
In postgres 12 with GPDB7, statistics can be used on index columns that are defined as an expression.
This PR captures extracts the additional catalog properties namely `StatisticsColumns` and `StatisticsValues` and uses it to generate the corresponding `ALTER INDEX` DDL statements to set index statistics

Postgres Reference:
postgres/postgres@e4fca461

### DIFF
--- a/backup/postdata.go
+++ b/backup/postdata.go
@@ -7,6 +7,8 @@ package backup
  */
 
 import (
+	"strings"
+
 	"github.com/greenplum-db/gp-common-go-libs/gplog"
 	"github.com/greenplum-db/gpbackup/toc"
 	"github.com/greenplum-db/gpbackup/utils"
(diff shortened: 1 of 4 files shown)


backup/postdata.go
```diff
@@ -43,6 +45,15 @@ func PrintCreateIndexStatements(metadataFile *utils.FileWithByteCount, toc *toc.
 				metadataFile.MustPrintf("\nALTER TABLE %s REPLICA IDENTITY USING INDEX %s;", tableFQN, index.Name)
 				toc.AddMetadataEntry(section, entry, start, metadataFile.ByteCount)
 			}
+			if index.StatisticsColumns != "" && index.StatisticsValues != ""{
+				cols := strings.Split(index.StatisticsColumns, ",")
+				vals := strings.Split(index.StatisticsValues, ",")
```
jimmyyih, line 50: Should we have a sanity check to make sure len(cols) == len(vals)?  It's most likely fine without it but thoughts?

integration/postdata_queries_test.go
```diff
@@ -229,6 +229,24 @@ PARTITION BY RANGE (date)
...
 			structmatcher.ExpectStructsToMatchExcluding(&expectedIndex, &results[0], "Oid")
 		})
+		It("returns a slice of an index with statistics on expression columns", func() {
+			testutils.SkipIfBefore7(connectionPool)
+			testhelper.AssertQueryRuns(connectionPool, "CREATE TABLE public.simple_table(i int, j int, k int)")
+			defer testhelper.AssertQueryRuns(connectionPool, "DROP TABLE public.simple_table")
+			testhelper.AssertQueryRuns(connectionPool, "CREATE INDEX simple_table_idx1 ON public.simple_table(i, (i+100), (j * 8))")
+			testhelper.AssertQueryRuns(connectionPool, "ALTER INDEX public.simple_table_idx1 ALTER COLUMN 2 SET STATISTICS 400")
+			testhelper.AssertQueryRuns(connectionPool, "ALTER INDEX public.simple_table_idx1 ALTER COLUMN 3 SET STATISTICS 500")
+			defer testhelper.AssertQueryRuns(connectionPool, "DROP INDEX public.simple_table_idx1")
```
jimmyyih, line 239: This defer is unnecessary.  The `DROP TABLE public.simple_table` will automatically drop the index.

jimmyyih (approved):

backup/postdata.go
shivzone: The only way that can happen is if the catalog gets corrupted. Nevertheless safer to add the check.

PR #483, opened 2021-01-26
Update go version in CI pipeline

roicos:
We were using golang 1.13 in our CI pipeline. this commit upgrades to the latest.


jimmyyih (commented):
I see how the Golang version is being updated for our test jobs in the CI pipeline but the build job doesn't seem to have been changed.  The `go.mod` file is updated but that is just a directive that controls features.  Downloading the build container and running it on Docker, the Golang version is still 1.13.4.  Is there a corresponding container image update PR that wasn't noted in the tracker story?

PR #484, opened 2021-01-29
Integration tests fix for Photon OS

roicos:
-Photon OS has a known issue of having
  default locale available is "en_US.UTF-8".
  However some our integration tests assume
  that it's en_US.utf-8. This commet makes
  regexpr assumptions for collation and ctype
  case insensitive and introduces a function
  that is trying to find utf8 locale for
  current environment.


jimmyyih (approved):

kyeap-vmware (approved):

jimmyyih:
@roicos Is this PR still needed?

PR #489, opened 2021-02-17
Detect and panic on views with anyarray type casts in view definition

jimmyyih:
View definitions with anyarray typecasts are not restorable. If we
detect an anyarray type cast, we error out and panic to prevent a
corrupted backup that will cause gprestore to error out later. A
helpful message is given as well to make sure the user is notified
that they must manually drop the view or recreate the view without
explicit array type casts.

### DIFF
--- a/backup/queries_relations.go
+++ b/backup/queries_relations.go
@@ -15,6 +15,7 @@ import (
 	"github.com/greenplum-db/gpbackup/options"
 	"github.com/greenplum-db/gpbackup/toc"
 	"github.com/greenplum-db/gpbackup/utils"
+	"github.com/pkg/errors"
 )
 
 func relationAndSchemaFilterClause() string {
@@ -334,11 +335,20 @@ func GetAllViews(connectionPool *dbconn.DBConn) []View {
 
 	// Remove all views that have NULL definitions. This can happen
 	// if the query above is run and a concurrent view drop happens
-	// just before the pg_get_viewdef function execute.
+	// just before the pg_get_viewdef function execute. Also, error
+	// out if a view has anyarray typecasts in their view definition
+	// as those will error out during restore. Anyarray typecasts can
+	// show up if the view was created with array typecasting in an
+	// OpExpr (Operator Expression) argument.
 	verifiedResults := make([]View, 0)
 	for _, result := range results {
 		if result.Definition.Valid {
-			verifiedResults = append(verifiedResults, result)
+			if strings.Contains(result.Definition.String, "::anyarray") {
+				gplog.Fatal(errors.Errorf("Detected anyarray type cast in view definition for View '%s'", result.FQN()),
+					"Drop the view or recreate the view without explicit array type casts.")
+			} else {
+				verifiedResults = append(verifiedResults, result)
+			}
 		} else {
 			gplog.Warn("View '%s.%s' not backed up, most likely dropped after gpbackup had begun.", result.Schema, result.Name)
 		}
--- a/integration/predata_relations_queries_test.go
+++ b/integration/predata_relations_queries_test.go
@@ -414,6 +414,20 @@ PARTITION BY LIST (gender)
 			Expect(results).To(HaveLen(1))
 			structmatcher.ExpectStructsToMatchExcluding(&view, &results[0], "Oid")
 		})
+		It("PANIC on views with anyarray typecasts in its view definition", func() {
+			// The view definition gets incorrectly converted and stored as
+			// `SELECT '{1}'::anyarray = NULL::anyarray`. This issue is fixed
+			// in later versions of GPDB.
+			if (connectionPool.Version.AtLeast("5.28.6") && connectionPool.Version.Before("6")) ||
+				connectionPool.Version.AtLeast("6.14.1") {
+				Skip("test only applicable to GPDB 4.3.X, GPDB 5.0.0 - 5.28.5, and GPDB 6.0.0 - 6.14.0")
+			}
+			testhelper.AssertQueryRuns(connectionPool, "CREATE VIEW public.opexpr_array_typecasting AS SELECT '{1}'::int[] = NULL::int[]")
+			defer testhelper.AssertQueryRuns(connectionPool, "DROP VIEW public.opexpr_array_typecasting")
+
+			defer testhelper.ShouldPanicWithMessage(`[CRITICAL]:-Detected anyarray type cast in view definition for View 'public.opexpr_array_typecasting': Drop the view or recreate the view without explicit array type casts.`)
+			_ = backup.GetAllViews(connectionPool)
+		})
 		It("returns a slice for a view with options", func() {
 			testutils.SkipIfBefore6(connectionPool)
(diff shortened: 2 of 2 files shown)


backup/queries_relations.go
```diff
@@ -334,11 +334,19 @@ func GetAllViews(connectionPool *dbconn.DBConn) []View {
...
+	// all views that have anyarray typecasts in their view definition
+	// as those will error out during restore. Anyarray typecasts can
+	// show up if the view was created with array typecasting in an
+	// OpExpr (Operator Expression) argument.
 	verifiedResults := make([]View, 0)
 	for _, result := range results {
 		if result.Definition.Valid {
-			verifiedResults = append(verifiedResults, result)
+			if strings.Contains(result.Definition.String, "::anyarray") {
+				gplog.Warn("View '%s.%s' not backed up due to detected anyarray typecasts in the view definition. Recreate the view without array typecasting to fix the issue.", result.Schema, result.Name)
```
jimmyyih, line 346: @oakbarrett Is this warning message okay?  I'm a bit meh on the second sentence.

soumyadeep2007: "Recreate the view without explicit array type casts..."
might be slightly better.

integration/predata_relations_queries_test.go
```diff
@@ -414,6 +416,17 @@ PARTITION BY LIST (gender)
...
+			// The view definition gets incorrectly converted/stored as
+			// `SELECT '{1}'::anyarray = '{2}'::anyarray`. This is a known GPDB issue.
+			testhelper.AssertQueryRuns(connectionPool, "CREATE VIEW public.opexpr_array_typecasting AS SELECT '{1}'::int[] = '{2}'::int[]")
+			defer testhelper.AssertQueryRuns(connectionPool, "DROP VIEW public.opexpr_array_typecasting")
+
+			results := backup.GetAllViews(connectionPool)
+
+			Expect(results).To(HaveLen(0))
+			Expect(logFile).To(Say(regexp.QuoteMeta(`[WARNING]:-View 'public.opexpr_array_typecasting' not backed up due to detected anyarray typecasts in the view definition. Recreate the view without array typecasting to fix the issue.`)))
+		})
```
soumyadeep2007, line 429: This test is going to break for versions of 5x and 6x carrying the fix. This test should be guarded by a version check?

jimmyyih: No idea what you mean.  I don't see a GPDB 5X/6X version with this issue fixed yet. ;)

jimmyyih:
Thinking about cascaded views... we won't be able to skip to prevent a gprestore failure.  It might be better to just error out instead of warn (does the other warning of a deleted view also require upgrading the warning to an error? 🤔).  At the same time, the idea of `CREATE VIEW fooview AS SELECT 'lol::anyarray'` being falsely detected is kinda unrealistic as an edge case... but it technically is valid so another detection method may be needed instead.

integration/predata_relations_queries_test.go
soumyadeep2007: But we will have one soon :)

backup/queries_relations.go
jimmyyih: Updated error message and changed from WARN to FATAL due to possible cascaded views.

jimmyyih:
Updated patch to FATAL instead of WARN/skip since cascaded views can be a likely scenario.  As for the `CREATE VIEW fooview AS SELECT 'lol::anyarray'` edge case, I think it should be okay since it's unrealistic for that text constant to be in a view definition.

backup/queries_relations.go
```diff
@@ -334,11 +335,20 @@ func GetAllViews(connectionPool *dbconn.DBConn) []View {
...
+	// out if a view has anyarray typecasts in their view definition
+	// as those will error out during restore. Anyarray typecasts can
+	// show up if the view was created with array typecasting in an
+	// OpExpr (Operator Expression) argument.
 	verifiedResults := make([]View, 0)
 	for _, result := range results {
 		if result.Definition.Valid {
-			verifiedResults = append(verifiedResults, result)
+			if strings.Contains(result.Definition.String, "::anyarray") {
+				gplog.Fatal(errors.Errorf("Detected anyarray type cast in view definition for View '%s'", result.FQN()),
```
oakbarrett, line 347: This message works (I had a drafted comment yesterday, but never submitted it).   We will also work with Docs so the end user is aware of the issue and the steps to mitigate it.

roicos (approved):

backup/queries_relations.go
oakbarrett: @jimmyyih 

at what stage will this panic? 

i.e. here, or is the check earlier? 

20210222:14:58:04 gpbackup:gpadmin:mdw:029858-[DEBUG]:-Retrieving CAST information
**20210222:14:58:04 gpbackup:gpadmin:mdw:029858-[DEBUG]:-Retrieving views**
20210222:14:58:04 gpbackup:gpadmin:mdw:029858-[DEBUG]:-Writing CREATE SEQUENCE statements to metadata file

jimmyyih: It will PANIC and exit gracefully after the `Retrieving views` log message:
```
[DEBUG]:-Retrieving views
[CRITICAL]:-Detected anyarray type cast in view definition for View 'public.fooview': Drop the view or recreate the view without explicit array type casts.
```

integration/predata_relations_queries_test.go
jimmyyih: Added version checks to the test:
```
+                       // `SELECT '{1}'::anyarray = NULL::anyarray`. This issue is fixed
+                       // in later versions of GPDB.
+                       if (connectionPool.Version.AtLeast("5.28.6") && connectionPool.Version.Before("6")) ||
+                               connectionPool.Version.AtLeast("6.14.1") {
+                               Skip("test only applicable to GPDB 4.3.X, GPDB 5.0.0 - 5.28.5, and GPDB 6.0.0 - 6.14.0")
+                       }
```

This assumes that the patch referenced in the commit message is backported to GPDB 5.28.6 and a fix is created for GPDB 7 (our GPDB 7 pipeline job is currently disabled so no worries on time).

backup/queries_relations.go
lenapivotal: @jimmyyih is "type cast" one word or two? If two, I need to fix some of the references in the docs, for consistency.

jimmyyih: @lenapivotal Honestly... I have no idea and always used "typecast" until @soumyadeep2007  came along and used "type cast"... which had me ask the same question you're asking.  Quick Googling shows both are used for the same definition (both "typecast" and "type cast" are used in Postgres as well).  IMO, it probably doesn't matter and they seem interchangeable.

jimmyyih:
Mainly waiting for https://github.com/greenplum-db/gpdb/pull/11558 to be merged to solidify the GPDB version check in the test but more PR reviews/approvals would also help with PR merge confidence.

backup/queries_relations.go
lenapivotal: ok, thank you.

soumyadeep2007 (approved):

PR #490, opened 2021-02-18
Properly filter out implicitly created indexes

roicos:
- When we create constraints, GPDB also implicitly creates
   the index that normally would be filtered out when we dump
   indexes. However the gpbackup assumptions we make about
   the implicit index name are not correct in case of long
   table or column names. Index names have a limitation of 63
   characters. When table name + column name is longer than 63
   characters, GPDB will strip some characters off and we can't
   predict the exact way it will do it. This results in the
   implicit index not being properly filtered out and will be dumped
   twice in the metadata file during backup (CREATE UNIQUE INDEX and
   ALTER TABLE ADD CONSTRAINT UNIQUE). Then on restore we have two
   indexes - one from running sql query to create the index and
   another one - from running sql query to add constraint

### DIFF
--- a/backup/queries_postdata.go
+++ b/backup/queries_postdata.go
@@ -25,18 +25,15 @@ import (
  * to get indexes, but multiple unique indexes can be created on the
  * same column so we only want to filter out the implicit ones.
  */
-func ConstructImplicitIndexNames(connectionPool *dbconn.DBConn) string {
-	query := `
-	SELECT DISTINCT
-		n.nspname || '.' || t.relname || '_' || a.attname || '_key' AS string
-	FROM pg_constraint c
-		JOIN pg_class t ON c.conrelid = t.oid
-		JOIN pg_namespace n ON t.relnamespace = n.oid
-		JOIN pg_attribute a ON c.conrelid = a.attrelid
-		JOIN pg_index i ON i.indrelid = c.conrelid
-	WHERE a.attnum > 0
-		AND i.indisunique = 't'
-		AND i.indisprimary = 'f'`
+func ConstructImplicitIndexOidList(connectionPool *dbconn.DBConn) string {
+	query := fmt.Sprintf(`
+	SELECT i.indexrelid
+	FROM pg_index i
+		JOIN pg_depend d on i.indexrelid = d.objid
+		JOIN pg_constraint c on d.refobjid = c.oid
+	WHERE i.indexrelid >= %d
+		AND i.indisunique is true
+		AND i.indisprimary is false;`, FIRST_NORMAL_OBJECT_ID)
 	indexNames := dbconn.MustSelectStringSlice(connectionPool, query)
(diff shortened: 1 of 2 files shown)


integration/postdata_queries_test.go
```diff
@@ -14,23 +15,34 @@ import (
...
+	Describe("ConstructImplicitIndexOidList", func() {
 		It("returns an empty string if there are no implicit indexes", func() {
 			testhelper.AssertQueryRuns(connectionPool, "CREATE TABLE public.simple_table(i int)")
 			defer testhelper.AssertQueryRuns(connectionPool, "DROP TABLE public.simple_table")
 
-			indexNameSet := backup.ConstructImplicitIndexNames(connectionPool)
+			indexNameSet := backup.ConstructImplicitIndexOidList(connectionPool)
 
 			Expect(indexNameSet).To(Equal(""))
 		})
```
jimmyyih, line 26: IMO, this basic test is useless (of course there's no implicit index; it's just a simple table).  Just to add better coverage for the function we changed, can you change this test to add an explicit unique index on this table (e.g. just add a line to do `CREATE UNIQUE INDEX public.simple_table_unique_index ON public.simple_table USING btree(i)`).

backup/queries_postdata.go
```diff
@@ -25,18 +25,15 @@ import (
...
-		AND i.indisprimary = 'f'`
+func ConstructImplicitIndexOidList(connectionPool *dbconn.DBConn) string {
+	query := fmt.Sprintf(`
+	SELECT i.indexrelid
+	FROM pg_index i
+		JOIN pg_depend d on i.indexrelid = d.objid
+		JOIN pg_constraint c on d.refobjid = c.oid
+	WHERE i.indexrelid >= %d
+		AND i.indisunique is true
+		AND i.indisprimary is false;`, FIRST_NORMAL_OBJECT_ID)
```
soumyadeep2007, line 36: Behavior change: We are only looking at user tables now. That is intended. Correct?

jimmyyih: Implicitly created indexes on user tables, yes.  Either way, users aren't able to create indexes on catalog tables.  If you're referencing the 16384 OID check, that's mainly to reduce the number of rows by excluding catalog entries.  Doing that not only helps the Query Dispatcher process but also gpbackup in memory management and map lookup.

soumyadeep2007: Shouldn't we apply the same filter to `GetIndexes` to only consider user created explicit indexes?

jimmyyih: Probably, but it's out of scope for this PR and should be done as part of an overall refactoring effort (there's a lot of places that should filter only on user-created objects).  It's a somewhat known refactoring chore that's low priority.

soumyadeep2007 (approved):

oakbarrett (approved):

PR #500, opened 2021-05-17
Quote table names for exclude relation filter.

tm-drtina:
Include relation filter first quotes table names before fetching OIDs from the DB. This step was missing in exclude relation filter, which made exclusion of tables, where table name or schema name was a reserved word, impossible.

Also modified two tests I found that would suffer from this behavior.

### DIFF
--- a/backup/queries_relations.go
+++ b/backup/queries_relations.go
@@ -24,7 +24,10 @@ func relationAndSchemaFilterClause() string {
 	}
 	filterRelationClause = SchemaFilterClause("n")
 	if len(MustGetFlagStringArray(options.EXCLUDE_RELATION)) > 0 {
-		excludeOids := getOidsFromRelationList(connectionPool, MustGetFlagStringArray(options.EXCLUDE_RELATION))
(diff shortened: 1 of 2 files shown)


jimmyyih (approved):
Confirmed reported issue and tested the PR.  I think there might be more work to be done for exclusion edge cases but this PR should be good enough for now.

oakbarrett (approved):

shivzone (approved):
Nice catch !

jimmyyih:
I modified the commit message to be more descriptive (mostly just adding your PR opening comment and that this only fixes gpbackup exclusion filter):
```
Author: Tomas Drtina <tm.drtina@gmail.com>
Date:   Mon May 17 19:35:56 2021 +0200

    Quote table names for gpbackup exclude relation filter

    Include relation filter first quotes table names before fetching OIDs
    from the DB. This step was missing in exclude relation filter, which
    made exclusion of tables, where table name or schema name was a
    reserved word, impossible.

    Also modified two tests I found that would suffer from this behavior.
```

Thanks for the contribution!

Issue #494, opened 2021-05-05
gprestore not support parallel restoration of indexes


cobolbaby:
In the process of restoring data using gprestore, when the `jobs` option is specified, the tablespace setting of the index may be incorrect when the index is rebuilt.

```
$ gprestore --plugin-config /opt/greenplum/config/s3_backup.yml --timestamp 20210430155944 --create-db --jobs 4 --on-error-continue
...
20210502:09:48:33 gprestore:gpadmin:gp6mdw:029875-[INFO]:-Restoring post-data metadata
20210502:09:51:02 gprestore:gpadmin:gp6mdw:029875-[DEBUG]:-Post-data objects restored:  10% (41/404)
20210502:10:09:55 gprestore:gpadmin:gp6mdw:029875-[DEBUG]:-Post-data objects restored:  20% (81/404)
20210502:10:10:16 gprestore:gpadmin:gp6mdw:029875-[DEBUG]:-Post-data objects restored:  30% (122/404)
20210502:10:10:16 gprestore:gpadmin:gp6mdw:029875-[DEBUG]:-Post-data objects restored:  30% (123/404)
20210502:10:10:34 gprestore:gpadmin:gp6mdw:029875-[DEBUG]:-Post-data objects restored:  50% (202/404)
20210502:10:10:43 gprestore:gpadmin:gp6mdw:029875-[DEBUG]:-Post-data objects restored:  60% (243/404)
20210502:10:11:03 gprestore:gpadmin:gp6mdw:029875-[DEBUG]:-Post-data objects restored:  70% (283/404)
20210502:10:11:20 gprestore:gpadmin:gp6mdw:029875-[DEBUG]:-Error encountered when executing statement: ALTER INDEX dw.idx_dw_fact_cpu_sno_parts_sno SET TABLESPACE tbs_ssd01; Error was: ERROR: relation "dw.idx_dw_fact_cpu_sno_parts_sno" does not exist (SQLSTATE 42P01)
20210502:10:11:28 gprestore:gpadmin:gp6mdw:029875-[DEBUG]:-Error encountered when executing statement: ALTER INDEX dw.idx_dw_fact_cpu_sn_sno SET TABLESPACE tbs_ssd01; Error was: ERROR: relation "dw.idx_dw_fact_cpu_sn_sno" does not exist (SQLSTATE 42P01)
[...]

PR #501, opened 2021-05-25
Properly handle postdata metadata for parallel postdata restore

jimmyyih:
Postdata metadata (e.g. ALTER INDEX, ALTER EVENT TRIGGER, COMMENT ON)
was not being handled for parallel postdata restore. There was no
dependency information stored in the backup TOC file and gprestore had
no logic to handle postdata object dependencies. This would result in
"does not exist" errors when the postdata metadata SQL statements were
being executed before their dependent postdata objects were finished
being created.

To fix this issue, we now specifically identify postdata metadata in
the backup TOC file during gpbackup. When gprestore with --jobs flag
is specified, the postdata metadata is filtered into a third batch of
SQL statements which are executed after the postdata objects are
created in the second batch postdata execution.

Reported by user cobolbaby:
https://github.com/greenplum-db/gpbackup/issues/494

### DIFF
--- a/backup/predata_acl.go
+++ b/backup/predata_acl.go
@@ -62,6 +62,19 @@ func PrintStatements(metadataFile *utils.FileWithByteCount, toc *toc.TOC,
 		start := metadataFile.ByteCount
 		metadataFile.MustPrintf("\n\n%s\n", statement)
 		section, entry := obj.GetMetadataEntry()
+
+		/*
+		 * Postdata metadata needs to be specifically marked as such to
+		 * help with gprestore postdata parallel restore (--jobs).
+		 */
+		if section == "postdata" {
+			entry.ObjectType = fmt.Sprintf("%s METADATA", entry.ObjectType)
+			if entry.Schema != "" {
+				entry.ReferenceObject = utils.MakeFQN(entry.Schema, entry.Name)
+			} else {
+				entry.ReferenceObject = entry.Name
+			}
+		}
 		toc.AddMetadataEntry(section, entry, start, metadataFile.ByteCount)
 	}
 }
--- a/end_to_end/end_to_end_suite_test.go
+++ b/end_to_end/end_to_end_suite_test.go
@@ -1490,4 +1490,62 @@ var _ = Describe("backup and restore end to end tests", func() {
 			"--metadata-only")
 		assertRelationsCreated(restoreConn, 4)
 	})
+	It("runs gprestore with jobs flag and postdata has metadata", func() {
+		if useOldBackupVersion {
+			Skip("This test is not needed for old backup versions")
+		}
+
+		if backupConn.Version.Before("6") {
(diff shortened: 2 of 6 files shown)


roicos (approved):

backup/predata_acl.go
```diff
@@ -62,6 +62,19 @@ func PrintStatements(metadataFile *utils.FileWithByteCount, toc *toc.TOC,
...
+		 * help with gprestore postdata parallel restore (--jobs).
+		 */
+		if section == "postdata" {
+			entry.ObjectType = fmt.Sprintf("%s METADATA", entry.ObjectType)
+			if entry.Schema != "" {
+				entry.ReferenceObject = utils.MakeFQN(entry.Schema, entry.Name)
+			} else {
+				entry.ReferenceObject = entry.Name
+			}
+		}
```
kyeap-vmware, line 77: Might be worth it in the future to comment that this is specifically a backwards compatibility fix.

kyeap-vmware (approved):

end_to_end/end_to_end_suite_test.go
```diff
@@ -1490,4 +1490,62 @@ var _ = Describe("backup and restore end to end tests", func() {
...
+		// Create a trigger. Currently for triggers, the only metadata is COMMENT.
+		testhelper.AssertQueryRuns(backupConn, `CREATE TRIGGER postdata_trigger AFTER INSERT OR DELETE OR UPDATE ON postdata_metadata.foobar FOR EACH STATEMENT EXECUTE PROCEDURE pg_catalog."RI_FKey_check_ins"();`)
+		testhelper.AssertQueryRuns(backupConn, "COMMENT ON TRIGGER postdata_trigger ON postdata_metadata.foobar IS 'hello';")
+
+		// Create an event trigger. Currently for event triggers, there are 2 possible
+		// pieces of metadata: ENABLE and COMMENT.
+		if backupConn.Version.AtLeast("6") {
+			testhelper.AssertQueryRuns(backupConn, "CREATE OR REPLACE FUNCTION postdata_metadata.postdata_eventtrigger_func() RETURNS event_trigger AS $$ BEGIN END $$ LANGUAGE plpgsql;")
+			testhelper.AssertQueryRuns(backupConn, "CREATE EVENT TRIGGER postdata_eventtrigger ON sql_drop EXECUTE PROCEDURE postdata_metadata.postdata_eventtrigger_func();")
+			testhelper.AssertQueryRuns(backupConn, "ALTER EVENT TRIGGER postdata_eventtrigger DISABLE;");
```
kyeap-vmware, line 1537: Unneeded semicolon at the end of this line.

backup/predata_acl.go
jimmyyih: Should be fine.  It's only when there's a patch that is NOT backwards compatible where we should explicitly state as such.

PR #502, opened 2021-05-25
Gprestore --redirect-schema works with --metadata-only

roicos:
- Validation has been changed to make --redirect-schema
   and --metadata-only flags working together for gprestore.

### DIFF
--- a/end_to_end/end_to_end_suite_test.go
+++ b/end_to_end/end_to_end_suite_test.go
@@ -844,6 +844,28 @@ var _ = Describe("backup and restore end to end tests", func() {
 			assertDataRestored(restoreConn, schema3TupleCounts)
 			assertRelationsCreatedInSchema(restoreConn, "schema2", 0)
 		})
+		It("runs --redirect-schema with --matadata-only", func(){
+			skipIfOldBackupVersionBefore("1.17.0")
+			testhelper.AssertQueryRuns(restoreConn,
+				"DROP SCHEMA IF EXISTS schema_to_redirect CASCADE; CREATE SCHEMA \"schema_to_redirect\";")
+			defer testhelper.AssertQueryRuns(restoreConn,
+				"DROP SCHEMA schema_to_redirect CASCADE")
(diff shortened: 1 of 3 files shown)


end_to_end/end_to_end_suite_test.go
```diff
@@ -844,6 +844,27 @@ var _ = Describe("backup and restore end to end tests", func() {
...
+			testhelper.AssertQueryRuns(backupConn,
+				"CREATE TABLE foo.bar(i int)")
+			timestamp := gpbackup(gpbackupPath, backupHelperPath, "--metadata-only", "--include-schema", "foo")
+			gprestore(gprestorePath, restoreHelperPath, timestamp,
+				"--redirect-db", "restoredb",
+				"--redirect-schema", "schema3",
+				"--include-table", "foo.bar",
+				"--metadata-only")
+			assertRelationsCreatedInSchema(restoreConn, "schema3", 1)
+			assertDataRestored(restoreConn, map[string]int{"schema3.bar": 0})
```
jimmyyih, line 866: Would suggest making the schema and table name more unique than "foo", "bar", and "schema3".  Not only would that help prevent test object conflicts (moreso for future tests), it would also help debug test failures since grepping for `foo.bar` in the pg_log directory could be cumbersome vs a very unique test schema and table name.

Also, for the CREATE TABLE, some data should be added to make the data restore assertion valid.  Something like `CREATE TABLE schema.table AS SELECT generate_series(1,10);` should work fine.

jimmyyih (approved):
Please amend your commit message to give credit to the issue reporter.

Appending something like this to the commit message should be fine:
```
Reported by Github user Konstantin-La:
https://github.com/greenplum-db/gpbackup/issues/491
```

And also note in the commit message that this is a regression from v1.18:
```
This was a regression introduced in gpbackup v1.18.0 from this commit:
https://github.com/greenplum-db/gpbackup/commit/f85de46254dc21b0841191255274588ab6d49f37
```

PR #503, opened 2021-06-02
Set xmloption session GUC to content for GPDB 6+

kyeap-vmware:
This fix makes it so that XML document data can be restored when the
xmloption GUC is set to content. The xmloption GUC can be interchanged
in sessions between content and document, but cannot reliably do so
during restore. Postgres introduced the fix specifically for
dump/restore operations.

reference commits:
https://github.com/greenplum-db/gpdb/commit/78f84fe0f478142c490dcffd5eb46704b0131b9e
https://github.com/postgres/postgres/commit/8ba485422933e0ae48e8c4e1686ea59cfab75111

### DIFF
--- a/end_to_end/end_to_end_suite_test.go
+++ b/end_to_end/end_to_end_suite_test.go
@@ -844,7 +844,7 @@ var _ = Describe("backup and restore end to end tests", func() {
 			assertDataRestored(restoreConn, schema3TupleCounts)
 			assertRelationsCreatedInSchema(restoreConn, "schema2", 0)
 		})
-		It("runs --redirect-schema with --matadata-only", func(){
+		It("runs --redirect-schema with --matadata-only", func() {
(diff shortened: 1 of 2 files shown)


jimmyyih:
Try using this repro as an end_to_end test:
```
// Set up the XML table that contains XML content
CREATE TABLE xml_test AS SELECT xml 'fooxml';

// Set up database that has xmloption default to document instead of content
CREATE DATABASE document_db;
ALTER DATABASE document_db SET xmloption TO document;

// Take backup of the xml_test table
gpbackup --dbname postgres --include-table public.xml_test

// Restore it to document_db where restore will only succeed if xmloption session GUC is set to content
gprestore --timestamp 20210602123059 --redirect-db document_db
```

jimmyyih (approved):

PR #504, opened 2021-06-04
Updated s3 plugin test to use n1-standard-2 due to comply with new chunksize

shivzone:
With the updated s3 plugin chunksize for both backup and restore
(500MB) intended for stability and performance of production workloads,
the existing instance type for s3 plugin test (n1-standard-1)
isn't sufficient considering the default concurrency of 6. Bumping up
the instance type to n1-standard-2 to accomodate the new defaults.

Working pipeline: https://dp.ci.gpdb.pivotal.io/teams/main/pipelines/dev:gpbackup_shivram (Note this isn't public)


jimmyyih:
Would it be better to modify the generated s3 plugin configuration file in `ci/scripts/s3-plugin-tests.bash` to reduce the chunk sizes?  Understandably, changing from n1-standard-1 to n1-standard-2 is just an extra couple nickles an hour but that can add up for this simple regression test.

shivzone:
> Would it be better to modify the generated s3 plugin configuration file in `ci/scripts/s3-plugin-tests.bash` to reduce the chunk sizes? Understandably, changing from n1-standard-1 to n1-standard-2 is just an extra couple nickles an hour but that can add up for this simple regression test.

I thought about that but was concerned that we wouldn't be testing our default configs agianst the functional tests in `s3_plugin_test`. That said `s3_plugin_perf` is available to exercise the default configs but the test scenarios might not be the same, so i'm good with that (if you can confirm again).

jimmyyih (approved):

shivzone:
I've updated to keep the instance types the same and instead set the upload/download chunksize to 100MB for s3 plugin test

PR #505, opened 2021-06-10
Avoid lock acquisition for foreign tables

hughcapet:
`backupDataForAllTables` function used to attempt to acquire AccessShareLock on foreign tables (which is not allowed by greenplum source code). It leads to "`fdw_table_name" is not a table (SQLSTATE 42809)` error and a failure of the whole backup operation.
This behavior is fixed by skipping lock acquisition not only for the main worker thread (worker 0) but also for the tables, that don't require data backup (there is no need to acquire locks in non-main threads on both external and foreign tables).

Nevertheless, it is probably better to exclude the `if table.SkipDataBackup()` [check](https://github.com/greenplum-db/gpbackup/blob/master/backup/data.go#L88-L89) from `BackupSingleTableData ` function and move it [at the beginning](https://github.com/greenplum-db/gpbackup/blob/master/backup/data.go#L148) of a worker goroutine to skip such tables earlier. However I didn't dare to introduce such changes, as there are explicit [unit tests](https://github.com/greenplum-db/gpbackup/blob/master/backup/data_test.go#L176-L193) on execution of this function with external and foreign tables, which in this case will be broken.

### DIFF
--- a/backup/data.go
+++ b/backup/data.go
@@ -85,32 +85,27 @@ func CopyTableOut(connectionPool *dbconn.DBConn, table Table, destinationToWrite
 }
 
 func BackupSingleTableData(table Table, rowsCopiedMap map[uint32]int64, counters *BackupProgressCounters, whichConn int) error {
-	if table.SkipDataBackup() {
-		gplog.Verbose("Skipping data backup of table %s because it is either an external or foreign table.", table.FQN())
+	atomic.AddInt64(&counters.NumRegTables, 1)
+	numTables := counters.NumRegTables //We save this so it won't be modified before we log it
+	if gplog.GetVerbosity() > gplog.LOGINFO {
+		// No progress bar at this log level, so we note table count here
+		gplog.Verbose("Writing data for table %s to file (table %d of %d)", table.FQN(), numTables, counters.TotalRegTables)
 	} else {
-
-		atomic.AddInt64(&counters.NumRegTables, 1)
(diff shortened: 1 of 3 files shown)


jimmyyih:
Nice catch!  Appreciate the very detailed PR.  This is definitely an unintended regression from https://github.com/greenplum-db/gpbackup/commit/c876db964a218a6be1726be96d3260515ed8f752.

> Nevertheless, it is probably better to exclude the if table.SkipDataBackup() check from BackupSingleTableData function and move it at the beginning of a worker goroutine to skip such tables earlier. However I didn't dare to introduce such changes, as there are explicit unit tests on execution of this function with external and foreign tables, which in this case will be broken.

Definitely agree with you.  In fact, I'm not even sure why the external/foreign tables are even passed into the data dumping step (`backupData()`); they should ideally have been filtered out to avoid all these duplicated checks... we'll have to look into this for future refactoring.

For now though, it should be safe to go with the idea you outlined (move the `table.SkipDataBackup()` check to the beginning of the worker goroutine) and just delete the two unit tests you outlined.  Would you be able to update the PR to do that?  If not, we could do it for you.

hughcapet:
@jimmyyih, thank you for the prompt reply! I will update the PR shortly

jimmyyih (approved):

kyeap-vmware (approved):

jimmyyih:
Thanks @hughcapet for the contribution!  My apologies for the slow merging.

PR #507, opened 2021-06-11
Make gprestore to inform helpers to skip the table

roicos:
- When we restore with --single-data-file we have a named
   pipe between data file and COPY. If COPY fails before
   pipe was opened for writing - the attempt to open the pipe
   will be blocked waiting forever for someone to read on the
   other end. If --on-error-continue is specified - another COPY
   command will be launched, but it would not get a pipe to read from
   bacause helpers are blocked on the previous one.

   To prevent this we need some form of communication between gprestore
   and gpbackup_helpers. The solution implemented in this commit
   is when gprestore needs to skip the table because it errored out and
   has --on-error-continue - it creates skip file with oid of the object
   to skip and copies it to all segments. Then on helper side on segment
   we have goroutine that checks for skip file on interval and if this
   happens - we setup an error and go to the next loop if we have
   --on-error-continue instead of waiting forever on the old pipe.

### DIFF
--- a/helper/restore_helper.go
+++ b/helper/restore_helper.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"os/exec"
 	"strings"
+	"time"
 
 	"github.com/greenplum-db/gpbackup/toc"
 	"github.com/greenplum-db/gpbackup/utils"
@@ -36,6 +37,12 @@ type RestoreReader struct {
 	bufReader  *bufio.Reader
 	seekReader io.ReadSeeker
 	readerType ReaderType
+}
+
+type RestoreWriterInfo struct {
+	writer	*bufio.Writer
+	file	*os.File
+	err	error
 }
 
 func (r *RestoreReader) positionReader(pos uint64) error {
@@ -97,6 +104,9 @@ func doRestoreAgent() error {
 	log(fmt.Sprintf("Using reader type: %s", reader.readerType))
 
 	for i, oid := range oidList {
+		skipCh := make(chan bool, 1)
+		writerInfoCh := make(chan RestoreWriterInfo, 1)
+
 		if wasTerminated {
 			return errors.New("Terminated due to user request")
 		}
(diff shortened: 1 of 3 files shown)


jimmyyih:
The concurrency from the two new go routines might complicate the already-delicate restore_helper code.  Could you try to do the skip file check in the same pipe opening section by using `syscall.O_NONBLOCK` and have retry logic for error `EAGAIN` (maybe use sleep 100ms since there will probably be a sleep involved in the loop)?  Let's try that out and then decide between this implementation or the nonblocking one.

helper/restore_helper.go
```diff
@@ -114,40 +124,73 @@ func doRestoreAgent() error {
...
+					time.Sleep(100 * time.Millisecond)
+				}
+			}
+		}()
+
+		go func() {
+			info := RestoreWriterInfo{}
+			info.writer, info.file, info.err = getRestorePipeWriter(currentPipe)
+			writerInfoCh <- info
+		}()
```
kyeap-vmware, line 149: This might be for another story.
This go routine blocking forever was discussed before, but if the file is skipped would it be possible to open the pipe to unblock this go routine and abort writing anything to it as a cleanup step to allow this go routine to exit?

utils/agent_remote.go
```diff
@@ -219,3 +220,31 @@ func CheckAgentErrorsOnSegments(c *cluster.Cluster, fpInfo filepath.FilePathInfo
...
+	localSkipFile, err := operating.System.TempFile("", "gprestore_skip")
+	gplog.FatalOnError(err, "Cannot open temporary file to write skip oid")
+	defer func() {
+		err = operating.System.Remove(localSkipFile.Name())
+		if err != nil {
+			gplog.Warn("Cannot remove temporary oid file: %s, Err: %s", localSkipFile.Name(), err.Error())
+		}
+	}()
+
+	WriteOidsToFile(localSkipFile.Name(), []string{oid})
```
kyeap-vmware, line 234: I don't see the oid written to the skip file being used anywhere. Is this needed?

utils/agent_remote.go
```diff
@@ -219,3 +220,31 @@ func CheckAgentErrorsOnSegments(c *cluster.Cluster, fpInfo filepath.FilePathInfo
 	}
 	return nil
 }
+
+func CreateSkipFileOnSegments(oid string, c *cluster.Cluster, fpInfo filepath.FilePathInfo) {
+	localSkipFile, err := operating.System.TempFile("", "gprestore_skip")
```
kyeap-vmware, line 225: I think the localSkipFile can be deleted from master after it is scp'ed to all segments.

utils/agent_remote.go
jimmyyih: We're saving a step by just touching the file where the oid is already in the filename.  I think that should be fine.

PR #508, opened 2021-06-17
Adds test replicate_backup fails if already exists (ddboost)

roicos:
- If on ddboost we are trying to replicate backup on that was already
   replicated (even partially) we should fail with
   "DDBoost open file in exclusive mode failed on remote".
   This commit simply adds the test for this case.

### DIFF
--- a/plugins/plugin_test.sh
+++ b/plugins/plugin_test.sh
@@ -360,6 +360,17 @@ if [[ "$plugin" == *gpbackup_ddboost_plugin ]] && [[ "$plugin_config" == *ddboos
     echo "[RUNNING] replicate_backup"
     $plugin replicate_backup $plugin_config $time_second_for_repl
 
+    set +e
+    echo "[RUNNING] replicate_backup second time to verify error"
+    output=`$plugin replicate_backup $plugin_config $time_second_for_repl 2>&1`
+    echo "replicate backup second time, output: $output"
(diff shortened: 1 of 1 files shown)


plugins/plugin_test.sh
```diff
@@ -373,6 +373,39 @@ if [[ "$plugin" == *gpbackup_ddboost_plugin ]] && [[ "$plugin_config" == *ddboos
...
+
+    echo "[RUNNING] replicate_backup"
+    $plugin replicate_backup $plugin_config $time_second_for_repl
+
+    echo "[RUNNING] replicate_backup second time"
+    output=`$plugin replicate_backup $plugin_config $time_second_for_repl 2>&1`
+    if [[ ! "$output" == *"DDBoost open file in exclusive mode failed on remote"* ]]; then
+        echo "Error doesn't match when trying to replicate already replicated backup"
+        exit 1
+    fi
```
jimmyyih, line 404: I would suggest moving this into the replication test above instead around line 362 so that we verify that `restore_data` still works.  Can just add it as `[RUNNING] replicate_backup second time to verify error`.

jimmyyih (approved):

kyeap-vmware (approved):

shivzone (approved):
@roicos looks good to merge

PR #510, opened 2021-07-01
COPY communicates with helper through non-blocking pipe

roicos:
- When we restore with --single-data-file, we write data from the data file to a pipe, and then
   the COPY command reads from this pipe to load data to the database.
   The way we open pipe is blocking - that means opening it for writing would be blocked until
   the reader opens it. If COPY (reader) fails for whatever reason - gpbackup_helper (writer)
   will be blocked opening the pipe.
   If --on-error-continue is specified - another COPY command will be launched,
   but it would not get a pipe to read from because helpers are blocked on the previous one.

 - To prevent this, when an error occurs and we have --on-error-continue, gprestore will
   create skip_oid file on every segment to inform gpbackup-helpers to also skip the table.
   On segments, gpbackup_helper (writer) will now keep trying to open the pipe in non-blocking mode (syscall.O_NONBLOCK)
   and check if gets back the specific syscall.ENXIO error (meaning that the COPY (reader)
   are not ready to consume from it yet). When getting this error gpbackup_helper will try
   to discover a skip file for the current table to check if we are in the
   "error on restore + --on-error-continue" scenario and skip the pipe for the specific table if needed.

### DIFF
--- a/end_to_end/end_to_end_suite_test.go
+++ b/end_to_end/end_to_end_suite_test.go
@@ -3,6 +3,7 @@ package end_to_end_test
 import (
 	"flag"
 	"fmt"
+	"github.com/pkg/errors"
 	"io/ioutil"
 	"math/rand"
 	"os"
@@ -668,6 +669,61 @@ var _ = Describe("backup and restore end to end tests", func() {
 				"public.corrupt_table": 0,
 				"public.good_table1":   10,
 				"public.good_table2":   10})
+		})
+		It(`Creates skip file on segments for corrupted table for helpers to discover the file and skip it`, func(){
+			if useOldBackupVersion {
+				Skip("This test is not needed for old backup versions")
+			}
+			command := exec.Command("tar", "-xzf", "resources/corrupt-db.tar.gz", "-C", backupDir)
+			mustRunCommand(command)
+
+			testhelper.AssertQueryRuns(restoreConn,
+				"CREATE TABLE public.corrupt_table (i integer);")
+			defer testhelper.AssertQueryRuns(restoreConn,
+				"DROP TABLE public.corrupt_table")
+
+			// we know that broken value goes to seg2, so seg1 should be
+			// ok. Connect in utility mode to seg1.
+			segmentOne := backupCluster.ByContent[1]
+			port := segmentOne[0].Port
+			segConn := testutils.SetupTestDBConnSegment("restoredb", port)
+			defer segConn.Close()
+
+			// Take ACCESS EXCLUSIVE LOCK on public.corrupt_table which will
+			// make COPY on seg1 block until the lock is released. By that
+			// time, COPY on seg2 will fail and gprestore will create a skip
+			// file for public.corrupt_table. When the lock is released on seg1,
+			// the restore helper should discover the file and skip the table.
+			segConn.Begin(0)
+			segConn.Exec("LOCK TABLE public.corrupt_table IN ACCESS EXCLUSIVE MODE;")
+
+			gprestoreCmd := exec.Command(gprestorePath,
+				"--timestamp", "20190809230424",
+				"--redirect-db", "restoredb",
+				"--backup-dir", path.Join(backupDir, "corrupt-db"),
+				"--data-only", "--on-error-continue",
+				"--include-table", "public.corrupt_table")
+			_, err := gprestoreCmd.CombinedOutput()
+			Expect(err).To(HaveOccurred())
+
+			segConn.Commit(0)
+			homeDir := os.Getenv("HOME")
+			helperLogs, _ := path.Glob(path.Join(homeDir, "gpAdminLogs/gpbackup_helper*"))
+			cmdStr := fmt.Sprintf("tail -n 40 %s | grep \"Skip file has been discovered for entry\" || true", helperLogs[len(helperLogs)-1])
+
+			attemts := 1000
+			err = errors.New("Timeout to discover skip file")
+			for attemts > 0 {
+				output := mustRunCommand(exec.Command("bash", "-c", cmdStr))
+				if strings.TrimSpace(string(output)) == "" {
+					time.Sleep(5 * time.Millisecond)
+					attemts--
+				} else {
+					err = nil
+					break
+				}
+			}
+			Expect(err).NotTo(HaveOccurred())
 		})
 		It(`ensure gprestore on corrupt backup with --on-error-continue logs error tables`, func() {
 			command := exec.Command("tar", "-xzf",
--- a/helper/restore_helper.go
+++ b/helper/restore_helper.go
@@ -9,6 +9,8 @@ import (
 	"os"
 	"os/exec"
 	"strings"
+	"syscall"
+	"time"
 
 	"github.com/greenplum-db/gpbackup/toc"
 	"github.com/greenplum-db/gpbackup/utils"
@@ -95,7 +97,6 @@ func doRestoreAgent() error {
 		return err
 	}
 	log(fmt.Sprintf("Using reader type: %s", reader.readerType))
-
 	for i, oid := range oidList {
 		if wasTerminated {
 			return errors.New("Terminated due to user request")
@@ -118,13 +119,31 @@ func doRestoreAgent() error {
 		end = tocEntries[uint(oid)].EndByte
 
 		log(fmt.Sprintf("Opening pipe for oid %d: %s", oid, currentPipe))
-		writer, writeHandle, err = getRestorePipeWriter(currentPipe)
-		if err != nil {
-			// In the case this error is hit it means we have lost the
-			// ability to open pipes normally, so hard quit even if
-			// --on-error-continue is given
-			_ = utils.RemoveFileIfExists(currentPipe)
-			return err
+		for {
+			writer, writeHandle, err = getRestorePipeWriter(currentPipe)
+			if err != nil {
+				if errors.Is(err, syscall.ENXIO) {
+					// the reader is not ready, so writer will fail with very specific error
+					// discover if gprestore created a skip file for this entry
+					skipFileName := fmt.Sprintf("%s_skip_%d", *pipeFile, oid)
+					if *onErrorContinue && utils.FileExists(skipFileName) {
+						log(fmt.Sprintf("Skip file has been discovered for entry %d, skipping it", oid))
+						err = nil
+						goto LoopEnd
+					} else {
+						// keep trying to open the pipe
+						time.Sleep(100 * time.Millisecond)
+					}
+				} else {
+					// In the case this error is hit it means we have lost the
+					// ability to open pipes normally, so hard quit even if
+					// --on-error-continue is given
+					_ = utils.RemoveFileIfExists(currentPipe)
+					return err
+				}
+			} else {
+				break
+			}
 		}
 
 		log(fmt.Sprintf("Data Reader - Start Byte: %d; End Byte: %d; Last Byte: %d", start, end, lastByte))
@@ -229,8 +248,7 @@ func getRestoreDataReader(toc *toc.SegmentTOC, oidList []int) (*RestoreReader, e
 }
 
 func getRestorePipeWriter(currentPipe string) (*bufio.Writer, *os.File, error) {
-	// Opening this pipe will block until a reader connects to the pipe
-	fileHandle, err := os.OpenFile(currentPipe, os.O_WRONLY, os.ModeNamedPipe)
+	fileHandle, err := os.OpenFile(currentPipe, os.O_WRONLY|syscall.O_NONBLOCK, os.ModeNamedPipe)
 	if err != nil {
 		return nil, nil, err
 	}
--- a/testutils/functions.go
+++ b/testutils/functions.go
@@ -2,6 +2,10 @@ package testutils
 
 import (
 	"fmt"
+	"github.com/greenplum-db/gp-common-go-libs/gplog"
+	"github.com/greenplum-db/gp-common-go-libs/operating"
+	"github.com/jmoiron/sqlx"
+	"github.com/pkg/errors"
 	"io"
 	"os"
 	"os/exec"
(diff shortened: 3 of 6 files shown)


utils/agent_remote.go
```diff
@@ -219,3 +220,31 @@ func CheckAgentErrorsOnSegments(c *cluster.Cluster, fpInfo filepath.FilePathInfo
...
+	localSkipFile, err := operating.System.TempFile("", "gprestore_skip")
+	gplog.FatalOnError(err, "Cannot open temporary file to write skip oid")
+	defer func() {
+		err = operating.System.Remove(localSkipFile.Name())
+		if err != nil {
+			gplog.Warn("Cannot remove temporary oid file: %s, Err: %s", localSkipFile.Name(), err.Error())
+		}
+	}()
+
+	WriteOidsToFile(localSkipFile.Name(), []string{oid})
```
jimmyyih, line 234: Ah okay, I understand what @kyeap-vmware was trying to convey in https://github.com/greenplum-db/gpbackup/pull/507#discussion_r664971901 now.  Yeah, as we had discussed previously, we do not need to write the oid in the file since the oid is in the filename.

utils/agent_remote.go
```diff
@@ -219,3 +220,31 @@ func CheckAgentErrorsOnSegments(c *cluster.Cluster, fpInfo filepath.FilePathInfo
...
+
+		return fmt.Sprintf(`scp %s %s:%s`, sourceFile, hostname, dest)
+	}
+	remoteOutput := c.GenerateAndExecuteCommand("Scp skip file to segments", cluster.ON_LOCAL|cluster.ON_SEGMENTS, generateScpCmd)
+
+	errMsg := "Failed to scp skip file"
+	errFunc := func(contentID int) string {
+		return "Failed to run scp"
+	}
+	c.CheckClusterError(remoteOutput, errMsg, errFunc, false)
```
jimmyyih, line 249: Since we don't really need the oid written to the file, we can probably make this function much simpler by using something like this:
```
remoteOutput := globalCluster.GenerateAndExecuteCommand("Touch skip file", cluster.ON_SEGMENTS, func(contentID int) string {
                        return fmt.Sprintf("touch %s_skip_%s", fpInfo.GetSegmentPipeFilePath(contentID), oid)
                })
globalCluster.CheckClusterError(remoteOutput, "Error while touching skip file", func(contentID int) string {
                        return fmt.Sprintf("Could not create skip file %s_skip_%s", fpInfo.GetSegmentPipeFilePath(contentID), oid)
                })
```

helper/restore_helper.go
```diff
@@ -118,13 +120,31 @@ func doRestoreAgent() error {
...
+					// discover if gprestore created a skip file for this entry
+					skipFileName := fmt.Sprintf("%s_skip_%d", *pipeFile, oid)
+					log(fmt.Sprintf("looking for skip file %s", skipFileName))
+					if utils.FileExists(skipFileName) {
+						log(fmt.Sprintf("Skip file discovered for entry %d, skipping it", oid))
+						goto LoopEnd
+					} else {
+						// keep trying to open the pipe
+						time.Sleep(100 * time.Millisecond)
+					}
```
jimmyyih, line 137: Hmm, wondering if we should have some sort of timeout here... but that could be tricky since each restore_helper/COPY interaction are asynchronous which could give false positive timeouts.  Just an idea to think about I guess.

helper/restore_helper.go
```diff
@@ -118,13 +120,31 @@ func doRestoreAgent() error {
...
-		if err != nil {
-			// In the case this error is hit it means we have lost the
-			// ability to open pipes normally, so hard quit even if
-			// --on-error-continue is given
-			_ = utils.RemoveFileIfExists(currentPipe)
-			return err
+		for {
+			writer, writeHandle, err = getRestorePipeWriter(currentPipe)
+			if err != nil {
+				if errors.Is(err, syscall.ENXIO) {
```
jimmyyih, line 126: Ah okay so it's ENXIO for writers and EAGAIN for readers.  Got it.

jimmyyih (commented):
I definitely think we should go with this implementation instead of https://github.com/greenplum-db/gpbackup/pull/507.  This PR flows better and we don't have to worry about complicated Go-routine interactions.

kyeap-vmware (approved):
LGTM. Is there any additional testing that should be included with the changes?

helper/restore_helper.go
```diff
@@ -118,13 +119,31 @@ func doRestoreAgent() error {
...
+					// the reader is not ready, so writer will fail with very specific error
+					// discover if gprestore created a skip file for this entry
+					skipFileName := fmt.Sprintf("%s_skip_%d", *pipeFile, oid)
+					log(fmt.Sprintf("looking for skip file %s", skipFileName))
```
jimmyyih, line 129: I think we can get rid of this log message (it was spamming my log file).  We can safely assume that if we see `Opening pipe for oid...` that this logic of pipe open retry and skip file search is happening.

utils/agent_remote.go
```diff
@@ -219,3 +220,12 @@ func CheckAgentErrorsOnSegments(c *cluster.Cluster, fpInfo filepath.FilePathInfo
...
 	return nil
 }
+
+func CreateSkipFileOnSegments(oid string, c *cluster.Cluster, fpInfo filepath.FilePathInfo) {
+	remoteOutput := c.GenerateAndExecuteCommand("Touch skip file", cluster.ON_SEGMENTS, func(contentID int) string {
+		return fmt.Sprintf("touch %s_skip_%s", fpInfo.GetSegmentPipeFilePath(contentID), oid)
+	})
+	c.CheckClusterError(remoteOutput, "Error while touching skip file", func(contentID int) string {
+		return fmt.Sprintf("Could not create skip file %s_skip_%s", fpInfo.GetSegmentPipeFilePath(contentID), oid)
+	})
```
jimmyyih, line 230: Did this suggestion work as expected in your manual testing?

jimmyyih (commented):
Kevin brings up a good point in that we could have a more precise end_to_end test to make sure this is well covered.  Let's give it an attempt.  Let me know if you need help creating a test.

helper/restore_helper.go
```diff
@@ -118,13 +119,31 @@ func doRestoreAgent() error {
...
+					if utils.FileExists(skipFileName) {
+						log(fmt.Sprintf("Skip file discovered for entry %d, skipping it", oid))
+						goto LoopEnd
```
jimmyyih, line 132: As you mentioned offline, "device not configured" seems to be the error message for syscall.ENXIO.  In that case, let's set `err` to `nil` before jumping to `LoopEnd` to suppress the message since we've properly handled it with the skip file logic.

testutils/functions.go
```diff
@@ -58,6 +62,58 @@ func SetupTestDbConn(dbname string) *dbconn.DBConn {
...
+		Driver:   dbconn.GPDBDriver{},
+		User:     username,
+		DBName:   dbname,
+		Host:     host,
+		Port:     port,
+		Tx:       nil,
+		Version:  dbconn.GPDBVersion{},
+	}
+
+	connStr := fmt.Sprintf("postgres://%s@%s:%d/%s?sslmode=disable&statement_cache_capacity=0&gp_session_role=utility", conn.User, conn.Host, conn.Port, conn.DBName)
```
jimmyyih, line 98: Need to remember here that this will not work for GPDB 7X+.  We can take care of that in the future.

end_to_end/end_to_end_suite_test.go
```diff
@@ -669,6 +669,49 @@ var _ = Describe("backup and restore end to end tests", func() {
...
+			Expect(err).To(HaveOccurred())
+
+			segConn.Commit(0)
+
+			homeDir := os.Getenv("HOME")
+			helperLogs, _ := path.Glob(path.Join(homeDir, "gpAdminLogs/gpbackup_helper*"))
+			helperLogBytes, err := ioutil.ReadFile(helperLogs[0])
+			Expect(err).NotTo(HaveOccurred())
+			helperLogContent := string(helperLogBytes)
+			Expect(strings.Contains(helperLogContent, "Skip file discovered for entry 16937, skipping it"))
```
jimmyyih, line 713: Your `Expect` here will always return true.  You need to change this to something like:
```
-                       helperLogContent := string(helperLogBytes)
-                       Expect(strings.Contains(helperLogContent, "Skip file discovered for entry 16937, skipping it"))
+                       Expect(string(helperLogBytes)).To(ContainSubstring("Skip file discovered for entry 16937, skipping it"))
```

helper/restore_helper.go
```diff
@@ -126,9 +126,9 @@ func doRestoreAgent() error {
 					// the reader is not ready, so writer will fail with very specific error
 					// discover if gprestore created a skip file for this entry
 					skipFileName := fmt.Sprintf("%s_skip_%d", *pipeFile, oid)
-					log(fmt.Sprintf("looking for skip file %s", skipFileName))
 					if utils.FileExists(skipFileName) {
-						log(fmt.Sprintf("Skip file discovered for entry %d, skipping it", oid))
+						log(fmt.Sprintf("Skip file is discovered for entry %d, skipping it", oid))
```
jimmyyih, line 130: Let's keep the message from before "Skip file discovered" or you could change it to "Skip file has been discovered" if you want.  Whichever, make sure to sync it with the test since it's currently not.

end_to_end/end_to_end_suite_test.go
```diff
@@ -669,6 +669,49 @@ var _ = Describe("backup and restore end to end tests", func() {
...
+			port := segmentOne[0].Port
+			segConn := testutils.SetupTestDBConnSegment("restoredb", port)
+			defer segConn.Close()
+
+			// take access exclusive lock on corrupt_table
+			// this wil make COPY on seg1 to wait until lock is released
+			// by that time COPY on seg0 will fail and gprstore
+			// should create skip file for public.corrupt_table
+			// when lock is released for seg1 - gpbackup_helper should
+			// discover this file and skip the table.
```
jimmyyih, line 693: Can you format this to help with readability.  Something like this should be fine:
```
                        // Take ACCESS EXCLUSIVE LOCK on public.corrupt_table which will
                        // make COPY on seg1 block until the lock is released. By that
                        // time, COPY on seg0 will fail and gprestore will create a skip
                        // file for public.corrupt_table. When the lock is released on seg1,
                        // the restore helper should discover the file and skip the table.
```

end_to_end/end_to_end_suite_test.go
```diff
@@ -669,6 +669,49 @@ var _ = Describe("backup and restore end to end tests", func() {
...
+			//skipIfOldBackupVersionBefore("1.21.0")
+			command := exec.Command("tar", "-xzf", "resources/corrupt-db.tar.gz", "-C", backupDir)
+			mustRunCommand(command)
+
+			testhelper.AssertQueryRuns(restoreConn,
+				"CREATE TABLE public.corrupt_table (i integer);")
+			defer testhelper.AssertQueryRuns(restoreConn,
+				"DROP TABLE public.corrupt_table")
+
+			// we know that broken value goes to segment0, so segment1 should be ok
```
jimmyyih, line 682: Let's mention utility mode here to help future readers:
```
                        // We know that the broken value will go into seg0 so seg1 should be
                        // okay. Connect in utility mode to seg1.
```

end_to_end/end_to_end_suite_test.go
```diff
@@ -669,6 +669,49 @@ var _ = Describe("backup and restore end to end tests", func() {
...
+			defer segConn.Close()
+
+			// take access exclusive lock on corrupt_table
+			// this wil make COPY on seg1 to wait until lock is released
+			// by that time COPY on seg0 will fail and gprstore
+			// should create skip file for public.corrupt_table
+			// when lock is released for seg1 - gpbackup_helper should
+			// discover this file and skip the table.
+			segConn.Begin(0)
+			segConn.Exec("LOCK TABLE public.corrupt_table IN access exclusive MODE;")
```
jimmyyih, line 695: Let's upper it to `ACCESS EXCLUSIVE MODE`.

end_to_end/end_to_end_suite_test.go
jimmyyih, line 713: This doesn't really work that well for future test runs.  Maybe we should tail the last helper log file instead similar to this:
```
tail -n 40 /Users/jyih/gpAdminLogs/gpbackup_helper_20210719.log
```

Maybe we could take inspiration from the `integration/helper_test.go` function:
```
func printHelperLogOnError(helperErr error) {
        if helperErr != nil {
                homeDir := os.Getenv("HOME")
                helperFiles, _ := filepath.Glob(filepath.Join(homeDir, "gpAdminLogs/gpbackup_helper_*"))
                command := exec.Command("tail", "-n 20", helperFiles[len(helperFiles)-1])
                output, _ := command.CombinedOutput()
                fmt.Println(string(output))
        }
}
```

helper/restore_helper.go
```diff
@@ -118,13 +119,31 @@ func doRestoreAgent() error {
...
-			_ = utils.RemoveFileIfExists(currentPipe)
-			return err
+		for {
+			writer, writeHandle, err = getRestorePipeWriter(currentPipe)
+			if err != nil {
+				if errors.Is(err, syscall.ENXIO) {
+					// the reader is not ready, so writer will fail with very specific error
+					// discover if gprestore created a skip file for this entry
+					skipFileName := fmt.Sprintf("%s_skip_%d", *pipeFile, oid)
+					if utils.FileExists(skipFileName) {
```
jimmyyih, line 129: Hmm, I'm trying to think if we need to utilize the `*onErrorContinue` variable here to prevent checking for skip file if the user did not set `--on-error-continue`.  It shouldn't matter but what are your thoughts on that?  If anything, we could change this conditional to something like this:
```
if *onErrorContinue && utils.FileExists(fmt.Sprintf("%s_skip_%d", *pipeFile, oid)) {
```

end_to_end/end_to_end_suite_test.go
```diff
@@ -669,6 +669,55 @@ var _ = Describe("backup and restore end to end tests", func() {
...
+			// before it got there.
+			time.Sleep(100*time.Millisecond)
+			segConn.Commit(0)
+
+			homeDir := os.Getenv("HOME")
+			helperLogs, _ := path.Glob(path.Join(homeDir, "gpAdminLogs/gpbackup_helper*"))
+			command = exec.Command("tail", "-n 40", helperLogs[len(helperLogs)-1])
+			helperLogBytes, err := command.CombinedOutput()
+			Expect(err).NotTo(HaveOccurred())
+			Expect(string(helperLogBytes)).To(ContainSubstring("Skip file has been discovered for entry 16937, skipping it"))
```
jimmyyih, line 719: The sleep here is not very ideal.  We should probably instead loop on a grep check for the expected substring.  You can have a fail timeout value of maybe a minute or two and loop every 5 milliseconds.

You can even simplify the `exec.Command` to use something like `tail -n 40 <helper_log> | grep "Skip file has been discovered"` and check for success instead of using the Expect/ContainSubstring combo.

end_to_end/end_to_end_suite_test.go
```diff
@@ -669,6 +670,59 @@ var _ = Describe("backup and restore end to end tests", func() {
 				"public.good_table1":   10,
 				"public.good_table2":   10})
 		})
+		It(`Creates skip file on segments for corrupted table for helpers to discover the file and skip it`, func(){
+			//skipIfOldBackupVersionBefore("1.21.0")
```
jimmyyih, line 674: This is meant for the backwards compatibility test (e.g. back up with old gpbackup version and restore with bleeding-edge latest gprestore).  For this test, we're using a static corrupt backup so we can skip this test for the backwards compatibility testing.

Let's change this to the following:
```
if useOldBackupVersion {
        Skip("This test is not needed for old backup versions")
}
```

jimmyyih (approved):
I think this should be good to go after fixing the backwards compatibility test part to skip.

PR #519, opened 2021-07-20
Add flag validation unit tests

jimmyyih:
There has always been issues with flag combinations due to the
somewhat complex validation checks we do. Sometimes, this leads to
unfortunate regressions when a new feature gets added that includes a
new flag. To prevent regressions from happening, this patch adds some
unit tests to solidify what flag combinations should be valid/invalid
in an easy-to-comprehend format.

Most of the unit test scenarios are obtained from gpbackup/gprestore documentation on what is and is not valid together.  There are some TODO items left over where a flag combination seems to defy documentation or my assumption on its validity.

### DIFF
--- a/backup/validate_test.go
+++ b/backup/validate_test.go
@@ -1,12 +1,16 @@
 package backup_test
 
 import (
+	"strings"
+
 	"github.com/DATA-DOG/go-sqlmock"
+	"github.com/spf13/cobra"
 	"github.com/greenplum-db/gp-common-go-libs/testhelper"
 	"github.com/greenplum-db/gpbackup/backup"
 	"github.com/greenplum-db/gpbackup/options"
 
 	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/ginkgo/extensions/table"
 )
 
 var _ = Describe("backup/validate tests", func() {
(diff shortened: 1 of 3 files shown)


roicos (approved):

backup/validate_test.go
```diff
@@ -177,4 +181,94 @@ var _ = Describe("backup/validate tests", func() {
...
+			 */
+			// --exclude-schema combinations with other filters
+			Entry("--exclude-schema combos", "--exclude-schema schema1 --include-table schema.table2", false),
+			Entry("--exclude-schema combos", "--exclude-schema schema1 --include-table-file /tmp/file2", false),
+			Entry("--exclude-schema combos", "--exclude-schema schema1 --include-schema schema2", false),
+			Entry("--exclude-schema combos", "--exclude-schema schema1 --include-schema-file /tmp/file2", false),
+			Entry("--exclude-schema combos", "--exclude-schema schema1 --exclude-table schema.table2", false),
+			Entry("--exclude-schema combos", "--exclude-schema schema1 --exclude-table-file /tmp/file2", false),
+			Entry("--exclude-schema combos", "--exclude-schema schema1 --exclude-schema schema2", true),
+			Entry("--exclude-schema combos", "--exclude-schema schema1 --exclude-schema-file /tmp/file2", false),
```
kyeap-vmware, line 217: Seeing something like thing makes me wonder why all `--include*` and `--exclude*` flags are not allowed to be used all at once. Any ambiguousness can be resolved by `--include*` superseding any include/exclude conflicts.

For example 
`--exclude-schema` and `--include-table` could be useful if you wanted a full backup excluding an entire schema from a backup except for 1 table in the schema.
`--exclude-schema foo --include-table foo.bar26`

I think our current include exclude implementation doesn't allow for that though.

kyeap-vmware (approved):

backup/validate_test.go
jimmyyih: Probably due to different priorities.  It was probably easier to make them all exclusive instead of having biased logic to handle the vast amount of flag combinations.  Can you imagine a gpbackup command with a lot of filter flags?

E.g. `gpbackup --dbname postgres --include-table foo.table --exclude-schema foo --include-table-file /tmp/myfile --exclude-table bar.table --include-schema foo --include-schema foo2 --exclude-schema-file /tmp/myfile2 --include-table foo.table2`

Writing logic to handle the above example and any filter conflicts would be tedious whereas there's many other features that are more worth doing.

>For example
--exclude-schema and --include-table could be useful if you wanted a full backup excluding an entire schema from a backup except for 1 table in the schema.
--exclude-schema foo --include-table foo.bar26

At the moment, it's perfectly fine to just have the two split into separate backups (e.g. `gpbackup --exclude-schema foo` and `gpbackup --include-table foo.bar26`).  And then there's restore filtering which also gives a lot of options and versatility as well.  Overall, the filtering implementation covers most use cases... but yes I do agree it could be better.  At least we support `gprestore --include-schema --exclude-table` as a start.  Maybe we can add more later.

PR #531, opened 2021-10-29
Fix TerminateHangingCopySessions to work properly with GPDB6

hughcapet:
TerminateHangingCopySessions function used to query old names of columns
from pg_stat_activity table (current_query and procpid were changed to query and pid)
This also used to cause periodic fail of SIGINT end2end tests.


kyeap-vmware:
Thanks for catching this issue! The fix looks great. We will write some tests and merge this in.

PR #534, opened 2021-11-10
Queue parallelized COPY commands for gpbackup_helper to consume

kyeap-vmware:
To reduce the amount of time gpbackup_helper is spent blocked waiting
for COPY statements to return, gpbackup and gprestore can now queue COPY
commands to speed up --single-data-file backups and restores if the
database has cpu cycles to spare.  If a worker can't get a lock on a
table, the worker is terminated and the table is then processed by
Worker 0. Worker 0 has AccessShare locks for all tables, and acts as a
special goroutine to handle deferred tables.

Added flag --single-data-file-copy-prefetch (Default 1). In order for
this to be used during restore the backup must have been taken with
--single-data-file.

This can improve backup and restore times significantly for databases
that contain a lot of tables with very little data where the majority of
time is spent waiting for gpdb to initiate COPY.

### DIFF
--- a/backup/data.go
+++ b/backup/data.go
@@ -5,10 +5,12 @@ package backup
  */
 
 import (
+	"errors"
 	"fmt"
 	"strings"
 	"sync"
 	"sync/atomic"
+	"time"
 
 	"github.com/greenplum-db/gp-common-go-libs/dbconn"
 	"github.com/greenplum-db/gp-common-go-libs/gplog"
@@ -110,6 +112,159 @@ func BackupSingleTableData(table Table, rowsCopiedMap map[uint32]int64, counters
 	return nil
 }
 
+// backupDataForAllTablesCopyQueue does not backup tables in parallel. This is
+// specifically for single-data-file. While tables backed up one at a time, it
+// is possible to save a significant amount of time by queuing up the next
+// table to copy. Worker 0 does not have tables pre-assigned to it. The copy
+// queue uses worker 0 a special deferred worker in the event that the other
+// workers encounter locking issues. Worker 0 already has all locks on the
+// tables so it will not run into locking issues.
+func backupDataForAllTablesCopyQueue(tables []Table) []map[uint32]int64 {
+	var numExtOrForeignTables int64
+	for _, table := range tables {
+		if table.SkipDataBackup() {
+			numExtOrForeignTables++
+		}
+	}
+	counters := BackupProgressCounters{NumRegTables: 0, TotalRegTables: int64(len(tables)) - numExtOrForeignTables}
+	counters.ProgressBar = utils.NewProgressBar(int(counters.TotalRegTables), "Tables backed up: ", utils.PB_INFO)
+	counters.ProgressBar.Start()
+	rowsCopiedMaps := make([]map[uint32]int64, connectionPool.NumConns)
+	/*
+	 * We break when an interrupt is received and rely on
+	 * TerminateHangingCopySessions to kill any COPY statements
+	 * in progress if they don't finish on their own.
+	 */
+	tasks := make(chan Table, len(tables))
+	var oidMap sync.Map
+	var workerPool sync.WaitGroup
+	var copyErr error
+	// Record and track tables in a hashmap of oids and table states (preloaded with value Unknown).
+	// The tables are loaded into the tasks channel for the subsequent goroutines to work on.
+	for _, table := range tables {
+		oidMap.Store(table.Oid, Unknown)
+		tasks <- table
+	}
+	// We incremented numConns by 1 to treat connNum 0 as a special worker
+	rowsCopiedMaps[0] = make(map[uint32]int64)
+	for connNum := 1; connNum < connectionPool.NumConns; connNum++ {
+		rowsCopiedMaps[connNum] = make(map[uint32]int64)
+		workerPool.Add(1)
+		go func(whichConn int) {
+			defer workerPool.Done()
+			for table := range tasks {
+				if wasTerminated || copyErr != nil {
+					counters.ProgressBar.(*pb.ProgressBar).NotPrint = true
+					return
+				}
+
+				if table.SkipDataBackup() {
+					gplog.Verbose("Skipping data backup of table %s because it is either an external or foreign table.", table.FQN())
+					oidMap.Store(table.Oid, Complete)
+					continue
+				}
+				// If a random external SQL command had queued an AccessExclusiveLock acquisition request
+				// against this next table, the --job worker thread would deadlock on the COPY attempt.
+				// To prevent gpbackup from hanging, we attempt to acquire an AccessShareLock on the
+				// relation with the NOWAIT option before we run COPY. If the LOCK TABLE NOWAIT call
+				// fails, we catch the error and defer the table to the main worker thread, worker 0.
+				// Afterwards, we break early and terminate the worker since its transaction is now in an
+				// aborted state. We do not need to do this with the main worker thread because it has
+				// already acquired AccessShareLocks on all tables before the metadata dumping part.
+				err := LockTableNoWait(table, whichConn)
+				if err != nil {
+					if pgErr, ok := err.(*pgconn.PgError); ok && pgErr.Code != PG_LOCK_NOT_AVAILABLE {
+						copyErr = err
+						continue
+					}
+
+					if gplog.GetVerbosity() < gplog.LOGVERBOSE {
+						// Add a newline to interrupt the progress bar so that
+						// the following WARN message is nicely outputted.
+						fmt.Printf("\n")
+					}
+					gplog.Warn("Worker %d could not acquire AccessShareLock for table %s. Terminating worker and deferring table to main worker thread.",
+						whichConn, table.FQN())
+
+					oidMap.Store(table.Oid, Deferred)
+					// Rollback transaction since it's in an aborted state
+					connectionPool.MustRollback(whichConn)
+
+					// Worker no longer has a valid distributed transaction snapshot
+					break
+				}
+
+				err = BackupSingleTableData(table, rowsCopiedMaps[whichConn], &counters, whichConn)
+				if err != nil {
+					copyErr = err
+				}
+				oidMap.Store(table.Oid, Complete)
+			}
+		}(connNum)
+	}
+
+	// Special goroutine to handle deferred tables
+	// Handle all tables deferred by the deadlock detection. This can only be
+	// done with the main worker thread, worker 0, because it has
+	// AccessShareLocks on all the tables already.
+	deferredWorkerDone := make(chan bool)
+	go func() {
+		for _, table := range tables {
+			for {
+				state, _ := oidMap.Load(table.Oid)
+				if state.(int) == Unknown {
+					time.Sleep(time.Millisecond * 50)
+				} else if state.(int) == Deferred {
+					err := BackupSingleTableData(table, rowsCopiedMaps[0], &counters, 0)
+					if err != nil {
+						copyErr = err
+					}
+					oidMap.Store(table.Oid, Complete)
+					break
+				} else if state.(int) == Complete {
+					break
+				} else {
+					gplog.Fatal(errors.New("Encountered unknown table state"), "")
+				}
+			}
+		}
+		deferredWorkerDone <- true
+	}()
+
+	close(tasks)
+	workerPool.Wait()
+
+	// Check if all the workers were terminated. If they did, defer all remaining tables to worker 0
+	allWorkersTerminatedLogged := false
+	for _, table := range tables {
+		state, _ := oidMap.Load(table.Oid)
+		if state == Unknown {
+			if !allWorkersTerminatedLogged {
+				gplog.Warn("All copy queue workers terminated due to lock issues. Falling back to single main worker.")
+				allWorkersTerminatedLogged = true
+			}
+			oidMap.Store(table.Oid, Deferred)
+		}
+	}
+	// Main goroutine waits for deferred worker 0 by waiting on this channel
+	<-deferredWorkerDone
+
+	agentErr := utils.CheckAgentErrorsOnSegments(globalCluster, globalFPInfo)
+
+	if copyErr != nil && agentErr != nil {
+		gplog.Error(agentErr.Error())
+		gplog.Fatal(copyErr, "")
+	} else if copyErr != nil {
+		gplog.Fatal(copyErr, "")
+	} else if agentErr != nil {
+		gplog.Fatal(agentErr, "")
+	}
+
+	counters.ProgressBar.Finish()
+	printDataBackupWarnings(numExtOrForeignTables)
+	return rowsCopiedMaps
+}
+
 func backupDataForAllTables(tables []Table) []map[uint32]int64 {
 	var numExtOrForeignTables int64
 	for _, table := range tables {
@@ -157,8 +312,7 @@ func backupDataForAllTables(tables []Table) []map[uint32]int64 {
 				if whichConn != 0 {
 					err := LockTableNoWait(table, whichConn)
 					if err != nil {
-						// Postgres Error Code 55P03 translates to LOCK_NOT_AVAILABLE
-						if pgErr, ok := err.(*pgconn.PgError); ok && pgErr.Code != "55P03" {
+						if pgErr, ok := err.(*pgconn.PgError); ok && pgErr.Code != PG_LOCK_NOT_AVAILABLE {
 							copyErr = err
 							continue
 						}
--- a/backup/data.go
+++ b/backup/data.go
@@ -85,13 +85,14 @@ func CopyTableOut(connectionPool *dbconn.DBConn, table Table, destinationToWrite
 }
 
 func BackupSingleTableData(table Table, rowsCopiedMap map[uint32]int64, counters *BackupProgressCounters, whichConn int) error {
-	atomic.AddInt64(&counters.NumRegTables, 1)
-	numTables := counters.NumRegTables //We save this so it won't be modified before we log it
+	logMessage := fmt.Sprintf("Worker %d: Writing data for table %s to file", whichConn, table.FQN())
+	// Avoid race condition by incrementing counters in call to sprintf
+	tableCount := fmt.Sprintf(" (table %d of %d)", atomic.AddInt64(&counters.NumRegTables, 1), counters.TotalRegTables)
 	if gplog.GetVerbosity() > gplog.LOGINFO {
 		// No progress bar at this log level, so we note table count here
-		gplog.Verbose("Writing data for table %s to file (table %d of %d)", table.FQN(), numTables, counters.TotalRegTables)
+		gplog.Verbose(logMessage + tableCount)
 	} else {
-		gplog.Verbose("Writing data for table %s to file", table.FQN())
+		gplog.Verbose(logMessage)
 	}
 
 	destinationToWrite := ""
--- a/backup/global_variables.go
+++ b/backup/global_variables.go
@@ -20,6 +20,16 @@ import (
  */
 
 /*
+ Table backup state constants
+*/
+const (
+	Unknown int = iota
+	Deferred
+	Complete
+	PG_LOCK_NOT_AVAILABLE = "55P03"
+)
+
+/*
  * Non-flag variables
  */
 var (
@@ -98,6 +108,10 @@ func SetQuotedRoleNames(quotedRoles map[string]string) {
 }
 
 // Util functions to enable ease of access to global flag values
+
+func FlagChanged(flagName string) bool {
+	return cmdFlags.Changed(flagName)
+}
 
 func MustGetFlagString(flagName string) string {
 	return options.MustGetFlagString(cmdFlags, flagName)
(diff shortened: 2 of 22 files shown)


backup/global_variables.go
```diff
@@ -97,8 +97,23 @@ func SetQuotedRoleNames(quotedRoles map[string]string) {
 	quotedRoleNames = quotedRoles
 }
 
+func MinInt(a, b int) int {
```
shivzone, line 100: `gobal_variables.go` is only intended to handle global variable related getters or setters not util like functions, also for something like this its an overhead to have a function

backup/wrappers.go
```diff
@@ -39,7 +39,19 @@ func SetLoggerVerbosity() {
 
 func initializeConnectionPool(timestamp string) {
 	connectionPool = dbconn.NewDBConnFromEnvironment(MustGetFlagString(options.DBNAME))
-	connectionPool.MustConnect(MustGetFlagInt(options.JOBS))
+	var numConns int
```
shivzone, line 42: For future reference add a comment block explaining the difference between the two cases

shivzone:
My revew in still WIP ... but part of me feels if `--single-data-file-prefetch` flag is set without explicitly using `--single-data-file` can we automatically infer the user wants to backup with single data file ?

backup/global_variables.go
bmdoil: Since math.Min only accepts floats, it's actually faster to do 
```
func(a, b int) int {
	if a < b {
		return a
	} else {
		return b
	}
}(a, b)
```
vs. 
```
int(math.Min(float64(a), float64(b)))
```
BenchmarkMathMin-32             113506471               10.52 ns/op
BenchmarkIntMinAnon-32          133879941                8.647 ns/op

So, I've removed MinInt and instead create an anonymous function call in its place.
```
func CreateInitialSegmentPipes(oidList []string, c *cluster.Cluster, fpInfo filepath.FilePathInfo) int {
	// Create min(connections, tables) segment pipes on each host
	maxPipes := func(numConns, numTables int) int {
		if numConns < numTables {
			return numConns
		} else {
			return numTables
		}
	}(GetConnNums(), len(oidList))
	for i := 0; i < maxPipes; i++ {
		utils.CreateSegmentPipeOnAllHosts(oidList[i], c, fpInfo)
	}
	return maxPipes
}
```

options/flag.go
```diff
@@ -14,40 +14,41 @@ import (
 )
 
 const (
-	BACKUP_DIR            = "backup-dir"
```
roicos, line 17: Is it technically possible to make this diff look like one-line addition + whitespace additions?

shivzone:
Also, if possible add to the PR itself general comments about the core difference in the way we deferred tables are handled with `safeOIDMap`

backup/data.go
```diff
@@ -109,6 +110,191 @@ func BackupSingleTableData(table Table, rowsCopiedMap map[uint32]int64, counters
...
+	// Handle all tables deferred by the deadlock detection. This can only be
+	// done with the main worker thread, worker 0, because it has
+	// AccessShareLocks on all the tables already.
+	deferredWorkerDone := make(chan bool)
+	go func() {
+		for _, table := range tables {
+			for {
+				switch oidMap.SafeGet(table.Oid) {
+				case Unknown:
+					time.Sleep(time.Millisecond * 50)
```
shivzone, line 247: why do we sleep here ?

backup/data.go
```diff
@@ -109,6 +110,191 @@ func BackupSingleTableData(table Table, rowsCopiedMap map[uint32]int64, counters
...
+	return s.oidMap[key]
+}
+
+func (s *safeOIDMap) List() map[uint32]int {
+	s.mutex.RLock()
+	defer s.mutex.RUnlock()
+	return s.oidMap
+}
+
+func backupDataForAllTablesPrefetch(tables []Table) []map[uint32]int64 {
```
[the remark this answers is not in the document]
shivzone: Given that this is a fairly large change for the sake of time lets keep these functions separate for now and can refactor them outside of the scope of this PR

backup/validate.go
```diff
@@ -102,6 +102,9 @@ func validateFlagCombinations(flags *pflag.FlagSet) {
 	options.CheckExclusiveFlags(flags, options.NO_COMPRESSION, options.COMPRESSION_TYPE)
 	options.CheckExclusiveFlags(flags, options.NO_COMPRESSION, options.COMPRESSION_LEVEL)
 	options.CheckExclusiveFlags(flags, options.PLUGIN_CONFIG, options.BACKUP_DIR)
+	if flags.Changed(options.SINGLE_DATA_FILE_COPY_PREFETCH) && !MustGetFlagBool(options.SINGLE_DATA_FILE) {
+		gplog.Fatal(errors.Errorf("--single-data-file-copy-prefetch must be specified with --single-data-file"), "")
```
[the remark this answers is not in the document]
jimmyyih: Having it explicitly stated in the flag name will prevent any confusion so fine with me.

backup/data.go
```diff
@@ -109,6 +110,191 @@ func BackupSingleTableData(table Table, rowsCopiedMap map[uint32]int64, counters
...
+				}
+				// If a random external SQL command had queued an AccessExclusiveLock acquisition request
+				// against this next table, the --job worker thread would deadlock on the COPY attempt.
+				// To prevent gpbackup from hanging, we attempt to acquire an AccessShareLock on the
+				// relation with the NOWAIT option before we run COPY. If the LOCK TABLE NOWAIT call
+				// fails, we catch the error and defer the table to the main worker thread, worker 0.
+				// Afterwards, we break early and terminate the worker since its transaction is now in an
+				// aborted state. We do not need to do this with the main worker thread because it has
+				// already acquired AccessShareLocks on all tables before the metadata dumping part.
+
```
jimmyyih, line 201: Should remove this random newline that's separating the comment block with the below reference of `LockTableNoWait()`.

backup/data.go
```diff
@@ -109,6 +110,191 @@ func BackupSingleTableData(table Table, rowsCopiedMap map[uint32]int64, counters
...
+					err := BackupSingleTableData(table, rowsCopiedMaps[0], &counters, 0)
+					if err != nil {
+						copyErr = err
+					}
+					oidMap.SafeSet(table.Oid, Complete)
+				}
+
+				if oidMap.SafeGet(table.Oid) == Complete {
+					break
+				}
```
jimmyyih, line 258: This seems weird to have a `oidMap.SafeGet()` case be outside the switch but I understand why it was done.  Have you tried maybe using a break label?  Something like this maybe:
```
outerLoop:
        for {
                switch oidMap.SafeGet(table.Oid) {
                case Unknown:
                        time.Sleep(time.Millisecond * 50)
                case Complete:
                        break outerLoop
                case Deferred:
                        err := BackupSingleTableData(table, rowsCopiedMaps[0], &counters, 0)
                        if err != nil {
                                copyErr = err
                        }
                        oidMap.SafeSet(table.Oid, Complete)
                }
        }
```

backup/data.go
```diff
@@ -109,6 +110,191 @@ func BackupSingleTableData(table Table, rowsCopiedMap map[uint32]int64, counters
...
+	for oid := range oidMap.List() {
+		if oidMap.SafeGet(oid) == Unknown {
+			if !allWorkersTerminatedLogged {
+				gplog.Warn("All prefetch workers terminated due to lock issues. Falling back to single main worker.")
+				allWorkersTerminatedLogged = true
+			}
+			oidMap.SafeSet(oid, Deferred) // All deferred tables are processed by worker 0
+		}
+	}
+	<-deferredWorkerDone
```
jimmyyih, line 277: Very nice edge-case handling here.  Could you please add a brief comment above the logic explaining what this is doing (e.g. `Check if all the workers were terminated. If they did, defer all remaining tables to worker 0.`)

end_to_end/end_to_end_suite_test.go
```diff
@@ -1475,6 +1536,137 @@ var _ = Describe("backup and restore end to end tests", func() {
...
+		}()
+
+		// gpbackup has finished
+		output, _ := cmd.CombinedOutput()
+		stdout := string(output)
+
+		// Sanity check that 9 deadlock traps were placed during the test
+		Expect(accessExclBlockedLockCount).To(Equal(9))
+
+		// No non-main worker should have been able to run COPY due to deadlock detection
```
jimmyyih, line 1650: Since no workers would have remained, we should probably test for that new warning you guys added:
`gplog.Warn("All prefetch workers terminated due to lock issues. Falling back to single main worker.")`

backup/data.go
jimmyyih: Probably to prevent the Goroutine from constantly spinning.  The loop will be constantly hitting this `Unknown` case the majority of the time so the sleep is good... and 50ms seems about the right amount.

backup/validate.go
```diff
@@ -121,6 +124,13 @@ func validateFlagValues() {
 		gplog.Fatal(errors.Errorf("Timestamp %s is invalid.  Timestamps must be in the format YYYYMMDDHHMMSS.",
 			MustGetFlagString(options.FROM_TIMESTAMP)), "")
 	}
+	if MustGetFlagInt(options.SINGLE_DATA_FILE_COPY_PREFETCH) < 1 {
+		gplog.Fatal(errors.Errorf("--single-data-file-copy-prefetch %d is invalid. Must be at least 1",
+			MustGetFlagInt(options.SINGLE_DATA_FILE_COPY_PREFETCH)), "")
+	}
+	if MustGetFlagInt(options.SINGLE_DATA_FILE_COPY_PREFETCH) > 4 {
+		gplog.Warn("Recommend a max value of 4 for --single-data-file-copy-prefetch")
+	}
```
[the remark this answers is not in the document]
shivzone: I agree, this should just be a documenation exercise and not us warning the user. That way there is an argument to also include a note with the `--jobs` flag

shivzone (commented):
My general concern in the way we use the `oidMap` is with the possibility of heavy lock contention on it based on # of tables, value of `single-data-file-copy-prefetch`.
Even though we have upto n writers (n = `single-data-file-copy-prefetch`) and 1 reader, that can occur simultaneously at any given point in time, each writer is guaranteed to update non overlapping `oid` based on how they are distributed to the workers using the channel.

I have 2 suggestions/questions here:
1. Can we simply use sync.Map. As per its documentation `optimized for two common use cases: (1) when the entry for a given key is only ever written once but read many times, as in caches that only grow, or (2) when multiple goroutines read, write, and overwrite entries for disjoint sets of keys. In these two cases, use of a Map may significantly reduce lock contention compared to a Go map paired with a separate Mutex or RWMutex.`
Both of this applies to our use case. Might definitely be worth doing a quick benchmark this vs what we've written.

2. Since we anyway iterate over tables in a fixed order, can we simply use a Fixed size array and have each entry simply correspond to a particular oid/table. We don't need to worry about concurrent writes as workers will never be operating on the same `oid`. We can resolve having 1 writer and reader concurrency simply with atomic load and store operations

backup/data.go
jimmyyih: The other workaround is to just use the typical `if/else if/else` block since this break issue is specific to switches.  There's nothing wrong with doing that.  The fact that an additional `if` call after the switch was required kinda signifies that a switch was not the correct choice to begin with (at least without the break label).  Between a break label or changing it to an `if/else if/else` block, I would favor the `if/else if/else` block for its simplicity.

backup/data.go
jimmyyih, line 277: Is this where the main thread waits for the deferred worker 0 to complete by waiting on this channel?  If so, could you add a comment above it to make it clear (especially to non-Golang people) that it's waiting here.

backup/data.go
```diff
@@ -86,11 +86,9 @@ func CopyTableOut(connectionPool *dbconn.DBConn, table Table, destinationToWrite
 }
 
 func BackupSingleTableData(table Table, rowsCopiedMap map[uint32]int64, counters *BackupProgressCounters, whichConn int) error {
-	atomic.AddInt64(&counters.NumRegTables, 1)
-	numTables := counters.NumRegTables //We save this so it won't be modified before we log it
 	if gplog.GetVerbosity() > gplog.LOGINFO {
 		// No progress bar at this log level, so we note table count here
-		gplog.Verbose("Writing data for table %s to file (table %d of %d)", table.FQN(), numTables, counters.TotalRegTables)
+		gplog.Verbose("Writing data for table %s to file (table %d of %d)", table.FQN(), atomic.AddInt64(&counters.NumRegTables, 1), counters.TotalRegTables)
```
bmdoil, line 91: This resolves a reported race condition

backup/validate.go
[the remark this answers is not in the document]
bmdoil: Removed.

helper/restore_helper.go
```diff
@@ -48,15 +48,13 @@ func (r *RestoreReader) positionReader(pos uint64) error {
...
 			// Always hard quit if data reader has issues
-			_ = utils.RemoveFileIfExists(currentPipe)
 			return err
 		}
 		log(fmt.Sprintf("Data Reader seeked forward to %d byte offset", seekPosition))
 	case NONSEEKABLE:
 		numDiscarded, err := r.bufReader.Discard(int(pos))
 		if err != nil {
 			// Always hard quit if data reader has issues
-			_ = utils.RemoveFileIfExists(currentPipe)
```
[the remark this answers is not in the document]
bmdoil: Yes, it's now handled by DoCleanup() 
```
	for pipeName, _ := range createdPipesMap {
		log("Removing pipe %s", pipeName)
		err = deletePipe(pipeName)
		if err != nil {
			log("Encountered error removing pipe %s: %v", pipeName, err)
		}
	}
```

```
func deletePipe(pipe string) error {
	err := utils.RemoveFileIfExists(pipe)
	if err != nil {
		return err
	}

	delete(createdPipesMap, pipe)
	return nil
}
```

backup/data.go
bmdoil: With the sync.Map implementation returning the interface{}, I've made some additional changes. We try to load the oid from the map until it succeeds, then check the state. 
```
	deferredWorkerDone := make(chan bool)
	go func() {
		for _, table := range tables {
			var state interface{}
			var ok bool = false
			for !ok {
				state, ok = oidMap.Load(table.Oid)
			}
			if state.(int) == Unknown {
				time.Sleep(time.Millisecond * 50)
			} else if state.(int) == Deferred {
				err := BackupSingleTableData(table, rowsCopiedMaps[0], &counters, 0)
				if err != nil {
					copyErr = err
				}
				oidMap.Store(table.Oid, Complete)
			}
		}
		deferredWorkerDone <- true
	}()
```

bmdoil: With the sync.Map implementation returning the interface{}, I've made some additional changes. We try to load the oid from the map until it succeeds, then check the state. 
```
	deferredWorkerDone := make(chan bool)
	go func() {
		for _, table := range tables {
			var state interface{}
			var ok bool = false
			for !ok {
				state, ok = oidMap.Load(table.Oid)
			}
			if state.(int) == Unknown {
				time.Sleep(time.Millisecond * 50)
			} else if state.(int) == Deferred {
				err := BackupSingleTableData(table, rowsCopiedMaps[0], &counters, 0)
				if err != nil {
					copyErr = err
				}
				oidMap.Store(table.Oid, Complete)
			}
		}
		deferredWorkerDone <- true
	}()
```

backup/data.go
bmdoil: Had to make a change here given the sync.Map Range method, it's functionally the same.
```
	allWorkersTerminatedLogged := false
	oidMap.Range(func(oid, state interface{}) bool {
		if state.(int) == Unknown {
			if !allWorkersTerminatedLogged {
				gplog.Warn("All prefetch workers terminated due to lock issues. Falling back to single main worker.")
				allWorkersTerminatedLogged = true
			}
			state = Deferred
		}
		return true // return true to continue iterating over oidMap
		},
	)
```

backup/data.go
```diff
@@ -85,13 +86,14 @@ func CopyTableOut(connectionPool *dbconn.DBConn, table Table, destinationToWrite
...
+	logMessage := fmt.Sprintf("Writing data for table %s to file", table.FQN())
+	// Avoid race condition by incrementing counters in call to sprintf
+	tableCount := fmt.Sprintf(" (table %d of %d)", atomic.AddInt64(&counters.NumRegTables, 1), counters.TotalRegTables)
 	if gplog.GetVerbosity() > gplog.LOGINFO {
 		// No progress bar at this log level, so we note table count here
-		gplog.Verbose("Writing data for table %s to file (table %d of %d)", table.FQN(), numTables, counters.TotalRegTables)
+		gplog.Verbose(logMessage + tableCount)
 	} else {
-		gplog.Verbose("Writing data for table %s to file", table.FQN())
+		gplog.Verbose(logMessage)
```
bmdoil, line 96: This change resolved a reported race condition with numTables

backup/data.go
```diff
@@ -109,6 +111,142 @@ func BackupSingleTableData(table Table, rowsCopiedMap map[uint32]int64, counters
...
+	counters.ProgressBar.Start()
+	rowsCopiedMaps := make([]map[uint32]int64, connectionPool.NumConns)
+	/*
+	 * We break when an interrupt is received and rely on
+	 * TerminateHangingCopySessions to kill any COPY statements
+	 * in progress if they don't finish on their own.
+	 */
+	tasks := make(chan Table, len(tables))
+	// Record hashmap of oids to track which have been processed
+	var oidMap sync.Map
```
shivzone, line 134: Please make sure oidMap is initialized with `Unknown` value for all entries upfront

shivzone (commented):
The revised version using `sync.Map` and modifications to deferred queue handling looks a lot cleaner. 
Would be great if we can include a perf or scale test to evaluate stability of deferred queue handling by having a concurrent session holding an AccessExclusiveLock.

backup/data.go
bmdoil: Refactored deferred worker implementation
```
deferredWorkerDone := make(chan bool)
go func() {
	for _, table := range tables {
		for {
			state, _ := oidMap.Load(table.Oid)
			if state.(int) == Unknown {
				time.Sleep(time.Millisecond * 50)
			} else if state.(int) == Deferred {
				err := BackupSingleTableData(table, rowsCopiedMaps[0], &counters, 0)
				if err != nil {
					copyErr = err
				}
				oidMap.Store(table.Oid, Complete)
				break
			} else {  // state is Complete
				break
			}
		}
	}
	deferredWorkerDone <- true
}()
```

backup/data.go
```diff
@@ -110,6 +112,156 @@ func BackupSingleTableData(table Table, rowsCopiedMap map[uint32]int64, counters
...
+	 * TerminateHangingCopySessions to kill any COPY statements
+	 * in progress if they don't finish on their own.
+	 */
+	tasks := make(chan Table, len(tables))
+	var oidMap sync.Map
+	var workerPool sync.WaitGroup
+	var copyErr error
+	// Record hashmap of oids to track which have been processed
+	// Preload all table states to Unknown
+	// Loading the tables as tasks. Start goroutines that work on the tasks
```
jimmyyih, line 137: Nit: This comment block doesn't read very well.  Could you make it more cohesive.

Example:
> Record and track tables in a hashmap of oids and table states (preloaded with value Unknown). The tables are loaded into the tasks channel for the subsequent goroutines to work on.

backup/data.go
```diff
@@ -110,6 +112,156 @@ func BackupSingleTableData(table Table, rowsCopiedMap map[uint32]int64, counters
...
+			oidMap.Store(table.Oid, Deferred)
+		}
+	}
+	// Main goroutine waits for deferred worker 0 by waiting on this channel
+	<-deferredWorkerDone
+
+	var agentErr error
+	if MustGetFlagBool(options.SINGLE_DATA_FILE) {
+		agentErr = utils.CheckAgentErrorsOnSegments(globalCluster, globalFPInfo)
+	}
```
jimmyyih, line 249: Presumably, this function can only be called when --single-data-file is given so the `if` conditional is not really needed?

shivzone (approved):
Nice work

end_to_end/end_to_end_suite_test.go
jimmyyih: Unresolving this comment as no change was reflected in recent PR updates nor was there a rejection to the idea of adding the warning check.

bmdoil: Thanks, I had added this warning check then reverted it and forgot to unresolve. The test finishes quickly enough that the deferred goroutine backs up all the deferred tables and exits before the main goroutine even hits the for loop. I'll restructure the test to ensure the terminated worker check is hit.

backup/data.go
bmdoil: Correct, removed the if.

jmcatamney (approved):
LGTM, but it's been long enough since I worked on gpbackup that getting another approval before merging just in case would be ideal.

backup/data.go
```diff
@@ -109,6 +112,152 @@ func BackupSingleTableData(table Table, rowsCopiedMap map[uint32]int64, counters
 	return nil
 }
 
+func backupDataForAllTablesCopyQueue(tables []Table) []map[uint32]int64 {
```
shivzone, line 115: Add a comment note above thisfunction to describe in a high level the difference here. Call out the fact that we do not pre-assign work to worker 0 and purely use worker 0 only to handle deferred work in a separate go routine

PR #532, opened 2021-10-29
Handle empty tables set for an incremental data backup properly

hughcapet:
There is a potential unhandled 'out of range' error in backup process:

1. Create new db
2. Create AO table
3. Create external table
4. Run gpbackup for the new db with the following params: --leaf-partition-data --single-data-file
5. Right after that (with no data change) run gpbackup for the same db with the following params: --leaf-partition-data --single-data-file --incremental

This results in:

```
runtime error: index out of range [0] with length 0: goroutine 1 [running]:
...
/opt/go/src/runtime/panic.go:969 +0x1b9
github.com/greenplum-db/gpbackup/backup.backupData(0xc000d1e000, 0x18, 0x26)
/home/jenkins/ci/bigtop/build/gpbackup/rpm/BUILD/gpbackup-1.20.4_arenadata1/src/github.com/greenplum-db/backup/backup.go:270 +0x93c
github.com/greenplum-db/gpbackup/backup.DoBackup()
```

PR #536, opened 2021-11-23
Fix single-data-file incremental backup panic if all tables are filtered

kyeap-vmware:
When taking a single-data-file incremental backup of a database of AO
and external tables and no heap tables, gpbackup will panic if there are
no tables to backup. This is because AO tables with no incremental
changes are filtered out first while external tables are filtered out
just before creating the first table's pipe for single-data-file. If all
tables get filtered out, when gpbackup tries access the first table's
oid to create the first pipe, it will cause an 'out of range' panic
because the table list is empty. With this commit, gpbackup avoids this
situation by filtering out all external tables early before backupData
is called. This will cause backupData to return early because it knows
there is no data to backup.

### DIFF
--- a/backup/backup.go
+++ b/backup/backup.go
@@ -119,13 +119,18 @@ func DoBackup() {
 
 	gplog.Info("Gathering table state information")
 	metadataTables, dataTables := RetrieveAndProcessTables()
+	dataTables, numExtOrForeignTables := GetBackupDataSet(dataTables)
+	if len(dataTables) == 0 {
+		gplog.Warn("No tables in backup set contain data. Performing metadata-only backup instead.")
+		backupReport.MetadataOnly = true
+	}
 	// This must be a full backup with --leaf-parition-data to query for incremental metadata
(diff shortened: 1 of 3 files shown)


backup/backup.go
```diff
@@ -119,13 +119,18 @@ func DoBackup() {
 
 	gplog.Info("Gathering table state information")
 	metadataTables, dataTables := RetrieveAndProcessTables()
+	dataTables, numExtOrForeignTables := GetBackupDataSet(dataTables)
+	if len(dataTables) == 0 {
```
shivzone, line 123: If there are no changes to AO tables and everythign else is filtered out, how about simply making this a no-op ?

kyeap-vmware: At this point it is not clear yet wether all tables will be filtered out. All `GetBackupDataSet` does is filter out external and foreign data tables. Later on, `FilterTablesForIncremental` will filter out AO tables that have no changes.

bmdoil (approved):
+1

kyeap-vmware:
This is a PR with an updated commit message. Original PR here https://github.com/greenplum-db/gpbackup/pull/532.

PR #538, opened 2021-11-24
Log AccessExclusiveLock held on relation when worker cannot obtain AccessShareLock

bmdoil:
In case we aren't able to grab an AccessShareLock on the table being backed up, we terminate the worker thread and its corresponding connection and let the main worker thread (Worker 0) handle them. This can happen when there is another session with an AccessExclusiveLock on the same table being handled by the worker thread.

When this occurs, log the AccessExclusiveLock being held on the table.

```
[WARNING]:-Worker 1 could not acquire AccessShareLock for table public.foo. 
Terminating worker and deferring table to main worker thread.
[WARNING]:-Locks held on table public.foo: [{"Oid":70037,"Database":"testdb","Relation":"public.foo",
"Mode":"AccessExclusiveLock","Application":"","Granted":"false","User":"bdoil","Pid":"2080006"}]
```

### DIFF
--- a/backup/backup.go
+++ b/backup/backup.go
@@ -1,6 +1,7 @@
 package backup
 
 import (
+	"encoding/json"
 	"errors"
 	"fmt"
 	"os"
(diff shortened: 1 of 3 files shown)


kyeap-vmware (approved):
LGTM

shivzone (approved):

PR #542, opened 2021-12-25
update a mistake in readme.md

ChrisYuan:
I checked the Makefile and there is no 'build_mac' tag, so I make this change.


bmdoil (approved):
Nice catch, thank you for the contribution.

kyeap-vmware (approved):

PR #545, opened 2022-02-01
Feature: Enable synchronized database snapshot export for gpbackup

bmdoil:
This feature enables the use of synchronized database snapshots across parallel gpbackup workers to ensure data consistency and improve resilience when using multiple jobs.

This also extends the snapshot functionality of pg_dump in two ways.
1) The backup is not terminated if a worker connection fails to obtain a lock.
2) Tables are backed up in batches within transactions so fewer locks must be held in total

Requires Greenplum Database >= 6.21.0

Details:
- The lead worker connection creates the database snapshot by calling pg_export_snapshot during connection pool 
initialization. The lead worker then acquires ACCESS SHARE locks on each table in the backup
- The parallel worker connections each begin a transaction and import the snapshot
- Tables in the backup set are placed in a task queue
- Parallel worker connections will each select a table from the task queue, then ensure the snapshot is set. To avoid deadlocks, try to acquire an ACCESS SHARE lock on the table. If the lock fails, table is placed in a deferred queue, transaction rolled back,  and a new table is selected. If the lock is granted, the table is copied out and the tablesInXact count is incremented for that worker. Tables are batched in groups of 10 before the transaction is committed and the locks are released.
- In a separate goroutine, the lead worker scans the deferred queue and backs up these tables, since the worker already holds the ACCESS SHARE locks.
- Since transactions are committed in batches during the backup, there are fewer concurrent locks held at any particular time, reducing the likelihood of shared memory issues.

If distributed snapshot is not supported, each parallel worker must maintain a single valid transaction during the backup. If an attempt to acquire a lock on a table fails, the table is deferred to the lead worker since worker N no longer has a valid distributed snapshot it must be terminated.

This commit also merges logic for backupDataForAllTablesCopyQueue into backupDataForAllTables for an easier to maintain codepath and cleans up some of the tests.

Discussion: https://groups.google.com/a/greenplum.org/g/gpdb-dev/c/C6cY8yIbcps


backup/data.go
```diff
@@ -112,21 +112,19 @@ func BackupSingleTableData(table Table, rowsCopiedMap map[uint32]int64, counters
...
+* If supported by the database, a synchronized database snapshot is used
+* to sync all parallel workers' view of the database and ensure data consistency across workers
+* Tables are then batched in groups of MAX_TABLES_PER_XACT in each transaction to reduce concurrent locks held.
+* Attempt to acquire ACCESS SHARE lock on table to avoid potential deadlock. If worker is unable to acquire a lock,
+* defer table to worker 0, which holds all ACCESS SHARE locks for the backup set.
+* If synchronized snapshot is not supported and worker is unable to acquire a lock, the
+* worker must be terminated because the session no longer has a valid distributed snapshot
+*/
+func backupDataForAllTables(tables []Table) []map[uint32]int64 {
+	const MAX_TABLES_PER_XACT = 10 // TODO: Determine if this should be tunable
```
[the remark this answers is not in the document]
bmdoil: Good question, 10 is arbitrary and I chose it to start this conversation on whether it's necessary to batch COPY FROM in transactions. From testing, the run time difference from resetting the transaction is small. Are there any possible concerns around the number of transactions generated for large backup sets? Say 100k+ small/medium sized tables on a weekly or even daily backup schedule? My thought is no but I'd like confirmation.

backup/data.go
```diff
@@ -145,23 +143,35 @@ func backupDataForAllTablesCopyQueue(tables []Table) []map[uint32]int64 {
...
-				if table.SkipDataBackup() {
-					gplog.Verbose("Skipping data backup of table %s because it is either an external or foreign table.", table.FQN())
-					oidMap.Store(table.Oid, Complete)
-					continue
+				if backupSnapshot != "" && connectionPool.Tx[whichConn] == nil {
+					err := SetSynchronizedSnapshot(connectionPool, whichConn, backupSnapshot)
+					if err != nil {
+						gplog.Warn("Worker %d error setting snapshot", whichConn)
+						oidMap.Store(table.Oid, Deferred)
+						xactError = connectionPool.Rollback(whichConn) //TODO: Check for interrupt before each commit/rollback?
```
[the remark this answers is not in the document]
bmdoil: Yes, although I'm wondering how defensive we need to be. If we receive a term signal after the check, then connections are closed in DoCleanup and we'll try to commit/rollback on a closed connection. That should be fine using Rollback/Commit instead of MustRollback/MustCommit, which has been the convention, and properly handling xactError.

backup/data.go
```diff
@@ -145,23 +143,35 @@ func backupDataForAllTablesCopyQueue(tables []Table) []map[uint32]int64 {
...
-
-				if table.SkipDataBackup() {
-					gplog.Verbose("Skipping data backup of table %s because it is either an external or foreign table.", table.FQN())
-					oidMap.Store(table.Oid, Complete)
-					continue
+				if backupSnapshot != "" && connectionPool.Tx[whichConn] == nil {
+					err := SetSynchronizedSnapshot(connectionPool, whichConn, backupSnapshot)
+					if err != nil {
+						gplog.Warn("Worker %d error setting snapshot", whichConn)
+						oidMap.Store(table.Oid, Deferred)
```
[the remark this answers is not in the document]
bmdoil: Changed the warning to fatal.

backup/data.go
```diff
@@ -175,40 +185,57 @@ func backupDataForAllTablesCopyQueue(tables []Table) []map[uint32]int64 {
...
+					}
 				}
-
 				err = BackupSingleTableData(table, rowsCopiedMaps[whichConn], &counters, whichConn)
 				if err != nil {
 					copyErr = err
+					break
+				} else {
+					oidMap.Store(table.Oid, Complete)
+					atomic.AddInt32(&tablesInXact, 1)
```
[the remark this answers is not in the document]
bmdoil: Removed tracking of tables in transaction

backup/data.go
```diff
@@ -157,11 +161,11 @@ func backupDataForAllTablesCopyQueue(tables []Table) []map[uint32]int64 {
 					counters.ProgressBar.(*pb.ProgressBar).NotPrint = true
 					return
 				}
-
-				if table.SkipDataBackup() {
-					gplog.Verbose("Skipping data backup of table %s because it is either an external or foreign table.", table.FQN())
-					oidMap.Store(table.Oid, Complete)
-					continue
```
soumyadeep2007, line 164: We seem to have lost this functionality of skipping?

backup/wrappers.go
```diff
@@ -38,25 +38,36 @@ func SetLoggerVerbosity() {
...
+	gplog.Verbose(fmt.Sprintf("Initializing %d database connections", numConns))
 	connectionPool.MustConnect(numConns)
+
 	utils.ValidateGPDBVersionCompatibility(connectionPool)
 	InitializeMetadataParams(connectionPool)
+	// Begin transactions, initialize the synchronized snapshot, and set session GUCs
 	for connNum := 0; connNum < connectionPool.NumConns; connNum++ {
-		connectionPool.MustExec(fmt.Sprintf("SET application_name TO 'gpbackup_%s'", timestamp), connNum)
-		// BEGIN TRANSACTION
 		connectionPool.MustBegin(connNum)
```
soumyadeep2007, line 61: Can you add `// Begins transaction in repeatable read` here?

backup/data.go
```diff
@@ -112,21 +112,18 @@ func BackupSingleTableData(table Table, rowsCopiedMap map[uint32]int64, counters
...
-		if table.SkipDataBackup() {
-			numExtOrForeignTables++
-		}
-	}
-	counters := BackupProgressCounters{NumRegTables: 0, TotalRegTables: int64(len(tables)) - numExtOrForeignTables}
+/*
+* Iterate through tables, backup data from each table in set.
+* If supported by the database, a synchronized database snapshot is used
+* to sync all parallel workers' view of the database and ensure data consistency across workers
+* Tables are then batched in groups of MAX_TABLES_PER_XACT in each transaction to reduce concurrent locks held.
```
soumyadeep2007, line 119: About the number of tables to take up inside a single transaction for each worker:

* We should update the comment if we are not going to do batching. `Each table is dumped in a separate transaction. This reduces the number of concurrent locks held, as we would relinquish the lock at transaction end`
* If we decide on a number > 1, we will have to respect the `max_locks_per_transaction` GUC.
* For now, since user tables in GP can be assumed to be sufficiently large, the time taken to `COPY` them out should dominate transactional overhead (which is incurred while setting the snapshot). Since we don't have a way to distribute work based on the size of the tables, we can't adopt a strategy of deciding the batch size based on the size of the tables involved. This is perhaps something we can pursue down the line.
* Naive q: Are partition hierarchies represented in a flat fashion with all members showing up in the array of tables? I ask because if dumping the root => dumping the entire hierarchy, we could be doing too much in one transaction.

bmdoil: >  Are partition hierarchies represented in a flat fashion with all members showing up in the array of tables? 

This is the case if `--leaf-partition-data` is used otherwise it's only the parent.

soumyadeep2007: > This is the case if --leaf-partition-data is used otherwise it's only the parent.

Hmmm, so if this flag is not used, we would be using one transaction for the entire partition hierarchy and then only release all those locks once we are done with all of them?

bmdoil: > only release all those locks once we are done with all of them

Correct, there is room for improvement there but would require a more significant refactor of how we gather and handle partition table info.

soumyadeep2007: Sounds good. Lets call it out with a comment for now.

backup/data.go
```diff
@@ -112,21 +112,18 @@ func BackupSingleTableData(table Table, rowsCopiedMap map[uint32]int64, counters
...
+* Attempt to acquire ACCESS SHARE lock on table to avoid potential deadlock. If worker is unable to acquire a lock,
+* defer table to worker 0, which holds all ACCESS SHARE locks for the backup set.
+* If synchronized snapshot is not supported and worker is unable to acquire a lock, the
+* worker must be terminated because the session no longer has a valid distributed snapshot
+*/
+func backupDataForAllTables(tables []Table) []map[uint32]int64 {
```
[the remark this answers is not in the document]
bmdoil: I agree that the `backupDataForAllTables` is pretty ugly right now and needs a refactor. I did poke at pulling out common parts but it wasn't as straightforward as I hoped. I think that effort deserves more time and its own PR.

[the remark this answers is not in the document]
soumyadeep2007: Okay lets add a FIXME for now and track this internally.

end_to_end/locks_test.go
```diff
@@ -0,0 +1,418 @@
...
+	})
+	It("runs gpbackup with jobs flag and COPY deadlock handling occurs", func() {
+		if useOldBackupVersion {
+			Skip("This test is not needed for old backup versions")
+		}
+		// Acquire AccessExclusiveLock on public.foo to block gpbackup when it attempts
+		// to grab AccessShareLocks before its metadata dump section.
+		backupConn.MustExec("BEGIN; LOCK TABLE public.foo IN ACCESS EXCLUSIVE MODE")
+
+		// Execute gpbackup with --jobs 9 since there are 9 tables to back up
```
jmcatamney, line 32: Nit: This copy pasted comment and the one on line 295 give the wrong number of tables, it should be 10 and 2.

jmcatamney (approved):

PR #548, opened 2022-04-01
fix a typo in file: backup/validate.go

ChrisYuan:
There is a typo in file backup/validate.go, miss a blank space, may cause log like this: theprevious backup. So I just fix it.


kyeap-vmware (approved):
nice catch

PR #550, opened 2022-04-04
Explicitly cast timezone to text in GetRoles function

bmdoil:
If an implicit cast is created
```
CREATE CAST (timestamp without time zone AS text) WITH FUNCTION
pg_catalog.text(timestamp without time zone) AS IMPLICIT;
```

the GetRoles function will error with
```
ERROR: operator is not unique: timestamp without time zone || unknown
(SQLSTATE 42725)
```

this commit resolves that issue by explicitly casting the result value of timezone(zone, timestamp) to text.


jmcatamney (approved):
LGTM, though remember to un-Focus the new test before merging.

PR #554, opened 2022-04-18
Allow backups to be restored to a cluster of a different size

jmcatamney:
This commit introduces the ability to restore a backup to a cluster that
has a different number of segments than the cluster which was originally
backed up.

Restoring to a larger cluster (e.g. back up 2 segments and restore to 3
segments) loads each segment's data to the segment with the corresponding
content ID, then redistributes the data.

Restoring to a smaller cluster (e.g. back up 5 segments and restore to 3
segments) restores segment data in a round-robin manner where segments
with a content ID greater than the highest one in the destination cluster
"wrap around" to content 0, restoring to one batch of segments at a time
until all the data is loaded, then redistributes the data.

Any cases where an N-to-N restore is performed (including the default case
of a normal restore) use COPY ON SEGMENT to load the data; other cases use
goroutines to load data through individual COPY commands executed on each
segment in parallel.

The --resize-cluster flag is introduced with this commit, and must be passed
when restoring to a different-sized cluster in order to avoid any accidental
attempts to restore to a cluster of a different size.  It also introduces
the SegmentCount attribute to the configuration file, which must be present
in order to perform a resize restore; backups taken with older versions of
gpbackup may be restored with this flag if the user adds the parameter to the
config file manually, but taking a new backup is recommended.


restore/data.go
```diff
@@ -54,7 +164,115 @@ func CopyTableIn(connectionPool *dbconn.DBConn, tableName string, tableAttribute
...
 	}
 	numRows, _ := result.RowsAffected()
-	return numRows, err
+	return numRows, nil
+}
+
+// Load backup data to individual segments using one goroutine persegment.  Used
+// when restoring to a cluster that is of a different size than the cluster on
+// which a backup was taken.
+func loadTableDataUsingParallelCopy(query string, tableName string, contentMap map[int]int) (int64, error) {
```
jimmyyih, line 173: I'm a bit hesitant on this implementation since it does many individual utility-mode COPY commands which skips things like twophase commit and recording into the distributed log (no distributed transaction id)... but at least the data redistribution ALTER command afterwards will kinda add all that.  Hence it should probably be okay but there should be documentation on expectations when there is a restore failure.  If one utility-mode COPY fails, there are no ROLLBACK calls nor are there any query cancels.  The user should be notified that the table is only partially restored and would need to manually clean up the failed table(s).

restore/data.go
```diff
@@ -54,7 +164,115 @@ func CopyTableIn(connectionPool *dbconn.DBConn, tableName string, tableAttribute
...
+
+	var numRows int64
+	for _, rowsCopied := range rowsCopiedMap {
+		numRows += rowsCopied
+	}
+	return numRows, nil
+}
+
+// Connects to specific segment in utility mode
+func setupSegmentUtilityConnection(dbname string, port int) *dbconn.DBConn {
```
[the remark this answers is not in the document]
jmcatamney: We had that as a helper function initially to defer figuring out how best to handle utility mode connections, but I've moved it over to dbconn in the PR [here](https://github.com/greenplum-db/gp-common-go-libs/pull/65), take a look.

end_to_end/end_to_end_suite_test.go
```diff
@@ -1906,4 +1914,93 @@ var _ = Describe("backup and restore end to end tests", func() {
 		gprestore(gprestorePath, restoreHelperPath, timestamp,
 			"--redirect-db", "document_db")
 	})
+	Describe("Restore to a different-sized cluster", func() {
+		// The backups for these tests were taken on GPDB version 6.20.3+dev.4.g9a08259bd1 build dev
```
bmdoil, line 1918: Can we include the flags used for the backups as a comment? Do we need to also test `--single-data-file` here in the e2e section?

restore/data.go
bmdoil: Was directing the data files to named pipes and using COPY ON SEGMENT considered and ruled out? Interested in the limitations of that method and how utility mode COPY commands is the preferred implementation.

end_to_end/end_to_end_suite_test.go
```diff
@@ -1906,4 +1914,93 @@ var _ = Describe("backup and restore end to end tests", func() {
...
+			mustRunCommand(command)
+
+			gprestoreCmd := exec.Command(gprestorePath,
+				"--timestamp", "20220415160842",
+				"--redirect-db", "restoredb",
+				"--backup-dir", path.Join(backupDir, "5-segment-db"),
+				"--on-error-continue")
+			output, err := gprestoreCmd.CombinedOutput()
+			Expect(err).To(HaveOccurred())
+			Expect(string(output)).To(ContainSubstring("Cannot restore a backup taken on a cluster with 5 segments to a cluster with 3 segments unless the --resize-cluster flag is used."))
```
bmdoil, line 2003: Also testing common error conditions during the restore would be useful. What can we expect if gprestore is terminated? If data file on one segment is corrupt?

restore/data.go
jmcatamney: Using COPY ON SEGMENT with empty data files on certain segments or pipes or some other method was one of the approaches considered when we were exploring resize restores originally way back when, and it didn't work out for various reasons; when we considered it again for this story, it looked like it would make the logic more hacky and complicated than the goroutine solution.

However, that _was_ before I found a way to unify the normal and resize code paths for the batch restores, so I'm going to look into that this afternoon and see if I can find a way to use COPY ON SEGMENT that works and is maintainable.

restore/data.go
```diff
@@ -54,7 +163,79 @@ func CopyTableIn(connectionPool *dbconn.DBConn, tableName string, tableAttribute
...
+	}
+	workerPool.Wait()
+
+	if len(copyErrs) != 0 {
+		return 0, <-copyErrs // We only return one error, so we don't care which segment's error this is
+	}
+
+	var numRows int64
+	for _, rowsCopied := range rowsCopiedMap {
+		numRows += rowsCopied
```
kyeap-vmware, line 222: We can use an atomic counter instead of recording number of copied rows and adding them up in the end.

restore/data.go
```diff
@@ -54,7 +163,79 @@ func CopyTableIn(connectionPool *dbconn.DBConn, tableName string, tableAttribute
...
+		workerPool.Add(1)
+		go func(backupContent int, restoreContent int) {
+			defer workerPool.Done()
+			gplog.Verbose(`Executing "%s" on segment %d`, query, restoreContent)
+			gplog.Verbose("Restoring data for %s from backup segment %d into restore segment %d", tableName, backupContent, restoreContent)
+
+			segmentConn := setupSegmentUtilityConnection(dbName, restoreContent)
+			defer segmentConn.Close()
+
+			segidRegex := regexp.MustCompile(`<SEGID>`)
```
kyeap-vmware, line 193: This line could be moved out of the loop if we want.

restore/data.go
jmcatamney: After some investigation into the COPY ON SEGMENT approach and an offline conversation with Jimmy and Ashwin, it looks like the goroutine solution is as "safe" as COPY ON SEGMENT from a server perspective (though we'll want to add a couple of docs notes around behavior when a user is querying data during the restore and when recovering from a failed restore), and as I wasn't able to find a good alternative solution that was as straightforward and overhead-less as the goroutine solution, I'm going to go ahead with the current approach for now and we can potentially look into refactoring it to use COPY ON SEGMENT at a later point.

bmdoil (approved):
LGTM, 

Before merging we should tag https://github.com/greenplum-db/gp-common-go-libs/commit/21efdd1fe7efb74bae03dd7dd5e35fcba3e3fd67 as `v1.0.7` and update our go.mod accordingly.

i-amikron:
Hi! 
Can we understand, when this feature will be implemented?
And will this feature work with backups that was created in single-data-file mode?
Thanks!

jmcatamney:
@i-amikron 

> Hi! Can we understand, when this feature will be implemented?

It's still under development, as I had to change the implementation after discovering a bug in GPDB that rendered the prior approach unusuable.  I hope to have it completed in the next few weeks.

> And will this feature work with backups that was created in single-data-file mode? Thanks!

Yes it will.

helper/restore_helper.go
```diff
@@ -77,27 +82,67 @@ func (r *RestoreReader) copyData(num int64) (int64, error) {
...
+		bytesRead, err = io.Copy(writer, r.bufReader)
+	}
+	return bytesRead, err
+}
+
 func doRestoreAgent() error {
-	segmentTOC := toc.NewSegmentTOC(*tocFile)
-	tocEntries := segmentTOC.DataEntries
+	// We need to track various values separately per content for resize restore
+	// TODO: stick these in a per-content struct once it's functional so it's not so hacky
```
AJR-VMware, line 99: Do we want to make this change still, or is it good as-is?

helper/restore_helper.go
```diff
@@ -77,27 +82,67 @@ func (r *RestoreReader) copyData(num int64) (int64, error) {
...
+			if isResizeRestore && c >= *origSize {
+				break
+			}
+			tocFileForContent := replaceContentInFilename(*tocFile, c)
+			segmentTOC[c] = toc.NewSegmentTOC(tocFileForContent)
+			tocEntries[c] = segmentTOC[c].DataEntries
+			c += *destSize
+		}
 	}
-	log(fmt.Sprintf("Using reader type: %s", reader.readerType))
```
AJR-VMware, line 100: Is this removed intentionally?  Seems like it might be useful to have still.

helper/restore_helper.go
```diff
@@ -219,15 +335,34 @@ func doRestoreAgent() error {
...
+	name = strings.ReplaceAll(name, fmt.Sprintf("gpbackup_%d", *content), fmt.Sprintf("gpbackup_%d", contentToRestore))
+	nameParts := strings.Split(name, ".")
+	filename := fmt.Sprintf("%s_%d", nameParts[0], oid)
+	if len(nameParts) == 2 {
+		filename = fmt.Sprintf("%s.%s", filename, nameParts[1])
+	}
+	return filename
+}
+
+// TODO: Hackish way to get the next ToC file, need to pass in a better path for interpolation
```
AJR-VMware, line 349: Are either of these worth cleaning up for the release?  IMO they're fine as-is but worth checking on TODO's with "hacky" in the comment.

helper/restore_helper.go
```diff
@@ -299,15 +434,16 @@ func getRestorePipeWriter(currentPipe string) (*bufio.Writer, *os.File, error) {
...
+func startRestorePluginCommand(fileToRead string, toc *toc.SegmentTOC, oidList []int) (io.Reader, bool, error) {
 	isSubset := false
 	pluginConfig, err := utils.ReadPluginConfig(*pluginConfigFile)
 	if err != nil {
 		logError(fmt.Sprintf("Error encountered when reading plugin config: %v", err))
 		return nil, false, err
 	}
 	cmdStr := ""
-	if pluginConfig.CanRestoreSubset() && *isFiltered && !strings.HasSuffix(*dataFile, ".gz") && !strings.HasSuffix(*dataFile, ".zst") {
+	// TODO confirm that "toc != nil" is appropriate
```
AJR-VMware, line 445: Nit: I believe toc != nil is fine?

restore/data.go
```diff
@@ -41,29 +46,64 @@ func CopyTableIn(connectionPool *dbconn.DBConn, tableName string, tableAttribute
...
+		}
+		rowsLoaded, _ := result.RowsAffected()
+		numRows += rowsLoaded
 	}
-	numRows, _ := result.RowsAffected()
+
 	return numRows, err
 }
 
+func setupSegmentUtilityConnection(dbname string, content int) *dbconn.DBConn {
```
AJR-VMware, line 84: This doesn't seem to be called anywhere.  Should it be removed?

jmcatamney: Yep, that was only needed for the goroutine version, good catch.

helper/restore_helper.go
jmcatamney: Yeah, these TODOs (and the other one you tagged) are fine now, I just didn't remove the comments after I refactored them before.  I'll take them out.

helper/restore_helper.go
jmcatamney: Whoops, I removed it as part of cleaning up some debugging logs, I'll put that back.

jmcatamney: Just kidding, that wasn't removed, it was actually moved into the single data file loop along with the start and end assignments (see line 202), which is where it should be.

AJR-VMware (approved):
Approved, feedback left in separate comments and addressed.

khuddlefish (approved):
I won't claim I understood everything but nothing stood out to me as looking weird. Let's do it :)

helper/helper.go
```diff
@@ -88,7 +91,7 @@ func InitializeGlobals() {
 	backupAgent = flag.Bool("backup-agent", false, "Use gpbackup_helper as an agent for backup")
 	content = flag.Int("content", -2, "Content ID of the corresponding segment")
 	compressionLevel = flag.Int("compression-level", 0, "The level of compression. O indicates no compression. Range of valid values depends on compression type")
-	compressionType = flag.String("compression-type", "gzip", "The type of compression. Valid values are 'gzip', 'zstd'")
+	compressionType = flag.String("compression-type", "gzip", "The type of compression. Valid values are 'gzip', 'zstd', 'none'")
```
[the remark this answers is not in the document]
jmcatamney: We actually don't need to check either one, since I ended up not doing the compression processing on the helper side.  Thanks for the catch!

integration/helper_test.go
```diff
@@ -52,7 +52,7 @@ here is some data
 )
 
 func gpbackupHelper(helperPath string, args ...string) *exec.Cmd {
-	args = append([]string{"--toc-file", tocFile, "--oid-file", oidFile, "--pipe-file", pipeFile, "--content", "1"}, args...)
+	args = append([]string{"--toc-file", tocFile, "--oid-file", oidFile, "--pipe-file", pipeFile, "--content", "1", "--single-data-file"}, args...)
```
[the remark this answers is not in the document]
jmcatamney: It does set it to always run in single-file mode, to preserve the previous behavior for those tests.  Since non-single-data-file behavior only happens as part of a resize restore, I figured that the end-to-end tests covered that sufficiently and duplicating all the unit tests wouldn't gain us much.  I could add some unit tests if you're concerned, though.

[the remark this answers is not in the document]
khuddlefish: If that's the previous test behavior then it should be ok

PR #563, opened 2022-07-07
 Exclude trigger constraints from predata backup

AJR-VMware:
Predata backup syntax does not need to print out trigger constraints.
The ALTER TABLE ADD CONSTRAINT syntax currently being output is invalid.
Trigger constraints are already handled correctly by postdata backup.
Added a trigger constraint to the end-to-end test suite in gbdp_objects.sql.
Also fixed a minor typo in unit test comment.


khuddlefish (approved):
LGTM. Code change makes sense, your comment is good, and you added testing.

PR #566, opened 2022-07-18
Require rsync on system

AJR-VMware:
Add utility function for detecting command on PATH.
Add checks for presence of rsync before its used.

Convert image used for building binaries in CI
Image used to build gpbackup go binaries was missing rsync.
Udpated to use a heavier build image.

### DIFF
--- a/utils/agent_remote.go
+++ b/utils/agent_remote.go
@@ -40,6 +40,11 @@ func CreateSegmentPipeOnAllHosts(oid string, c *cluster.Cluster, fpInfo filepath
 }
 
 func WriteOidListToSegments(oidList []string, c *cluster.Cluster, fpInfo filepath.FilePathInfo) {
+	rsync_exists := CommandExists("rsync")
+	if !rsync_exists {
+		gplog.Fatal(errors.New("Failed to find rsync on PATH. Please ensure rsync is installed."), "")
+	}
+
 	localOidFile, err := operating.System.TempFile("", "gpbackup-oids")
 	gplog.FatalOnError(err, "Cannot open temporary file to write oids")
 	defer func() {
--- a/utils/util.go
+++ b/utils/util.go
@@ -8,6 +8,7 @@ package utils
 import (
 	"fmt"
 	"os"
+	"os/exec"
 	"os/signal"
 	"regexp"
 	"strings"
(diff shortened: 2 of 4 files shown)


utils/util.go
```diff
@@ -140,9 +146,9 @@ func InitializeSignalHandler(cleanupFunc func(bool), procDesc string, termFlag *
 		fmt.Println() // Add newline after "^C" is printed
 		switch sig {
 		case unix.SIGINT:
-				gplog.Warn("Received an interrupt signal, aborting %s", procDesc)
+			gplog.Warn("Received an interrupt signal, aborting %s", procDesc)
 		case unix.SIGTERM:
-				gplog.Warn("Received a termination signal, aborting %s", procDesc)
+			gplog.Warn("Received a termination signal, aborting %s", procDesc)
```
bmdoil, line 151: This seems unrelated.

utils/agent_remote.go
```diff
@@ -40,6 +40,11 @@ func CreateSegmentPipeOnAllHosts(oid string, c *cluster.Cluster, fpInfo filepath
 }
 
 func WriteOidListToSegments(oidList []string, c *cluster.Cluster, fpInfo filepath.FilePathInfo) {
+	rsync_exists := CommandExists("rsync")
+	if !rsync_exists {
+		gplog.Fatal(errors.New("Failed to find rsync on PATH.  Please ensure rsync is installed."), "")
```
bmdoil, line 45: nit: remove 2nd space before "Please"

bmdoil (commented):
Can you updated the commit message to remove these lines since they no longer apply.
```
try removing autoreq disable

add requirement to ubuntu installer

Remove rsync from reqs
```

utils/util.go
AJR-VMware: I didn't do this one manually.  Must have been autoformat on save?  I'm not sure.  I'll remove this.  Good catch, thanks!

AJR-VMware: Fixed!

utils/agent_remote.go
AJR-VMware: Fixed!

bmdoil (approved):
+1

PR #569, opened 2022-07-29
Refactor gpbackup scale tests.

AJR-VMware:
This approach uses tpc-h data loaded into a gcs bucket to
test performance of gpbackup and restore at moderate scale,
under a variety of configurations.

Correctness is tested by comparing rowcounts where possible,
and comparing against a pre-stored metadata file in other cases.

Runtime is assessed against a rolling average for each test,
with all runs and pass thresholds saved in a GCP Cloud SQL
instance.


ci/scale/sql/pull_rowcount.sql
```diff
@@ -0,0 +1,12 @@
...
+  table_name, 
+  big.cnt_rows(table_name)
+FROM 
+  information_schema.tables
+WHERE
+  table_schema = 'big'
+  AND table_type='BASE TABLE'
+ORDER BY
+  table_schema,
+  table_name
```
bmdoil, line 12: nit: missing newline

ci/scale/scale_pipeline.yml
```diff
@@ -0,0 +1,428 @@
...
+  type: registry-image
+  source:
+    repository: gcr.io/data-gpdb-public-images/gpdb5-centos7-build-test
+    tag: latest
+
+- name: gpdb7-centos7-image
+  type: registry-image
+  source:
+    repository: gcr.io/data-gpdb-public-images/gpdb7-centos7-build
+    tag: latest
```
bmdoil, line 59: Ideally we'd only need to use 1 type of image for everything in this pipeline. 
Can `centos6-image` be replaced by `centos7-image`?
What dependencies in `gpdb7-centos7-image` specifically are needed for `scale-test-slack-notify`?

ci/scripts/scale_perf_tests.bash
```diff
@@ -0,0 +1,346 @@
...
+source env.sh
+# set format for logging
+export TIMEFORMAT="TEST RUNTIME: %E"
+export RESULTS_LOG_FILE=${RESULTS_LOG_FILE}
+
+# set parameters for reference time DB
+export RESULTS_DATABASE_HOST=${RESULTS_DATABASE_HOST}
+export RESULTS_DATABASE_USER=${RESULTS_DATABASE_USER}
+export RESULTS_DATABASE_NAME=${RESULTS_DATABASE_NAME}
+export RESULTS_DATABASE_PASSWORD=${RESULTS_DATABASE_PASSWORD}
```
bmdoil, line 34: Can we add the steps you did to create the results database into a README in `ci/scale` for the future in case it needs to be recreated?

ci/scripts/scale_test_slack_notify.bash
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+set -x
+set -euo pipefail
+
+yum install -y postgresql-devel python-dev
+pip3 install psycopg2-binary
+pip3 install slack-sdk
```
bmdoil, line 7: It looks like we're using `gpdb7-centos7-build` for the slack-notify job which is pretty heavy, then adding to it. Also pip has caused us a lot of headaches in the past when run as part of a job. 

We should build a container that resolves all the dependencies we need for our scale testing. We can add a follow-up story for this.

bmdoil (commented):
We should add a README to `ci/scale` with a high level overview of the scale testing framework setup and list of tests being run. This will help share context for future/outside contributors. 

Also, does `valid_metadata.sql` need to be committed to the repo? Is the idea that any change to the scaletestdb DDL will be accompanied by a change to `valid_metadata.sql`?

AJR-VMware:
> Also, does `valid_metadata.sql` need to be committed to the repo? Is the idea that any change to the scaletestdb DDL will be accompanied by a change to `valid_metadata.sql`?

Yeah, this was the approach we discussed where, in order to test whether the metadata being backed up/restored is valid, we'll need to have a "known-good" metadata file for our test schema.  This approach does, unfortunately, mean that any changes to the "wide" schema DDL would require regenerating and re-validating the valid_metadata.sql file.

ci/scale/scale_pipeline.yml
AJR-VMware: The Slack SDK for python2 has not been well maintained.  I was having a heck of a time trying to get it slack notifications working properly on python2.  I figured adding python3 to our older centos6 images was likely to break some of the scripts that are expecting a python2 environment.

ci/scripts/scale_perf_tests.bash
AJR-VMware: I included the DDL for the database into the repo, and manually created and configures the cloud instance.  Do you mean the manual steps for the creation+configuration?

ci/tasks/scale-tests-slack-notify.yml
```diff
@@ -0,0 +1,18 @@
...
+inputs:
+- name: gpbackup
+
+params:
+  RESULTS_DATABASE_HOST: 
+  RESULTS_DATABASE_USER:
+  RESULTS_DATABASE_NAME:
+  RESULTS_DATABASE_PASSWORD:
+  SLACK_WEBHOOK_URL:
+  OS: RHEL
```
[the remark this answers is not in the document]
AJR-VMware: It wasn't!  Copy/paste/edit from the original scale test yaml, forgot to clean that out.  Good catch!

ci/scripts/scale_perf_tests.bash
bmdoil: >Do you mean the manual steps for the creation+configuration?

Yes, just quick step-by-step.

ci/scripts/scale_perf_tests.bash
```diff
@@ -0,0 +1,346 @@
...
+gpload -f /home/gpadmin/gpload_yaml/region.yml
+gpload -f /home/gpadmin/gpload_yaml/supplier.yml
+
+echo "## Capturing row counts for comparison ##"
+psql -d scaletestdb -f /home/gpadmin/pull_rowcount.sql -o /home/gpadmin/rowcounts_orig.txt
+
+
+#####################################################################
+##################################################################### 
+echo "## Performing single-data-file, --no-compression, --copy-queue-size 8 backup test ##"
```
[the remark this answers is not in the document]
bmdoil: We'll track this effort internally.

ci/scale/scale_pipeline.yml
bmdoil: Understood. We'll track the effort to create a custom image internally.

ci/scripts/scale_test_slack_notify.bash
bmdoil: Will track this internally.

bmdoil (approved):
LGTM

PR #571, opened 2022-08-02
Migrate DP secrets to Vault

AJR-VMware:
Update secrets in template to use Vault paths.
Update secrets in /scale pipeline to use Vault paths.
Update gen_pipeline.py to print corrected fly invocation.
Update gen_pipeline.py to find/replace dev/prod token as needed.
Add README.md to /ci to explain the approach taken.


khuddlefish:
Secrets changing stuff looks good to me. Why are we removing the regression pipeline? Can that be split out into a separate commit with a message explaining the context?

PR #572, opened 2022-08-04
Add more logging to gpbackup and gprestore

AJR-VMware:


utils/io.go
```diff
@@ -92,10 +92,11 @@ func CopyFile(src, dest string) error {
 		var content []byte
 		content, err = ioutil.ReadFile(src)
 		if err != nil {
+			gplog.Error(fmt.Sprintf("Error: %v, encountered when reading file: %s", err, src))
 			return err
 		}
 		return ioutil.WriteFile(dest, content, info.Mode())
 	}
-
+	gplog.Error(fmt.Sprintf("Error: %v, encountered when trying to stat file: %s", err, src))
```
khuddlefish, line 100: Are there ever times when the file we are looking for doesn't exist and that would be ok? You get stat file errors when something isn't there and something we use it as a harmless check for whether something exists already or if it needs to be created.

AJR-VMware: This line is part of CopyFile, and is called only for copying plugin config files.  In the context it's used, not being able to find that file would be an error.  In general, I think if we're trying to copy a file and it doesn't exist I think we'd want to log an error for that.

PR #575, opened 2022-08-25
Ignore dropped or altered types

AJR-VMware:
Types that have been dropped or altered have different syntax.
These do not need to be included in backups, and are not handled
correctly in metadata printout.


khuddlefish:
Nice. Is there anything weird about restoring these or do they work fine with the syntax as long as we ignore the dropped ones?

AJR-VMware:
They work fine in my manual testing, once you ignore the dropped ones.

AJR-VMware:
The syntax currently looks like this, if you happen to have a dropped type.  
```sql
CREATE TYPE public.test_type2 AS (
	aa integer,
	"........pg.dropped.2........" -,
	c text
);
```

jmcatamney (approved):

PR #579, opened 2022-09-27
Move Constraints into sorted dependency backup

AJR-VMware:
Some View definitions may depend on constraints.
Constraints are printed after views, sometimes resulting in invalid SQL.

Move constraints into sorted dependency procedure.
Move constraints without NOT VALID clause back to predata section.
Move dropping of trigger constraints from printing to retrieval.
Split out domain constraints to keep current approach for those.
Fix some formatting and warnings in dependencies.go.

### DIFF
--- a/backup/backup.go
+++ b/backup/backup.go
@@ -254,12 +254,11 @@ func backupPredata(metadataFile *utils.FileWithByteCount, tables []Table, tableO
 
 	retrieveViews(&objects)
 	sequences := retrieveAndBackupSequences(metadataFile, relationMetadata)
-	constraints, conMetadata := retrieveConstraints()
+	domainConstraints := retrieveConstraints(&objects, metadataMap)
 
-	backupDependentObjects(metadataFile, tables, protocols, metadataMap, constraints, objects, sequences, funcInfoMap, tableOnly)
+	backupDependentObjects(metadataFile, tables, protocols, metadataMap, domainConstraints, objects, sequences, funcInfoMap, tableOnly)
 
 	backupConversions(metadataFile)
-	backupConstraints(metadataFile, constraints, conMetadata)
 
 	logCompletionMessage("Pre-data metadata metadata backup")
 }
--- a/backup/predata_shared_test.go
+++ b/backup/predata_shared_test.go
@@ -3,7 +3,6 @@ package backup_test
 import (
 	"database/sql"
 
-	"github.com/greenplum-db/gp-common-go-libs/testhelper"
 	"github.com/greenplum-db/gpbackup/backup"
 	"github.com/greenplum-db/gpbackup/testutils"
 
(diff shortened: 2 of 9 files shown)


backup/predata_shared_test.go
```diff
@@ -18,144 +17,130 @@ var _ = Describe("backup/predata_shared tests", func() {
...
 			It("prints an ADD CONSTRAINT [name] CHECK statement without keyword ONLY for a table with descendants (another table inherits it)", func() {
-				checkConstraint := backup.Constraint{Oid: 0, Name: "check1", ConType: "c", ConDef: sql.NullString{String: "CHECK (VALUE <> 42::numeric)", Valid: true}, OwningObject: "public.tablename", IsDomainConstraint: false, IsPartitionParent: false, ConIsLocal: true}
-				constraints := []backup.Constraint{checkConstraint}
-				backup.PrintConstraintStatements(backupfile, tocfile, constraints, emptyMetadataMap)
-				testutils.AssertBufferContents(tocfile.PostdataEntries, buffer, `ALTER TABLE public.tablename ADD CONSTRAINT check1 CHECK (VALUE <> 42::numeric);`)
+				backup.PrintConstraintStatement(backupfile, tocfile, checkConstraint, objectMetadata)
+				testutils.AssertBufferContents(tocfile.PredataEntries, buffer, `ALTER TABLE public.tablename ADD CONSTRAINT check1 CHECK (VALUE <> 42::numeric);`)
 			})
 		})
+		Context("Constraints with a NOT VALID clause", func() {
```
jmcatamney, line 141: Looks like either the NOT VALID test above should be moved here or this section should be removed; I have no preference either way.

integration/predata_shared_create_test.go
```diff
@@ -61,7 +61,8 @@ var _ = Describe("backup integration create statement tests", func() {
 			fkConstraint             backup.Constraint
 			checkConstraint          backup.Constraint
 			partitionCheckConstraint backup.Constraint
-			conMetadataMap           backup.MetadataMap
+			// conMetadataMap           backup.MetadataMap
```
jmcatamney, line 64: This can be removed.

integration/predata_shared_create_test.go
```diff
@@ -70,7 +71,8 @@ var _ = Describe("backup integration create statement tests", func() {
 			checkConstraint = backup.Constraint{Oid: 0, Schema: "public", Name: "check1", ConType: "c", ConDef: sql.NullString{String: "CHECK (a <> 42)", Valid: true}, OwningObject: "public.testtable", IsDomainConstraint: false, IsPartitionParent: false}
 			partitionCheckConstraint = backup.Constraint{Oid: 0, Schema: "public", Name: "check1", ConType: "c", ConDef: sql.NullString{String: "CHECK (id <> 0)", Valid: true}, OwningObject: "public.part", IsDomainConstraint: false, IsPartitionParent: true}
 			testhelper.AssertQueryRuns(connectionPool, "CREATE TABLE public.testtable(a int, b text) DISTRIBUTED BY (b)")
-			conMetadataMap = backup.MetadataMap{}
+			// conMetadataMap = backup.MetadataMap{}
```
jmcatamney, line 74: This can also be removed.

backup/predata_shared_test.go
AJR-VMware: Yep, I rewrote these and put them in the other spot.  Thanks!

integration/predata_shared_create_test.go
AJR-VMware: Yep!

integration/predata_shared_create_test.go
AJR-VMware: Agreed, good catch.

jmcatamney (approved):

backup/backup.go
```diff
@@ -254,12 +254,11 @@ func backupPredata(metadataFile *utils.FileWithByteCount, tables []Table, tableO
 
 	retrieveViews(&objects)
 	sequences := retrieveAndBackupSequences(metadataFile, relationMetadata)
-	constraints, conMetadata := retrieveConstraints()
+	domainConstraints := retrieveConstraints(&objects, metadataMap)
 
-	backupDependentObjects(metadataFile, tables, protocols, metadataMap, constraints, objects, sequences, funcInfoMap, tableOnly)
+	backupDependentObjects(metadataFile, tables, protocols, metadataMap, domainConstraints, objects, sequences, funcInfoMap, tableOnly)
```
khuddlefish, line 259: I notice you didn't change the variable name inside the backupDependentObjects function. Would probably be good to make that `domainConstraints` also

integration/predata_shared_create_test.go
```diff
@@ -61,7 +61,7 @@ var _ = Describe("backup integration create statement tests", func() {
 			fkConstraint             backup.Constraint
 			checkConstraint          backup.Constraint
 			partitionCheckConstraint backup.Constraint
-			conMetadataMap           backup.MetadataMap
```
khuddlefish, line 64: ^ The describe section name should be changed to be singular

khuddlefish (approved):
Looks pretty good. A few minor variable name change suggestions and possible test coverage suggestions

Do we need a test to cover that we are no longer querying triggers in GetConstraints?
Do we need a test to cover that we are splitting out domainConstraints from retrieveConstraints?

What's the deal with the VALID or NOT VALID ones?

integration/predata_shared_create_test.go
AJR-VMware: Yep!  Thanks!

backup/backup.go
AJR-VMware: Good catch, thank you!

AJR-VMware:
@khuddlefish 
The domainConstraints are still tested in dependencies_test.go Context "PrintDependentObjectStatements", and the non-domain constraints are also tested.  We put the trigger query and isdomain logic into the SQL query, and  the actual queries_shared_test.go  is very lightweight (it's currently one test that mocks the query).  It seems that currently we mostly trust that the queries work as expected, and rely on integration and end-to-end testing for those.   I'm not sure how else we'd want to test the query logic here, I'd be open to suggestions.

The NOT VALID clause was something that support was added for a couple years ago.  A NOT VALID constraint will not be applied to data already in the table when the constraint is added, but is applied to any new data after the fact.  To facilitate this, we had moved all constraints into postdata, so that data loading wouldn't cause a bunch of issues with grandfathered-in data in customer databases.  Obviously, though, we need views to be created in predata and therefore the constraints they rely on need to also be in predata.  My change here splits the difference, and moves constraints back to predata unless they actually contain a NOT VALID clause.  This does leave a corner case where a view relying on a NOT VALID constraint will not work correctly due to the order of processes, but that would have already been broken in the current state of things and seems less likely.  I don't really see a way around this particular corner case, would be open to discussing it and seeing if we could come up with one though.

bmdoil:
> @khuddlefish The domainConstraints are still tested in dependencies_test.go Context "PrintDependentObjectStatements", and the non-domain constraints are also tested. We put the trigger filter and isdomain logic into the SQL query, and the actual queries_shared_test.go is very lightweight (it's currently one test that mocks the query). It seems that currently we mostly trust that the queries work as expected, and rely on integration and end-to-end testing for those. I'm not sure how else we'd want to test the query logic here, I'd be open to suggestions.
> 
> The NOT VALID clause was something that support was added for a couple years ago. A NOT VALID constraint will not be applied to data already in the table when the constraint is added, but is applied to any new data after the fact. To facilitate this, we had moved all constraints into postdata, so that data loading wouldn't cause a bunch of issues with grandfathered-in data in customer databases. Obviously, though, we need views to be created in predata and therefore the constraints they rely on need to also be in predata. My change here splits the difference, and moves constraints back to predata unless they actually contain a NOT VALID clause. This does leave a corner case where a view relying on a NOT VALID constraint will not work correctly due to the order of processes, but that would have already been broken in the current state of things and seems less likely. I don't really see a way around this particular corner case, would be open to discussing it and seeing if we could come up with one though.

I'm curious why we care about NOT VALID at all in the context of backup and restore. It only has an effect when we add a foreign key (not supported in GPDB) or check constraint to an existing table, to avoid an expensive full table scan validating existing tuples meet the constraint. New tuples will always be checked against the constraint. We never restore a table DDL and some data, add a constraint, then restore more data in the same gprestore run, so isn't NOT VALID always a no-op? pg_dump doesn't handle the NOT VALID syntax at all.

AJR-VMware:
>I'm curious why we care about NOT VALID at all in the context of backup and restore. It only has an effect when we add a foreign key (not supported in GPDB) or check constraint to an existing table, to avoid an expensive full table scan validating existing tuples meet the constraint. New tuples will always be checked against the constraint. We never restore a table DDL and some data, add a constraint, then restore more data in the same gprestore run, so isn't NOT VALID always a no-op? pg_dump doesn't handle the NOT VALID syntax at all.

While I didn't make the original change (@kyeap-vmware did and may have some context here?), my assumption would be that the problematic control flow in a gprestore situation would be something like:

```
create predata -> copy in data -> create postdata
```
If the predata included a NOT VALID constraint, then wouldn't copying in the data in the middle step have the constraint applied to all data, even "grandfathered" data from the customer's database?  I assume that's why the constraints were pushed over to postdata in in the first place?

AJR-VMware:
Additional context on why constraints had originally been moved to the postdata section, per @kyeap-vmware .  The approach in this PR still addresses the original problem, but the context is kept here for visibility.

```
IIRC what were seeing is that customer's would sometimes add a constraint after a table has existed for a while. So there was data inside a table that violates the constraint. Previously, the constraint was placed during predata phase. what would then happen is that the table schema would restored with the constraint. But when we tried loading data, because some of the data violates the constraint, Data restore would fail.
So our solution was to move the constraint creation to after the data was loaded into the table. Restore table without constraint, restore data, restore constraint.
```

PR #582, opened 2022-10-13
Reorganize reader open logic

AJR-VMware:
Pre-open readers in map to avoid seek perf penalty.

Pin ginkgo version in Makefile.
v2.2.3 has a bug that is breaking our CI.  Pin to v2.2.0 for now.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,7 +31,7 @@ depend :
 	go mod download
 
 $(GINKGO) : # v1.14.0 is compatible with centos6 default gcc version
-	go install github.com/onsi/ginkgo/v2/ginkgo@latest
+	go install github.com/onsi/ginkgo/v2/ginkgo@v2.2.0
 
 $(GOIMPORTS) :
 	go install golang.org/x/tools/cmd/goimports@latest
--- a/helper/restore_helper.go
+++ b/helper/restore_helper.go
@@ -104,7 +104,8 @@ func doRestoreAgent() error {
 
 	var bytesRead int64
 	var lastError error
-	var reader *RestoreReader
+
+	readers := make(map[int]*RestoreReader)
 
 	oidList, err := getOidListFromFile()
 	if err != nil {
@@ -126,20 +127,32 @@ func doRestoreAgent() error {
 	}
 
 	if *singleDataFile {
-		c := *content
+		contentToRestore := *content
 		segmentTOC = make(map[int]*toc.SegmentTOC)
 		tocEntries = make(map[int]map[uint]toc.SegmentDataEntry)
 		start = make(map[int]uint64)
 		end = make(map[int]uint64)
 		lastByte = make(map[int]uint64)
 		for b := 0; b < batches; b++ {
-			if isResizeRestore && c >= *origSize {
+			// When performing a resize restore, if the content of the file we're being asked to read from
+			// is higher than any backup content, then no such file exists and we shouldn't try to open it.
+			if isResizeRestore && contentToRestore >= *origSize {
 				break
 			}
-			tocFileForContent := replaceContentInFilename(*tocFile, c)
-			segmentTOC[c] = toc.NewSegmentTOC(tocFileForContent)
-			tocEntries[c] = segmentTOC[c].DataEntries
-			c += *destSize
+			tocFileForContent := replaceContentInFilename(*tocFile, contentToRestore)
+			segmentTOC[contentToRestore] = toc.NewSegmentTOC(tocFileForContent)
+			tocEntries[contentToRestore] = segmentTOC[contentToRestore].DataEntries
+
+			filename := replaceContentInFilename(*dataFile, contentToRestore)
+			readers[contentToRestore], err = getRestoreDataReader(filename, segmentTOC[contentToRestore], oidList)
+
+			if err != nil {
+				logError(fmt.Sprintf("Error encountered getting restore data reader for single data file: %v", err))
+				return err
+			}
+			log(fmt.Sprintf("Using reader type: %s", readers[contentToRestore].readerType))
+
+			contentToRestore += *destSize
 		}
 	}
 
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -9,7 +9,7 @@ BACKUP=gpbackup
 RESTORE=gprestore
 HELPER=gpbackup_helper
 BIN_DIR=$(shell echo $${GOPATH:-~/go} | awk -F':' '{ print $$1 "/bin"}')
-GINKGO_FLAGS := -r --keep-going --randomize-suites --randomize-all --no-color
+GINKGO_FLAGS := -r --keep-going --randomize-suites --randomize-all
```
khuddlefish, line 12: Is the removal of `--no-color` possibly why we're seeing blinking? I notice the output of this job is flashing https://dp.ci.gpdb.pivotal.io/teams/main/pipelines/gpbackup_reorganize_reader_opens/jobs/GPDB5/builds/8 while previous runs are not. I see this issue in concourse that only seems to have been resolved somewhat recently https://github.com/concourse/concourse/issues/6542

helper/restore_helper.go
```diff
@@ -126,20 +127,32 @@ func doRestoreAgent() error {
...
+		contentToRestore := *content
 		segmentTOC = make(map[int]*toc.SegmentTOC)
 		tocEntries = make(map[int]map[uint]toc.SegmentDataEntry)
 		start = make(map[int]uint64)
 		end = make(map[int]uint64)
 		lastByte = make(map[int]uint64)
 		for b := 0; b < batches; b++ {
-			if isResizeRestore && c >= *origSize {
+			// When performing a resize restore, if the content of the file we're being asked to read from
+			// is higher than any backup content, then no such file exists and we shouldn't try to open it.
```
khuddlefish, line 138: Nice addition of comments. makes it easier to understand

helper/restore_helper.go
```diff
@@ -172,35 +185,24 @@ func doRestoreAgent() error {
 		contentToRestore := *content
 
 		for b := 0; b < batches; b++ {
-			if isResizeRestore {
-				// When performing a resize restore, if the content of the file we're being asked to read from
-				// is higher than any backup content, then no such file exists and we shouldn't try to open it.
+			if *singleDataFile {
```
khuddlefish, line 188: Does this work for when it's both singleDataFile and isResizeRestore?

helper/restore_helper.go
```diff
@@ -172,35 +185,24 @@ func doRestoreAgent() error {
...
+				start[contentToRestore] = tocEntries[contentToRestore][uint(oid)].StartByte
+				end[contentToRestore] = tocEntries[contentToRestore][uint(oid)].EndByte
+			} else if isResizeRestore {
```
khuddlefish, line 191: So to remind me, helpers only get used in singleDataFile and/or isResizeRestore cases, yes?

helper/restore_helper.go
```diff
@@ -172,35 +185,24 @@ func doRestoreAgent() error {
...
-					} else {
-						filename = constructSingleTableFilename(*dataFile, contentToRestore, oid)
-					}
-					reader, err = getRestoreDataReader(filename, nil, nil)
+					filename := constructSingleTableFilename(*dataFile, contentToRestore, oid)
+
+					// We pre-create readers above for the sake of not re-opening SDF readers.  For MDF we can't 
+					// re-use them but still having them in a map simplifies overall code flow.  We overwrite the 
+					// map entry here intentionally.
+					readers[contentToRestore], err = getRestoreDataReader(filename, nil, nil)
```
khuddlefish, line 200: So it looks like you may unnecessarily overwrite the first reader you created but for future oids that will be desired behavior?

helper/restore_helper.go
```diff
@@ -221,7 +223,6 @@ func doRestoreAgent() error {
 							err = nil
 							goto LoopEnd
 						} else {
-							// keep trying to open the pipe
```
khuddlefish, line 224: This comment seems somewhat helpful

helper/restore_helper.go
AJR-VMware: Yes.  The relevant logic for singleDataFile is handled in the block at line 129.  We need to do some additional per-oid bookkeeping for that case which is handled here.  For non-singleDataFile cases, we do all such bookkeeping in the block at 191, which overrides some of what block@129 does as needed.

helper/restore_helper.go
AJR-VMware: In the non-SDF case, here at 200, the map entries are not pre-generated in the block@129.  Instead they're populated here for each OID.  The map structure isn't strictly necessary at that point, but having it simplifies the code overall.

I suppose "overwrite" may have been the wrong choice of words, and reflected an earlier approach.  We should update that comment I think.

helper/restore_helper.go
AJR-VMware: Fair enough, we waffled a bit on removing it, as it was a working comment from Jamie's earlier work I think?  Regardless, we'll put it back in.

helper/restore_helper.go
AJR-VMware: Yes, I believe that's correct.  @jmcatamney that's right?

helper/restore_helper.go
khuddlefish: neat thanks

helper/restore_helper.go
jmcatamney: Yes, agents are used for single data file restores, resize restores, or both.  Standard multi-data-file restores don't use it.

Makefile
jmcatamney: The blinking is an unrelated issue, unfortunately.  We saw it before removing that flag.

jmcatamney: It looks like there are multiple Ginkgo blinking issues, with the color tag scenario being a separate case from the error scenario we previously observed.  We'll remove `--no-color` again for now to save us the eye strain until Concourse is updated.

PR #588, opened 2022-11-22
Add support of resize restore of replicated tables

AJR-VMware:
* Identify replicated tables in toc
* Write a file with the list of oids for those tables to segments
* Restore agent checks that file and skips re-restoring tables in larger to smaller case
* Add expand table query for replicated tables in smaller to larger case
* Add test cases for smaller to larger, vice versa, and normal restore with replicated tables
* Test cases are done with tarred backups, similar to other resize-cluster tests

Also, adjust SetupTestDBConnSegment to reference dbconn.Connect to correctly connect in utility mode to GPDB6 or GPDB7.


end_to_end/end_to_end_suite_test.go
```diff
@@ -1757,6 +1757,82 @@ LANGUAGE plpgsql NO SQL;`)
...
+		})
+
+		It("Can backup a 9-segment cluster and restore to current cluster with replicated tables", func() {
+			testutils.SkipIfBefore6(backupConn)
+			if useOldBackupVersion {
+				Skip("Resize-cluster was only added in version 1.25")
+			}
+
+		})
+		It("Can backup and restore replicated tables", func() {
```
khuddlefish, line 1833: This is empty. Remove?

khuddlefish (approved):

PR #590, opened 2022-11-23
Rebase fix gp7 incl leaf part

AJR-VMware:
This includes the e2e test leftover from the partition suffix change, and work to correct include filtering and oid list expansion for partitioned tables on GPDB7+.  Details are in the commit messages.

### DIFF
--- a/backup/dependencies.go
+++ b/backup/dependencies.go
@@ -1,6 +1,7 @@
 package backup
 
 import (
+	"fmt"
 	"strings"
 
 	"github.com/greenplum-db/gp-common-go-libs/dbconn"
@@ -144,8 +145,15 @@ type UniqueID struct {
 	Oid     uint32
 }
 
+type SortableDependency struct {
+	ClassID    uint32
+	ObjID      uint32
+	RefClassID uint32
+	RefObjID   uint32
+}
+
 // This function only returns dependencies that are referenced in the backup set
-func GetDependencies(connectionPool *dbconn.DBConn, backupSet map[UniqueID]bool) DependencyMap {
+func GetDependencies(connectionPool *dbconn.DBConn, backupSet map[UniqueID]bool, tables []Table) DependencyMap {
 	query := `SELECT
 	coalesce(id1.refclassid, d.classid) AS classid,
 	coalesce(id1.refobjid, d.objid) AS objid,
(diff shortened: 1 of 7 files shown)


end_to_end/end_to_end_suite_test.go
```diff
@@ -1856,4 +1856,59 @@ LANGUAGE plpgsql NO SQL;`)
...
+                     PARTITION May16 START (date '2016-05-01') INCLUSIVE ,
+                     PARTITION Jun16 START (date '2016-06-01') INCLUSIVE ,
+                     PARTITION Jul16 START (date '2016-07-01') INCLUSIVE ,
+                     PARTITION Aug16 START (date '2016-08-01') INCLUSIVE ,
+                     PARTITION Sep16 START (date '2016-09-01') INCLUSIVE ,
+                     PARTITION Oct16 START (date '2016-10-01') INCLUSIVE ,
+                     PARTITION Nov16 START (date '2016-11-01') INCLUSIVE ,
+                     PARTITION Dec16 START (date '2016-12-01') INCLUSIVE
+                                     END (date '2017-01-01') EXCLUSIVE);
+                   CREATE EXTERNAL TABLE testchema.external_apj (a INT,b DATE,c TEXT,d INT) LOCATION ('gpfdist://127.0.0.1/apj') format 'text';
```
khuddlefish, line 1886: Can we actually run this syntax on a 7 system successfully? I would have expected that it might not work anymore

backup/dependencies.go
```diff
@@ -170,16 +178,51 @@ JOIN pg_type t ON d.refobjid = t.oid
...
-		ObjID      uint32
-		RefClassID uint32
-		RefObjID   uint32
-	}, 0)
-
+	pgDependDeps := make([]SortableDependency, 0)
 	err := connectionPool.Select(&pgDependDeps, query)
 	gplog.FatalOnError(err)
 
+	if connectionPool.Version.AtLeast("7") && len(tables) > 0 {
```
khuddlefish, line 185: might be good to add some of the commit message info as a comment somewhere nearby

backup/dependencies.go
```diff
@@ -170,16 +178,51 @@ JOIN pg_type t ON d.refobjid = t.oid
...
+					AND dep.deptype = 'n'
+                )
+              SELECT
+                  'pg_constraint'::regclass::oid AS ClassID,
+                  con.oid AS ObjID,
+                  'pg_class'::regclass::oid AS RefClassID,
+                  foo.refobjid AS RefObjID
+              FROM
+                  pg_constraint con
+                  INNER JOIN foo
```
khuddlefish, line 213: Is this `foo` meaningful or is it programmer lingo sneaking in? If that can we change it to something meaningful?

end_to_end/end_to_end_suite_test.go
```diff
@@ -1909,6 +1910,141 @@ LANGUAGE plpgsql NO SQL;`)
 			Expect(err).ToNot(HaveOccurred())
 
 		})
+		It("Will correctly handle leaf-partition backups for filtered to child table with intermediate partitions on multiple versions of GPDB", func() {
+			testutils.SkipIfBefore7(backupConn)
```
khuddlefish, line 1914: Is this supposed to run on multiple versions like it says in the title or only for 7?

end_to_end/end_to_end_suite_test.go
AJR-VMware: The syntax is backwards-compatible, so this still works as expected on 7.  The under-the-hood mechanics for this has been rewritten though.

backup/dependencies.go
AJR-VMware: Agreed, done!

backup/dependencies.go
AJR-VMware: Just a placeholder, agreed it should be meaningful.  Done!

end_to_end/end_to_end_suite_test.go
AJR-VMware: I considered that, and would be willing to.  I was a bit concerned with the overhead of re-running the setup query each time, but it's not that much and only a few times, so maybe it's fine?

AJR-VMware: Ok, got these refactored as well.  The actual performance hit is pretty negligible imo.

khuddlefish: I'm not seeing the rework in this PR? Should I be or are you doing that somewhere else?

Also didn't get an answer to the first question. The title says "multiple versions of gpdb" but the `SkipIfBefore7` says only 7. That seems contradictory to me so just want to make sure we're clear about what versions we should be testing here.

AJR-VMware: Sorry about that!  

1. The rework is up now, with the further refactor to DescribeTable that Jamie suggested.
2. The multiple versions language was wrong, we went with 7+ syntax here, so this test is 7+ only.  The new test right before this one is 6 and 7, since we can get away with the same syntax for both versions.

khuddlefish (approved):

PR #591, opened 2022-11-29
Rebase fix gp7 logerrpersist

AJR-VMware:
Add GPDB7 support for pulling external table logging modifier.

### DIFF
--- a/backup/queries_externals.go
+++ b/backup/queries_externals.go
@@ -14,14 +14,14 @@ import (
 func GetExternalTableDefinitions(connectionPool *dbconn.DBConn) map[uint32]ExternalTableDefinition {
 	gplog.Verbose("Retrieving external table information")
 
-		// In GPDB 4.3, we need to get the error table's fully-qualified name
-		// if `LOG ERRORS INTO <err_table_name>` was used. If `LOG ERRORS` was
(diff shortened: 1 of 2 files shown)


backup/queries_externals.go
```diff
@@ -16,69 +14,94 @@ import (
...
+		e.fmtopts AS formatopts,
+		coalesce(e.command, '') AS command,
+		coalesce(e.rejectlimit, 0) AS rejectlimit,
+		coalesce(e.rejectlimittype, '') AS rejectlimittype,
+		e.fmterrtbl IS NOT NULL AND e.reloid = e.fmterrtbl AS logerrors,
+		'error_log_persistent=true' = any(e.options) AS logerrpersist,
+		pg_encoding_to_char(e.encoding) AS encoding,
+		e.writable
+	FROM pg_exttable e`
+
```
khuddlefish, line 58: Can we get a comment about why 6 is different from 5?

khuddlefish (approved):
Much easier to read with the different version queries. Only suggestion is a comment for the 6 one since all other have nice comments.

backup/queries_externals.go
AJR-VMware: Yep, just added it!

jmcatamney (approved):

PR #592, opened 2022-11-30
Fix distribution policy backup for materialized views

AJR-VMware:
In GPDB6.2 materialized views were added, but no support for their distribution policies were included. They will default to using the first column, but can have arbitrary policies.

Add support and test coverage for retrieving these policies and printing them out during backup.

### DIFF
--- a/backup/queries_relations.go
+++ b/backup/queries_relations.go
@@ -327,6 +327,7 @@ type View struct {
 	Definition     sql.NullString
 	Tablespace     string
 	IsMaterialized bool
+	DistPolicy     string
 }
 
 func (v View) GetMetadataEntry() (string, toc.MetadataEntry) {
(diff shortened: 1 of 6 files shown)


backup/queries_relations.go
```diff
@@ -411,12 +414,18 @@ func GetAllViews(connectionPool *dbconn.DBConn) []View {
...
-			} else {
-				verifiedResults = append(verifiedResults, result)
 			}
 		} else {
+			// do not append views with invalid definitions
 			gplog.Warn("View '%s.%s' not backed up, most likely dropped after gpbackup had begun.", result.Schema, result.Name)
+			continue
 		}
+
+		if result.IsMaterialized {
```
khuddlefish, line 424: So the distribution policies only apply to materialized views?

khuddlefish (approved):

backup/queries_relations.go
AJR-VMware: Yep.  Normal non-materialized views are not saved to disk, they're recomputed dynamically when queried.  So there's no need to distribute the data, and they don't get a policy.

PR #593, opened 2022-12-05
Rename exchanged partition table indexes

AJR-VMware:
In the case that a partitioned table has indexes, exchanging a leaf will result in both the original leaf and the new partition table having indexes with names matching the original table name.  This will cause restore to fail due to a name collision.  This is only an issue on GPDB6 and earlier.

To resolve this, we retrieve indexes meeting these criteria and rename them to the convention matching the exchanged leaf table's name before backing up.

This situation also happens with constraint indexes on GPDB7 and above, due to changes made to align with upstream.  An equivalent solution is applied in this case.

Note that we use two very similar functions to handle constraints and indexes.  This is necessary due to current limitations of generics in Go.

### DIFF
--- a/backup/queries_postdata.go
+++ b/backup/queries_postdata.go
@@ -8,6 +8,7 @@ package backup
 import (
 	"database/sql"
 	"fmt"
+	"strings"
 
 	"github.com/greenplum-db/gp-common-go-libs/dbconn"
 	"github.com/greenplum-db/gp-common-go-libs/gplog"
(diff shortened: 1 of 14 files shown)


backup/queries_postdata.go
```diff
@@ -240,6 +241,70 @@ func GetIndexes(connectionPool *dbconn.DBConn) []IndexDefinition {
 	return sortedIndexes
 }
 
+func GetRenameExchangedPartitionQuery(connection *dbconn.DBConn) string {
+	// In the case of exchanged partition tables, restoring index constraints with system-generated
+	// will cause a name collision in GPDB7+.  Rename those constraints to match their new owning
+	// tables.  In GPDB6 and below this renaming was done automatically by server code.
```
khuddlefish, line 247: Is this comment still accurate? It sort of sounds like we shouldn't have to do anything special for 6 and below because it was all done by server code, but it seems like we are doing something to handle those cases?

backup/queries_shared.go
```diff
@@ -236,6 +236,26 @@ func GetConstraints(connectionPool *dbconn.DBConn, includeTables ...Relation) []
 	}
 }
 
+func RenameExchangedPartitionConstraints(connectionPool *dbconn.DBConn, indexes *[]Constraint) {
```
khuddlefish, line 239: Is this what we were talking about with generics not letting us reuse the same function definition even though they were very similar?

khuddlefish, line 239: should the variable be called `constraints` instead of `indexes`?

khuddlefish (approved):

backup/queries_shared.go
AJR-VMware: Ah, yep.  Fixed!

AJR-VMware: Yeah, we sank a bunch of time into trying to have these be just one function.  Go really didn't want to play along.

backup/queries_postdata.go
AJR-VMware: Index constraints were handled automatically in 6 and below.  Other kinds of indexes were not, which we discovered in the course of this ticket.  It's a bit confusing, we had to fix a bug with indexes in 6-, and a bug with index constraints in 7+.  Super similar fixes in both cases, just handled awkwardly across GPDB versions.

PR #595, opened 2022-12-07
Add file count verification to resize restores

AJR-VMware:
In the resize-restore case, expected file counts will vary based on the sizes of the source and destination clusters.  Add logic to calculate the expected sizes, and fail if the correct number of files are not found.

### DIFF
--- a/restore/remote.go
+++ b/restore/remote.go
@@ -20,9 +20,7 @@ func VerifyBackupDirectoriesExistOnAllHosts() {
 	_, err := globalCluster.ExecuteLocalCommand(fmt.Sprintf("test -d %s", globalFPInfo.GetDirForContent(-1)))
 	gplog.FatalOnError(err, "Backup directory %s missing or inaccessible", globalFPInfo.GetDirForContent(-1))
 	if MustGetFlagString(options.PLUGIN_CONFIG) == "" || backupConfig.SingleDataFile {
-		isResizeRestore := MustGetFlagBool(options.RESIZE_CLUSTER)
-		origSize := len(globalCluster.ContentIDs) - 1
-		destSize := backupConfig.SegmentCount
+		origSize, destSize, isResizeRestore := GetResizeClusterInfo()
 
 		remoteOutput := globalCluster.GenerateAndExecuteCommand("Verifying backup directories exist", cluster.ON_SEGMENTS, func(contentID int) string {
 			if isResizeRestore { // Map origin content to destination content to find where the original files have been placed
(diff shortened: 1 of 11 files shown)


AJR-VMware:
@jmcatamney This is a bit different than what we reviewed.  I had to remove the "default sizes to 0" logic because there wasn't a clean way to do the logic for "not allowed to resize-cluster without the flag" verification.  Now we always carry around the true size values and conditionally set origSize to 0 in non-resize restores for the sake of communicating with the helper.

restore/remote.go
```diff
@@ -39,17 +37,48 @@ func VerifyBackupDirectoriesExistOnAllHosts() {
...
+		fileCount = len(globalTOC.DataEntries)
+	}
+
+	origSize, destSize, isResizeRestore := GetResizeClusterInfo()
+	batchMap := make(map[int]int, len(remoteOutput.Commands))
+	if isResizeRestore {
+		for i := 0 + 1; i < destSize; i++ {
+			// initialize expected file counts to zero so they can later be more easily incremented
+			batchMap[i] = 0
+		}
```
jmcatamney, line 60: This loop isn't necessary, as Go maps default to the zero value and incrementing an uninitialized value will have the desired behavior.

restore/remote.go
```diff
@@ -39,17 +37,48 @@ func VerifyBackupDirectoriesExistOnAllHosts() {
...
+			for i := 0; i < origSize; i++ {
+				batchMap[i] = fileCount
+			}
+		} else {
+			// Larger to smaller resize
+			for i := 0; i < origSize; i++ {
+				targetID := i % destSize
+				batchMap[targetID] += fileCount
+			}
+		}
```
jmcatamney, line 73: This can be simplified to
```
for i := 0; i < origSize; i++ {
	batchMap[i%destSize] += fileCount
}
```
as in the smaller-to-larger case `i % destSize` is `i`.

jmcatamney: Actually, the logic is the same in the non-resize scenario, too, so we don't even need to special-case that.  We can just fill `batchMap` in all cases, and use it without an `isResizeRestore` check on line 79.

restore/restore.go
```diff
@@ -595,8 +588,11 @@ func DoTeardown() {
 			return
 		}
 		reportFilename := globalFPInfo.GetRestoreReportFilePath(restoreStartTime)
-		origSize := backupConfig.SegmentCount
-		destSize := len(globalCluster.Segments) - 1
+		origSize, destSize, isResizeRestore := GetResizeClusterInfo()
+		if !isResizeRestore {
+			// we pass origSize to helpers, and they use >0 to gate resize-restore behavior
```
jmcatamney, line 593: As long as we're refactoring this anyway, it might make sense to explicitly pass an `--is-resize-restore` flag to the helper instead of relying on unsetting a variable here.

AJR-VMware: That makes sense.  I'll go make that change.

restore/remote.go
AJR-VMware: I can squeeze it down to just the one loop, good call!

I think we should still gate using the `batchMap` on `isResizeRestore`, though, because in older backups origSize will be set to 0, so we're going to need to gate its use either way?

restore/remote.go
AJR-VMware: Ah, I didn't know that.  Neat, thanks!

jmcatamney (approved):

PR #602, opened 2023-01-02
Refactor query strings for clarity

AJR-VMware:
Adding multiple levels of interpolation for version support has reduced readability of some queries.  Clean up a few of them where possible.

### DIFF
--- a/backup/queries_relations.go
+++ b/backup/queries_relations.go
@@ -202,9 +202,7 @@ type SequenceDefinition struct {
 }
 
 func GetAllSequences(connectionPool *dbconn.DBConn) []Sequence {
-	var query string
-	if connectionPool.Version.AtLeast("7") {
-		query = fmt.Sprintf(`
+	atLeast7Query := fmt.Sprintf(`
 		SELECT n.oid AS schemaoid,
 			c.oid AS oid,
 			quote_ident(n.nspname) AS schema,
@@ -229,9 +227,9 @@ func GetAllSequences(connectionPool *dbconn.DBConn) []Sequence {
 			AND %s
 			AND %s
 		ORDER BY n.nspname, c.relname`,
-			relationAndSchemaFilterClause(), ExtensionFilterClause("c"))
-	} else {
-		query = fmt.Sprintf(`
+		relationAndSchemaFilterClause(), ExtensionFilterClause("c"))
+
+	before7Query := fmt.Sprintf(`
 		SELECT n.oid AS schemaoid,
 			c.oid AS oid,
 			quote_ident(n.nspname) AS schema,
@@ -250,8 +248,15 @@ func GetAllSequences(connectionPool *dbconn.DBConn) []Sequence {
 			AND %s
 			AND %s
 		ORDER BY n.nspname, c.relname`,
-			relationAndSchemaFilterClause(), ExtensionFilterClause("c"))
+		relationAndSchemaFilterClause(), ExtensionFilterClause("c"))
+
+	query := ""
+	if connectionPool.Version.Before("7") {
+		query = before7Query
+	} else {
+		query = atLeast7Query
 	}
+
 	results := make([]Sequence, 0)
 	err := connectionPool.Select(&results, query)
 	gplog.FatalOnError(err)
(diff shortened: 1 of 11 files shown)


backup/queries_relations.go
```diff
@@ -371,31 +370,44 @@ func GetAllViews(connectionPool *dbconn.DBConn) []View {
...
+		pg_get_viewdef(c.oid) AS definition
+	FROM pg_class c
+		LEFT JOIN pg_namespace n ON n.oid = c.relnamespace
+		LEFT JOIN pg_tablespace t ON t.oid = c.reltablespace
+	WHERE c.relkind IN ('m', 'v')
+		AND %s
+		AND %s`, relationAndSchemaFilterClause(), ExtensionFilterClause("c"))
+
+	// Materialized views were introduced in GPDB 7 and backported to GPDB 6.2.
+	version6Query := fmt.Sprintf(`
```
khuddlefish, line 387: What is our convention for naming queries that apply to multiple versions? Since this is version6+ not just version6.

backup/queries_relations.go
```diff
@@ -371,31 +370,44 @@ func GetAllViews(connectionPool *dbconn.DBConn) []View {
...
+	SELECT
+		c.oid AS oid,
+		quote_ident(n.nspname) AS schema,
+		quote_ident(c.relname) AS name,
+		pg_get_viewdef(c.oid) AS definition,
 		coalesce(' WITH (' || array_to_string(c.reloptions, ', ') || ')', '') AS options,
```
khuddlefish, line 393: Are these coalesce statements also related to materialized views or something else? If they're not maybe could add a note about where they came from too.

backup/queries_table_defs.go
```diff
@@ -250,54 +251,40 @@ func GetColumnDefinitions(connectionPool *dbconn.DBConn) map[uint32][]ColumnDefi
...
-			aclCols = `CASE
-				WHEN a.attacl IS NULL THEN NULL
-				WHEN array_upper(a.attacl, 1) = 0 THEN a.attacl[0]
-				ELSE unnest(a.attacl) END`
-		}
+	ORDER BY a.attrelid, a.attnum`, relationAndSchemaFilterClause())
 
-		selectClause += fmt.Sprintf(`,
-		%s AS privileges,
+	version6Query := fmt.Sprintf(`
```
khuddlefish, line 272: Is there any note we can add to explain the changes between 5 and 6?

khuddlefish (commented):
I'm pretty sure the queries look correct. I assume you just followed the conditionals to build what the code ended up constructing. Could potentially use some more comments to explain the difference between versions, especially for gp6.

backup/queries_relations.go
AJR-VMware: We don't have a rigid convention as such.  It seems like in a lot of places we just leave the "current" version of the query with a base name.  In other places we were calling it "masterQuery", though that's been autoreplaced to "coordinatorQuery" which may be misleading given it doesn't necessarily run on the coordinator and the name was instead referring to the branch it was on.

Given that the version gate for picking it uses pretty clear logic I wasn't too worried about the name in this case, but I'd be open to a convention if we wanted to pick one.

backup/queries_relations.go
AJR-VMware: They are not, it's just that reloptions and tablespace info got added to pg_class in upstream versions between 5 and 6.  I'll add another comment about it.

backup/queries_table_defs.go
AJR-VMware: Yep, it's basically "6 added a bunch of new column info we need to capture".  I added a comment!

backup/queries_relations.go
khuddlefish: I looked at some examples. One thing I see a little that's kind of neat is `before6query` so that could be paired with `atleast6query` to mirror the language we use to test it. Otherwise, I would probably call it `mainQuery`. I think that shows up several places. `version6query` to me means that it is only for version6 so I don't like it much.

khuddlefish:
You mentioned this too, but I see several places where `masterQuery` was changed to `coordinatorQuery` instead of `mainQuery` by mistake (since it was always referring to the gpdb git branch, not the master node).

backup/queries_relations.go
jmcatamney: I think the `beforeNquery`/`atleastNquery` convention is a good idea.  A quick search doesn't find any existing instances of `mainQuery`, and regardless that one kind of implies (at least to me) "latest version only" in the same way that `version6query` implies "version 6 only."

AJR-VMware: Ok.  I'll do a full pass today and go with `beforeNQuery` and `atLastNQuery` across the board.  Thanks team.

khuddlefish:
> You mentioned this too, but I see several places where `masterQuery` was changed to `coordinatorQuery` instead of `mainQuery` by mistake (since it was always referring to the gpdb git branch, not the master node).

@AJR-VMware Are you going to address this in a separate PR?

khuddlefish:
> > > You mentioned this too, but I see several places where `masterQuery` was changed to `coordinatorQuery` instead of `mainQuery` by mistake (since it was always referring to the gpdb git branch, not the master node).
> > 
> > 
> > @AJR-VMware Are you going to address this in a separate PR?
> 
> @khuddlefish no, I changed them all in this PR. Did I miss one?

Oh ha. No, I just read it wrong. Looks good 👍

PR #610, opened 2023-02-08
Allow resize tests to work on multi-node clusters

jmcatamney:
Currently all end-to-end tests assume a local demo cluster and don't take advantage of any multi-node CI setups.  This commit adds a function to all of the resize restore tests that will take saved backups and copy the data directories over to segment hosts in order to test them in a multi-node context.

A few other changes are made to support this:
- The DataDomain test images don't like it when tar tries to extract the non-writable files that gpbackup creates in the coordinator data directory, so the existing saved backups are modified to make those files writable and a helper function is added to help with adding new saved backups in future.
- The lack of a 2-segment incremental rowcount file was obscured by the fact that all of these tests were pended before, so that file has been added in this commit.
- Various helper functions are modified to accept a hostname, now that they're dealing with segment hosts.


AJR-VMware:
@jmcatamney What were the changes to the zipped backups?  I see +/- N bytes on several of them, was just curious.

AJR-VMware (approved):

jmcatamney:
@AJR-VMware The change was making the metadata files writable for each of them to avoid the DDBoost tests complaining about trying to untar unwritable files; we didn't notice the issue before because the tests were all pended.

PR #611, opened 2023-02-15
Add sqlite

AJR-VMware:
Implement initialization, storage, and retrieval for backup history in a sqlite database.
Convert all existing usage of backup history from legacy yaml file.
Change or add unit tests to cover this new functionality.

Note: keep the function for locating the legacy yaml file around for use by gp-backup-manager in
migrating pre-existing files into new database.


jmcatamney (approved):

cobolbaby:
Why use sqlite instead of yaml file?

AJR-VMware:
> Why use sqlite instead of yaml file?

A couple reasons.
1. The yaml file has become borderline unusable at scale.  Installs have reached 20+GB on disk, which makes unmarshalling and altering it take longer than the backup itself in many cases.
2. Several requested features will be far more robust and reliable, as well as just plain easier to implement, with a database backing the history instead of yaml.

cobolbaby:
> A couple reasons.
> 
> 1. The yaml file has become borderline unusable at scale.  Installs have reached 20+GB on disk, which makes unmarshalling and altering it take longer than the backup itself in many cases.
> 2. Several requested features will be far more robust and reliable, as well as just plain easier to implement, with a database backing the history instead of yaml.

Have you considered persisting it into the gpperfmon database?

AJR-VMware:
> > A couple reasons.
> > 
> > 1. The yaml file has become borderline unusable at scale.  Installs have reached 20+GB on disk, which makes unmarshalling and altering it take longer than the backup itself in many cases.
> > 2. Several requested features will be far more robust and reliable, as well as just plain easier to implement, with a database backing the history instead of yaml.
> 
> Have you considered persisting it into the gpperfmon database?

We rejected hosting the history file within the cluster because this makes it impossible to list available backups if the cluster is down, which is a common use case.

PR #612, opened 2023-02-17
Add additional plugin tests to improve coverage

jmcatamney:
This commit expands the basic plugin tests to cover additional flags that might potentially perform differently when restoring with a plugin, and adds end-to-end tests for plugin-based resize restores.

### DIFF
--- a/plugins/plugin_test.sh
+++ b/plugins/plugin_test.sh
@@ -400,17 +400,22 @@ set -e
 #gpbackup --dbname $test_db --plugin-config $plugin_config $further_options > $log_file
 
 test_backup_and_restore_with_plugin() {
-    flags=$1
-    restore_filter=$2
+    flags_backup=$1
+    flags_restore=$2
+    restore_filter=$3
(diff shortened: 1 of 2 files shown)


plugins/plugin_test.sh
```diff
@@ -400,17 +400,22 @@ set -e
...
-    restore_filter=$2
+    flags_backup=$1
+    flags_restore=$2
+    restore_filter=$3
     test_db=plugin_test_db
     log_file="$logdir/plugin_test_log_file"
 
     psql -X -d postgres -qc "DROP DATABASE IF EXISTS $test_db" 2>/dev/null
-    createdb $test_db
+    psql -X -d postgres -qc "CREATE DATABASE $test_db" 2>/dev/null
```
AJR-VMware, line 410: Is there a reason createdb isn't working?  Seems like we'd want to keep our configuration compatible with our other utils if possible.

AJR-VMware (approved):
Approved with some minor nits.  LGTM!

PR #618, opened 2023-03-08
Move CI artifacts to GCS

AJR-VMware:
All possible intermediate and RC artifacts from gpbackup's CI moved from AWS into GCS.


ci/gpbackup-release-generated.yml
```diff
@@ -454,84 +454,76 @@ resources:
...
+    json_key: ((dp/dev/gcp_svc_acct_key))
+    versioned_file: gpbackup/release-candidates/gpbackup-gppkgs.tar.gz
+
+- name: gppkgs_release_rc
+  type: gcs
+  icon: google
+  source:
+    bucket: ((dp/prod/gcs-ci-bucket))
+    json_key: ((dp/prod/gcp_svc_acct_key))
+    regexp: gpbackup/published/gpbackup_tools-(.*).gppkg
```
khuddlefish, line 486: Why is this called `gpbackup_tools`? The file we upload to tanzunet is called `pivotal_greenplum_backup_restore-<version>-gp5-rhel-x86_64.gppkg` so I would expect this to be named the same unless this file is something different.

AJR-VMware: Our update-metadata.bash script renames things to that pattern before loading to tanzunet (see below).  If you'd like I can have us upload the renamed files instead, no problem.

```pushd workspace/files-to-upload
  # rename files to match the name inside metadata.yml
  for filename in *.gppkg; do
    newFilename=$(sed -e "s/gpbackup_tools-[+0-9a-z.]*/pivotal_greenplum_backup_restore-${TILE_RELEASE_VERSION}/" -e "s/RHEL/rhel/" -e "s/SLES/sles/" <<< "$filename")
    if ! [ "$filename" == "$newFilename" ]; then
      mv "$filename" "$newFilename"
    fi
  done
  # rename binary tarball
  mv "bin_gpbackup.tar.gz" "pivotal_greenplum_backup_restore-${TILE_RELEASE_VERSION}.tar.gz"
popd
```

AJR-VMware: Ok, after digging through this some more it looks like we remove the extra version files and such anyway, so matching the upload to exactly what pivnet-upload has will probably accomplish what we want, I think.

ci/gpbackup-release-generated.yml
```diff
@@ -1308,65 +1439,156 @@ jobs:
...
+        file: github_release_components_rhel8/gpbackup-*.tar.gz
+    - put: github_release_components_rc_ubuntu
+      params:
+        file: github_release_components_ubuntu/gpbackup-*.tar.gz
 
 - name: push-to-github
   plan:
   - get: rocky8-gpdb7-image
-  - get: github_release_components_rc
+  - get: github_release_components_rc_rhel6
```
khuddlefish, line 1461: There is an `unpack` parameter on this resource type you might find useful
https://github.com/frodenas/gcs-resource

ci/gpbackup-release-generated.yml
```diff
@@ -1308,65 +1439,156 @@ jobs:
...
       platform: linux
       inputs:
-        - name: github_release_components_rc
+        - name: github_release_components_rc_rhel6
+        - name: github_release_components_rc_rhel7
+        - name: github_release_components_rc_rhel8
+        - name: github_release_components_rc_ubuntu
       outputs:
-        - name: github_release_components_untarred
+        - name: github_release_components_untarred_rhel6
```
khuddlefish, line 1479: could we use a shorter name for this so the commands below are easier to read? like `components_untarred_rhel6`?

ci/pivnet_release/metadata.yml
```diff
@@ -17,60 +17,89 @@ release:
...
+      description:
+      upload_as: Backup and Restore for GP 6 on RHEL7 compressed tarball
+      file_type: Software
+      file_version: <TILE_RELEASE_VERSION>
+    - file: file://pivotal_greenplum_backup_restore-<TILE_RELEASE_VERSION>-rhel6.tar.gz
+      description:
+      upload_as: Backup and Restore for GP 6 on RHEL6 compressed tarball
+      file_type: Software
+      file_version: <TILE_RELEASE_VERSION>
+    - file: file://pivotal_greenplum_backup_restore-<TILE_RELEASE_VERSION>-gp6-ubuntu-amd64.gppkg
```
khuddlefish, line 66: Can we put this gppkg by the others? So it goes rhel8, rhel7, rhel6, ubuntu. gppkgs then tarballs

ci/gpbackup-release-generated.yml
```diff
@@ -453,85 +453,205 @@ resources:
     secret_access_key: ((aws-bucket-secret-access-key))
     regexp: pivnet_release_version/v-(.*)
 
-- name: github_release_components
-  type: s3
-  icon: amazon
+- name: github_release_components_rhel6
```
khuddlefish, line 456: Can we drop `github` from these release names? I know that's what it used to be but now that there's more of them it's very visible. Makes we think these are the packages for the github release but they're not.

khuddlefish (commented):
This is not bad though I do have a feeling that we're doing more work with passing around differently named resources than we have to. We probably don't have to do all the changes now to make it optimal.

Here's a thought I had that I'm wondering if it might help make things simpler.

Job 1. build go binaries and ddboost. Structure things together into the tarball form needed for tanzunet. End by creating the tarballs for tanzunet and you can name them their final form for that

Job 2. build gppkgs. Take the tanzunet tarballs and build gppkgs from them. They should have all the files and libraries you need inside them already. Name them with the final form

Job 3. Take the tanzunet tarballs and gppkgs you built and upload them to tanzunet

Job 4. Take the tanzunet tarballs and untar them. Select out the files you want for github and repackage. Upload to github.

Thoughts on something like that?

ci/scripts/update-metadata.bash
```diff
@@ -86,10 +110,29 @@ pushd workspace/files-to-upload
       mv "$filename" "$newFilename"
     fi
   done
-  # rename binary tarball
-  mv "bin_gpbackup.tar.gz" "pivotal_greenplum_backup_restore-${TILE_RELEASE_VERSION}.tar.gz"
+
+  # rename binary tarballs, to match the name inside metadata.yml metadta.yml
```
khuddlefish, line 114: metadta.yml typo?

khuddlefish (approved):

ci/scripts/update-metadata.bash
AJR-VMware: Yep, fixed!

PR #622, opened 2023-03-21
Update all dependencies for Go 1.19

jmcatamney:
The dependencies for this project were fairly out-of-date, not because it relied on anything specific in older versions of any dependencies but simply because they hadn't been updated in a while.

As part of a general effort to update GPDB Go projects to 1.19, this commit updates the required Go version and updates all dependencies to their currently-latest versions.

Additionally, now that gpbackup is using the same version of Go as the CI images, installation of Go during test setup is no longer required.


ci/templates/gpbackup-tpl.yml
```diff
@@ -1347,7 +1341,7 @@ jobs:
 - name: GPDB6
   plan:
   - in_parallel:
-    - get: centos7-gpdb6-image
+    - get: centos7-gpdb6-golang-image
```
AJR-VMware, line 1344: Need to regenerate the `*generated*.yml` pipeline files off the new template and add to the PR.  This may also need a manual rebase given  all the changes to this file, not sure.

AJR-VMware (approved):
LGTM!

PR #624, opened 2023-03-24
Update resource group handling for 7, and fix some other broken tests.

AJR-VMware:
- Resource groups had several breaking changes made for GPDB7.  Using either genericized or separate functions, support both the legacy and new structures.
- View definition for anonymous columns has been changed in GPDB7 to show a ?column? name.  Update our tests to match that behavior.
- Changes to the spacing and storage defaults in GPDB7 beta broke some of our more rigid tests. Update them so they pass again.
- Update Makefile to use gpsync if present.  Fall back to gpscp if gpsync isn't present, for backwards-compatibility

### DIFF
--- a/backup/metadata_globals_test.go
+++ b/backup/metadata_globals_test.go
@@ -141,21 +141,21 @@ GRANT TEMPORARY,CONNECT ON DATABASE testdb TO testrole;`,
 		var emptyResGroupMetadata = backup.MetadataMap{}
 		It("prints resource groups", func() {
 			testhelper.SetDBVersion(connectionPool, "5.9.0")
-			someGroup := backup.ResourceGroup{Oid: 1, Name: "some_group", CPURateLimit: "10", MemoryLimit: "20", Concurrency: "15", MemorySharedQuota: "25", MemorySpillRatio: "30"}
-			someGroup2 := backup.ResourceGroup{Oid: 2, Name: "some_group2", CPURateLimit: "20", MemoryLimit: "30", Concurrency: "25", MemorySharedQuota: "35", MemorySpillRatio: "10"}
-			resGroups := []backup.ResourceGroup{someGroup, someGroup2}
+			someGroup := backup.ResourceGroupBefore7{ResourceGroup: backup.ResourceGroup{Oid: 1, Name: "some_group", Concurrency: "15"}, CPURateLimit: "10", MemoryLimit: "20", MemorySharedQuota: "25", MemorySpillRatio: "30"}
+			someGroup2 := backup.ResourceGroupBefore7{ResourceGroup: backup.ResourceGroup{Oid: 2, Name: "some_group2", Concurrency: "25"}, CPURateLimit: "20", MemoryLimit: "30", MemorySharedQuota: "35", MemorySpillRatio: "10"}
+			resGroups := []backup.ResourceGroupBefore7{someGroup, someGroup2}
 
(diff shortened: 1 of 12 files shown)


backup/metadata_globals_test.go
```diff
@@ -248,11 +248,22 @@ GRANT TEMPORARY,CONNECT ON DATABASE testdb TO testrole;`,
...
-				`ALTER RESOURCE GROUP default_group SET CPU_RATE_LIMIT 1;`,
-				`ALTER RESOURCE GROUP default_group SET MEMORY_LIMIT 1;`)
+			if connectionPool.Version.Before("7") {
+				testutils.AssertBufferContents(tocfile.GlobalEntries, buffer,
+					`ALTER RESOURCE GROUP admin_group SET CPU_RATE_LIMIT 1;`,
+					`ALTER RESOURCE GROUP admin_group SET MEMORY_LIMIT 1;`,
+					`ALTER RESOURCE GROUP default_group SET CPU_RATE_LIMIT 1;`,
+					`ALTER RESOURCE GROUP default_group SET MEMORY_LIMIT 1;`)
+			} else { // GPDB7+
+				testutils.AssertBufferContents(tocfile.GlobalEntries, buffer,
```
AJR-VMware, line 258: @kainwen Would it be possible for someone with context on the new resource groups to assess whether: 

- system_group should be handled the same as the old default admin_group and default_group values
- Setting these values as defaults is appropriate. these may stay as-is depending on how the backup was taken, and will at minimum be set until downstream restore changes them.
- There are any other options that should be set in a similar way for these default groups.

Thank you!

backup/queries_globals.go
```diff
@@ -214,71 +209,93 @@ func (rg ResourceGroup) FQN() string {
...
+	    	JOIN pg_resgroupcapability t3 ON t3.resgroupid = g.oid AND t3.reslimittype = 3
+	    	JOIN pg_resgroupcapability t4 ON t4.resgroupid = g.oid AND t4.reslimittype = 4
+	    	JOIN pg_resgroupcapability t5 ON t5.resgroupid = g.oid AND t5.reslimittype = 5`
+
+		// The reslimittype 6 (memoryauditor) was introduced in GPDB
+		// 5.8.0. Default the value to '0' (vmtracker) since there could
+		// be a resource group created before 5.8.0 which will not have
+		// this memoryauditor field defined.
+		if connectionPool.Version.AtLeast("5.8.0") {
+			before7SelecClause += `, coalesce(t6.value, '0') AS memoryauditor`
```
jmcatamney, line 266: `Selec` -> `Select`, here and on line 275

backup/queries_globals.go
```diff
@@ -214,71 +209,93 @@ func (rg ResourceGroup) FQN() string {
...
+			before7SelecClause += `, coalesce(t7.value, '-1') AS cpuset`
+			before7FromClause += ` LEFT JOIN pg_resgroupcapability t7 ON t7.resgroupid = g.oid AND t7.reslimittype = 7`
+		}
+		query = fmt.Sprintf(`%s %s;`, before7SelecClause, before7FromClause)
+	} else { // GPDB7+
+		// Resource groups were heavily reworked for GPDB7
+		// See: https://github.com/greenplum-db/gpdb/commit/483adea86b50c1759460a6265b3d8e3f4198d92e
+		query = `
+			SELECT
+				G.oid       AS oid,
```
jmcatamney, line 284: Nitpick: can we lowercase these names (`g.oid`, `t1.value`, etc.) so they match the clauses above?

jmcatamney (approved):

PR #627, opened 2023-03-28
Remove foreign key constraint on restore_plan_timestamp

AJR-VMware:
While restore_plan_timestamp should technically always reference the timestamp of a prior backup, in many circumstances the backup history database may not be complete. Enforcing this relationship will cause an otherwise valid backup to fail, and it is better to allow it to succeed.  So remove the constraint.

Also add a TODO to consider logging a warning when such a backup is taken.

### DIFF
--- a/history/history.go
+++ b/history/history.go
@@ -155,13 +155,15 @@ func InitializeHistoryDatabase(historyDBPath string) (*sql.DB, error) {
 		}
 	}
 
+	// TODO -- consider warning if restore_plan_timestamp references a backup timestamp not present
+	// in historyDB? This scenario may be caused by lack of migration of legacy files, and may cause
+	// future backup-manager functionality such as CASCADE or cleanup to perform in unexpected ways.
 	createRestorePlansTable := `
 		CREATE TABLE IF NOT EXISTS restore_plans (
 			timestamp TEXT NOT NULL,
 			restore_plan_timestamp TEXT NOT NULL,
 			table_fqn TEXT NOT NULL,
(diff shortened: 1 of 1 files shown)


history/history.go
```diff
@@ -155,13 +155,15 @@ func InitializeHistoryDatabase(historyDBPath string) (*sql.DB, error) {
...
+	// in historyDB? This scenario may be caused by lack of migration of legacy files, and may cause
+	// future backup-manager functionality such as CASCADE or cleanup to perform in unexpected ways.
 	createRestorePlansTable := `
 		CREATE TABLE IF NOT EXISTS restore_plans (
 			timestamp TEXT NOT NULL,
 			restore_plan_timestamp TEXT NOT NULL,
 			table_fqn TEXT NOT NULL,
-			FOREIGN KEY(timestamp) REFERENCES backups(timestamp),
-			FOREIGN KEY(restore_plan_timestamp) REFERENCES backups(timestamp)
+			FOREIGN KEY(timestamp) REFERENCES backups(timestamp)
```
shivzone, line 166: Don't you have the same concerns with the timestamp field referencing backup timestamp ?

AJR-VMware: No, that field is generated by gpbackup as the main timestamp/key and applied during storage.  Honestly foreign keys here are kind of excessive protection since there shouldn't really be any way for those to fall out of sync, but are here as protection in case they get mangled in some unforeseen way.

jmcatamney (approved):

PR #631, opened 2023-03-31
Refactor restore plan historydb schema

AJR-VMware:
Update the historydb schema to support zero-table restore plans in a backup config, in the case of incremental backups with only heap tables, or other such edge cases.

### DIFF
--- a/end_to_end/end_to_end_suite_test.go
+++ b/end_to_end/end_to_end_suite_test.go
@@ -328,16 +328,10 @@ func getMetdataFileContents(backupDir string, timestamp string, fileSuffix strin
 }
 
 func saveHistory(myCluster *cluster.Cluster) {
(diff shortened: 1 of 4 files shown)


jmcatamney:
Could you add some tests for restore plans that look like all-AO and all-heap backups, just so we know those cases are covered?

AJR-VMware:
> Could you add some tests for restore plans that look like all-AO and all-heap backups, just so we know those cases are covered?

I have some in thee delete-* end to end tests on the manager side, but it'd be good to have some coverage in the main repo too, for sure.  Will do!

AJR-VMware:
> Could you add some tests for restore plans that look like all-AO and all-heap backups, just so we know those cases are covered?

@jmcatamney Ok, got coverage added

jmcatamney (approved):

PR #635, opened 2023-04-04
History opt logs

AJR-VMware:
Add the no-history flag to bypass writing a backup configuration entry to history db.
Add a nag-log to for users to run migrate-history if the legacy file is still around and hasn't been renamed by manager.

### DIFF
--- a/backup/backup.go
+++ b/backup/backup.go
@@ -395,14 +395,17 @@ func DoTeardown() {
 			backupReport.ConstructBackupParamsString()
 			backupReport.BackupConfig.SegmentCount = len(globalCluster.ContentIDs) - 1
 
-			historyDB, err := history.InitializeHistoryDatabase(historyDBName)
-			if err != nil {
-				gplog.Error(fmt.Sprintf("%v", err))
-			} else {
-				err = history.StoreBackupHistory(historyDB, &backupReport.BackupConfig)
-				historyDB.Close()
+			var err error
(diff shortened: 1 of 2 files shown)


backup/backup.go
```diff
@@ -383,26 +383,35 @@ func DoTeardown() {
...
+		historyFileLegacyName := globalFPInfo.GetBackupHistoryFilePath()
 		reportFilename := globalFPInfo.GetBackupReportFilePath()
 		configFilename := globalFPInfo.GetConfigFilePath()
 
 		time.Sleep(time.Second) // We sleep for 1 second to ensure multiple backups do not start within the same second.
 
+		// Check if legacy history file is still present, log warning if so. Only log if we're planning to use history db.
+		var err error
+		if _, err = os.Stat(historyFileLegacyName); err == nil && !MustGetFlagBool(options.NO_HISTORY) {
+			gplog.Warn("Legacy gpbackup_history file %s is still present. Please run 'gpbackup_manager --migrate-history' to ensure your backup history database is complete.", historyFileLegacyName)
```
jmcatamney, line 395: I don't know if `to ensure your backup history database is complete` is the best verbiage for this; "complete" can imply that they might miss some history somehow if they don't do the migration, and it isn't accurate if they have multiple history files in need of migration.

Something like `to add entries from that file to the history database` or `so that legacy history information will be accessible from the history database` would probably be better.

jmcatamney (approved):

PR #640, opened 2023-04-11
Modify filepath to be able to work with mirrors

jmcatamney:
While gpbackup doesn't deal with mirrors, gpbackup_manager uses the filepath module and it does need to interact with mirrors, so this commit modifies the filepath constructor to be able to either retrieve information for the primaries (and coordinator) or the mirrors (and standby), defaulting to the primaries to preserve current behavior.

### DIFF
--- a/filepath/filepath.go
+++ b/filepath/filepath.go
@@ -28,16 +28,24 @@ type FilePathInfo struct {
 	UserSpecifiedSegPrefix string
 }
 
-func NewFilePathInfo(c *cluster.Cluster, userSpecifiedBackupDir string, timestamp string, userSegPrefix string) FilePathInfo {
+func NewFilePathInfo(c *cluster.Cluster, userSpecifiedBackupDir string, timestamp string, userSegPrefix string, useMirrors ...bool) FilePathInfo {
 	backupFPInfo := FilePathInfo{}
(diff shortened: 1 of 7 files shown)


filepath/filepath.go
```diff
@@ -28,16 +28,22 @@ type FilePathInfo struct {
...
 	backupFPInfo.PID = os.Getpid()
 	backupFPInfo.UserSpecifiedBackupDir = userSpecifiedBackupDir
 	backupFPInfo.UserSpecifiedSegPrefix = userSegPrefix
 	backupFPInfo.Timestamp = timestamp
 	backupFPInfo.SegDirMap = make(map[int]string)
-	for _, segment := range c.Segments {
-		backupFPInfo.SegDirMap[segment.ContentID] = segment.DataDir
+
+	role := "p"
+	if len(useMirrors) == 1 && useMirrors[0] {
```
AJR-VMware, line 40: Can we add a comment that the current purpose for this is primarily backup manger deletes?  Having this sitting in here but not used within this repo may cause some confusion down the road.

AJR-VMware:
Re-reviewed after common-go-libs changes with new testExecutor, still looks good to me!

PR #655, opened 2023-05-09
Ensure gpbackup_history is always written correctly

AJR-VMware:
* cd41fa30e7bb239aefdfc56bb099a4dc8fe5df4e
    * Restructure writes to gpbackup_history so that the record is written out very early with an "In Progress" status, then updated to "Success" or "Failure" as part of DoCleanup
* 383554d13eaef305e873d9a873636184fac45fc7
    * Our DoCleanup relies on golang's `recover`, which will not fire if a panic happens inside a goroutine.  Identify 4 goroutines that can panic, and add individual `recover` handling to each of them.

### DIFF
--- a/backup/backup.go
+++ b/backup/backup.go
@@ -141,6 +141,38 @@ func DoBackup() {
 	gplog.Info("Metadata will be written to %s", metadataFilename)
 	metadataFile := utils.NewFileWithByteCountFromFile(metadataFilename)
 
+	/*
+	 * We check this in the backup report rather than the flag because we
+	 * perform a metadata only backup if the database contains no tables
+	 * or only external tables
+	 */
+	backupSetTables := dataTables
+	if !backupReport.MetadataOnly {
+		targetBackupRestorePlan := make([]history.RestorePlanEntry, 0)
+		if targetBackupTimestamp != "" {
+			gplog.Info("Basing incremental backup off of backup with timestamp = %s", targetBackupTimestamp)
+
+			targetBackupTOC := toc.NewTOC(targetBackupFPInfo.GetTOCFilePath())
+			targetBackupRestorePlan = history.ReadConfigFile(targetBackupFPInfo.GetConfigFilePath()).RestorePlan
+			backupSetTables = FilterTablesForIncremental(targetBackupTOC, globalTOC, dataTables)
+		}
+
+		backupReport.RestorePlan = PopulateRestorePlan(backupSetTables, targetBackupRestorePlan, dataTables)
+	}
+
+	// As soon as all necessary data is available, capture the backup into history database
+	if !MustGetFlagBool(options.NO_HISTORY) {
+		historyDBName := globalFPInfo.GetBackupHistoryDatabasePath()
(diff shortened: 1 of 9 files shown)


backup/backup.go
```diff
@@ -473,6 +460,31 @@ func DoCleanup(backupFailed bool) {
 			}
 			utils.CleanUpHelperFilesOnAllHosts(globalCluster, globalFPInfo)
 		}
+
+		// The gpbackup_history entry is written to the DB with an "In Progress" status very early
+		// on.  If we get to cleanup and the backup succeeded, mark it as a success, otherwise mark
+		// it as a failure. Between our signal handler and recovering panics, there should be no
+		// way for gpbackup to exit that does leaves the entry in the initial status.
```
jmcatamney, line 467: Nitpick: `does leaves` -> `leaves`

backup/data.go
```diff
@@ -165,11 +165,17 @@ func backupDataForAllTables(tables []Table) []map[uint32]int64 {
...
 	 */
+	panicChan := make(chan error)
 	rowsCopiedMaps[0] = make(map[uint32]int64)
 	for connNum := 1; connNum < connectionPool.NumConns; connNum++ {
 		rowsCopiedMaps[connNum] = make(map[uint32]int64)
 		workerPool.Add(1)
 		go func(whichConn int) {
+			defer func() {
+				if panicErr := recover(); panicErr != nil {
+					panicChan <- fmt.Errorf("Recovered from worker process panic: %v", panicErr)
```
jmcatamney, line 176: I'm not sure if we want to explicitly mention recovering from panics in these error messages.  Normal `Fatal` messages don't mention panicking, so doing that here might imply it's somehow worse than other critical errors and scare users.

I think just passing the error on without annotation should be fine, as the stack trace will already tell us where the error comes from; if we want some kind of extra message, I'd go with `Worker %d could not [blah]:` or similar

AJR-VMware: You're right, transparently passing the error is fine.  Only devs need to care about that stack, and the stack trace gives us that.

backup/data.go
```diff
@@ -165,11 +165,17 @@ func backupDataForAllTables(tables []Table) []map[uint32]int64 {
...
 	 */
+	panicChan := make(chan error)
 	rowsCopiedMaps[0] = make(map[uint32]int64)
 	for connNum := 1; connNum < connectionPool.NumConns; connNum++ {
 		rowsCopiedMaps[connNum] = make(map[uint32]int64)
 		workerPool.Add(1)
 		go func(whichConn int) {
+			defer func() {
+				if panicErr := recover(); panicErr != nil {
+					panicChan <- fmt.Errorf("%v", panicErr)
```
jmcatamney, line 176: One more nitpick: `fmt.Errorf` returns an error, so `fmt.Errorf("%v", panicErr)` is equivalent to just passing `panicErr` into the channel directly.

AJR-VMware: I checked when writing this and technically `recover` returns an `any`, so I thought an extra layer of protection here would be appropriate.

jmcatamney (approved):

PR #657, opened 2023-05-12
Fix filtering of tables with inheritance

jmcatamney:
If tables are included in a backup set that inherit from, or are inherited by, other tables and those other parent or child tables are not included in the backup set, gpbackup should ignore those inheritance links when constructing CREATE TABLE statements for included tables.
    
This commit also adds a --no-inherit flag, which will cause filtered backups to ignore child tables of included tables to preserve the old behavior.

### DIFF
--- a/options/options.go
+++ b/options/options.go
@@ -2,6 +2,7 @@ package options
 
 import (
 	"fmt"
+	"sort"
 	"strings"
 
 	"github.com/greenplum-db/gp-common-go-libs/dbconn"
@@ -214,7 +215,7 @@ func (o *Options) ExpandIncludesForPartitions(conn *dbconn.DBConn, flags *pflag.
 		return err
 	}
 
-	allFqnStructs, err := o.getUserTableRelationsWithIncludeFiltering(conn, quotedIncludeRelations)
+	allFqnStructs, err := o.getUserTableRelationsWithIncludeFiltering(conn, quotedIncludeRelations, MustGetFlagBool(flags, NO_INHERIT))
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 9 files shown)


options/options.go
```diff
@@ -281,24 +287,56 @@ func (o *Options) recurseTableDepend(conn *dbconn.DBConn, includeOids []string,
...
+	} else if recurseSource == "parent" {
+		// Given a parent partition table, find all the child partitions below it
+		dependQuery = parentQuery
+	} else if recurseSource == "both" {
+		// Given a table that inherits or is inherited by another table, find all
+		// other tables above and below it
+		dependQuery = fmt.Sprintf(`%[1]s
+			UNION %[2]s`, childQuery, parentQuery)
 	} else {
 		gplog.Error("Please fix calling of this function recurseTableDepend. Argument recurseSource only accepts 'parent' or 'child'.")
```
AJR-VMware, line 341: This error message needs to be updated.  It's internal-use-only, but should still be kept correct.

options/options.go
```diff
@@ -341,82 +379,56 @@ func (o *Options) recurseTableDepend(conn *dbconn.DBConn, includeOids []string,
...
+	parentOids, err := o.recurseTableDepend(connectionPool, includeOids, "child", o.isLeafPartitionData)
+	if err != nil {
+		return nil, err
+	}
+	if len(parentOids) > 0 {
+		parentAndExternalPartitionFilter = fmt.Sprintf(`OR c.oid IN (%s)`, strings.Join(parentOids, ", "))
+	}
+	if !no_inherit {
+		expandedOids := make([]string, 0)
+		for {
```
AJR-VMware, line 420: Infinite loops make me nervous.  I don't *think* there should be a way to get stuck here, but we've had hangs in places we didn't expect before.  Would it be possible to add an upper limit to loop iterations here (even if it's huge, like 100k or something), so that we will eventually fail out in case something wonky happens?

AJR-VMware (approved):
One more nit, but otherwise LGTM!

PR #658, opened 2023-05-13
Compress hang

AJR-VMware:
A previous change to this code restructured how we track which tables
are being backed up, and their status.  This led to the case where, if a
piped copy command failed, the error was captured but never displayed.
Instead we'd spin forever on an "Unknown" state in our sync.Map, causing
permanent hangs.

This PR restores the original behavior, where if a pipe fails the
backup of the table is marked "Complete" in the sync.map, and the error
is handled correctly downstream.

Given the issues with this function misbehaving, it needs better
coverage.  This PR adds some basic test coverage of it.  We also
tweak how FilePathInfo is structured, so that it is simpler to mock out
where data is copied out to during backup.

### DIFF
--- a/integration/data_backup_test.go
+++ b/integration/data_backup_test.go
@@ -1,0 +1,88 @@
+package integration
+
+import (
+	"fmt"
+	"os"
+	"path"
+
+	"github.com/blang/semver"
+	"github.com/greenplum-db/gp-common-go-libs/testhelper"
+	"github.com/greenplum-db/gpbackup/backup"
+	"github.com/greenplum-db/gpbackup/filepath"
+	"github.com/greenplum-db/gpbackup/utils"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("backup integration tests", func() {
+	Describe("BackupDataForAllTables", func() {
+		var (
+			testTables             []backup.Table
+			fpInfo                 filepath.FilePathInfo
+			origFPInfo             filepath.FilePathInfo
+			origPipeThroughProgram utils.PipeThroughProgram
+		)
+		BeforeEach(func() {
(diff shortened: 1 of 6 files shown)


integration/data_backup_test.go
```diff
@@ -0,0 +1,87 @@
...
+			testhelper.AssertQueryRuns(connectionPool, `CREATE TABLE dataTest.testtable2 (i int) DISTRIBUTED BY (i);`)
+		})
+		AfterEach(func() {
+			testhelper.AssertQueryRuns(connectionPool, `DROP SCHEMA dataTest CASCADE;`)
+			os.RemoveAll(fpInfo.BaseDataDir)
+			backup.SetFPInfo(origFPInfo)
+			utils.SetPipeThroughProgram(origPipeThroughProgram)
+		})
+		It("backs up multiple tables with valid data", func() {
+			Ω(func() { backup.BackupDataForAllTables(testTables) }).ShouldNot(Panic())
```
jmcatamney, line 62: `Ω` -> `Expect`

We don't use the symbol in gpbackup tests, and gomega now recommends against using it.

integration/data_backup_test.go
```diff
@@ -0,0 +1,87 @@
...
+			// as waiting around for 10 minutes while this times out is not ideal.  This is good
+			// for now.
+			dummyPipeThrough := utils.PipeThroughProgram{
+				Name:          "dummy",
+				OutputCommand: "doesnotexist",
+				InputCommand:  "doesnotexist",
+				Extension:     ".dne",
+			}
+			utils.SetPipeThroughProgram(dummyPipeThrough)
+			Ω(func() { backup.BackupDataForAllTables(testTables) }).Should(Panic())
```
jmcatamney, line 84: See above comment.

jmcatamney: Also, we have a `testhelper.ShouldPanicWithMessage` helper function that can be used in this scenario.

jmcatamney (approved):

PR #663, opened 2023-05-19
Improve BackupDataForAllTables

AJR-VMware:
Previous changes in support distributed snapshots for backing up data introduced bugs in how we handle errors when piping copy commands to the server. A prior attempt at fixing this was insufficient, so here we overhaul how the goroutines doing distributed data backup inform of errors, and how those errors are handled.

Also, some changes are necessary to our integration test suite to support testing of this code path. Specifically, the distributed code path is only invoked if there are multiple connections to the cluster available. So, we update to run all our tests with two connections open.

### DIFF
--- a/backup/data.go
+++ b/backup/data.go
@@ -149,8 +149,8 @@ func BackupDataForAllTables(tables []Table) []map[uint32]int64 {
 	 */
 	tasks := make(chan Table, len(tables))
 	var oidMap sync.Map
+	var isErroredBackup atomic.Bool
 	var workerPool sync.WaitGroup
-	var copyErr error
 	// Record and track tables in a hashmap of oids and table states (preloaded with value Unknown).
 	// The tables are loaded into the tasks channel for the subsequent goroutines to work on.
 	for _, table := range tables {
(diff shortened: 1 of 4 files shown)


backup/data.go
```diff
@@ -183,34 +183,33 @@ func BackupDataForAllTables(tables []Table) []map[uint32]int64 {
...
 					if pgErr, ok := err.(*pgconn.PgError); ok && pgErr.Code != PG_LOCK_NOT_AVAILABLE {
-						copyErr = err
-						oidMap.Store(table.Oid, Deferred)
+						isErroredBackup.Store(true)
 						err = connectionPool.Rollback(whichConn)
 						if err != nil {
 							gplog.Warn("Worker %d: %s", whichConn, err)
 						}
-						continue
+						gplog.Fatal(fmt.Errorf("Unexpectedly unable to take lock on table %s, %s", table.FQN(), pgErr.Error()), "")
```
bmdoil, line 212: Should we add a test to exercise this code path?

bmdoil (approved):
LGTM. Glad we're improving the error handling. On a related note, we still have the FIXME for `BackupDataForAllTables` 
```
* FIXME: Simplify BackupDataForAllTables by having one function for snapshot workflow and
* another without, then extract common portions into their own functions.
```
Since this part is out of scope for this PR, I'd rename the commit to something like `Improve error handling in BackupDataForAllTables`

backup/data.go
AJR-VMware: Probably a good idea, yeah. I'll write one up.

AJR-VMware: Done!

PR #672, opened 2023-06-02
Remove prodataaccess for GP7 and above

khuddlefish:
This catalog column was not present in upstream and maybe not very useful so it was removed from the catalog in GP7. The grammar will still be accepted in GP7 but ignored.

Dev pipeline here: https://dp.ci.gpdb.pivotal.io/teams/main/pipelines/dev:gpbackup_remove-prodataaccess

GPDB7 jobs are still red but it looks like they're failing due to some other issue


backup/queries_functions_test.go
```diff
@@ -82,21 +82,44 @@ SET baz to abc`))
...
-				sql.NullString{String: "mock_args3", Valid: true}, nil, sql.NullString{String: "mock_resulttype3", Valid: true}, "mock_volatility3",
-				false, false, "", 0, 0, "mock_dataaccess3", "mock_language3"}
-			rowNullResultType := []driver.Value{"4", "mock_schema4", "mock_table4", false, "mock_funcbody4", "mock_path4",
-				sql.NullString{String: "mock_args4", Valid: true}, sql.NullString{String: "mock_identargs4", Valid: true}, nil, "mock_volatility4",
-				false, false, "", 0, 0, "mock_dataaccess4", "mock_language4"}
+			var (
+				header                                                  []string
+				rowGood, rowNullArg, rowNullIdentArg, rowNullResultType []driver.Value
+			)
+			if connectionPool.Version.AtLeast("7") {
```
jmcatamney, line 89: Could you add a comment here noting that the (only) difference between the two versions is the presence/absence of the prodataaccess column?  With all the mocking boilerplate, it's hard to tell at first glance.

jmcatamney:
Aside from my comment about the comment, everything looks good.  I can approve this once that comment is added.

backup/queries_functions.go
```diff
@@ -170,7 +170,6 @@ func GetFunctions(connectionPool *dbconn.DBConn) []Function {
 				FROM pg_options_to_table(proconfig)), ' '), '') AS proconfig,
 			procost,
 			prorows,
-			prodataaccess,
```
AJR-VMware, line 173: A comment on the query string noting the differences would be good to add.

AJR-VMware (approved):
Aside from nit about comment, LGTM.

backup/queries_functions_test.go
khuddlefish: sure thing

backup/queries_functions.go
khuddlefish: There were more differences than just that that already existing and we hadn't put a comment for those so I was continuing that pattern. I wasn't sure what all of the other differences were and why (but I can see there are some) so I didn't want to make it look like that was the only change by only mentioned it.

AJR-VMware: Fair enough.  At some point I guess we should go back and try to document the differences.  Fine for now, though, certainly.

PR #673, opened 2023-06-06
Reduce CI flakiness

AJR-VMware:
We've had issues with our concourse CI handling some of these tests poorly. Add some flake protection to help improve the signal to noise ratio on our tests.

### DIFF
--- a/end_to_end/end_to_end_suite_test.go
+++ b/end_to_end/end_to_end_suite_test.go
@@ -1721,7 +1721,7 @@ LANGUAGE plpgsql NO SQL;`)
 			})
 		})
 	})
-	Describe("Restore to a different-sized cluster", func() {
+	Describe("Restore to a different-sized cluster", FlakeAttempts(3), func() {
 		if useOldBackupVersion {
 			Skip("This test is not needed for old backup versions")
 		}
@@ -1763,6 +1763,7 @@ LANGUAGE plpgsql NO SQL;`)
 				// These hangs are still being observed only in CI, and a definitive RCA has not yet been accomplished
 				completed := make(chan bool)
 				defer func() { completed <- true }() // Whether the test succeeds or fails, mark it as complete
+				defer GinkgoRecover()
 				go func() {
 					// No test run has been observed to take more than a few minutes without a hang,
 					// so loop 5 times and check for success after 1 minute each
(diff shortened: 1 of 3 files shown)


end_to_end/signal_handler_test.go
```diff
@@ -32,11 +32,10 @@ var _ = Describe("Signal handler tests", FlakeAttempts(3), func() {
 			cmd := exec.Command(gpbackupPath, args...)
 			go func() {
 				/*
-				* We use a random delay for the sleep in this test (between
-				* 1.0s and 1.5s) so that gpbackup will be interrupted at a
-				* different point in the backup process every time to help
-				* catch timing issues with the cleanup.
-				*/
+				 * We use a random delay for the sleep in this test (between 1.0s and 1.5s) so that
+				 * gpbackup will be interrupted at a different point in the backup process every time to hel catch
```
khuddlefish, line 36: teeny tiny nit. You have `hel` instead of `help`

khuddlefish (approved):
Tiny typo. Also not sure why the whitespace is changing in the comments, but I guess it's fine.

end_to_end/end_to_end_suite_test.go
```diff
@@ -1763,6 +1763,7 @@ LANGUAGE plpgsql NO SQL;`)
 				// These hangs are still being observed only in CI, and a definitive RCA has not yet been accomplished
 				completed := make(chan bool)
 				defer func() { completed <- true }() // Whether the test succeeds or fails, mark it as complete
+				defer GinkgoRecover()
```
khuddlefish, line 1766: I think I read GinkgoRecover goes inside the go func?

AJR-VMware:
> Tiny typo. Also not sure why the whitespace is changing in the comments, but I guess it's fine.

Jamie and I both have it set up to run `gofmt` on save, to keep the code aligned to the "official" linter rules.  We've noticed that some files were most recently edited without that setup in place.  As a general rule, as long as it doesn't obscure the commit too much, we usually just let `gofmt` restyle the code on all our work.  The alternative would be a "big bang" reformat and that seems excessive.

end_to_end/end_to_end_suite_test.go
AJR-VMware: I was patterning it off our only other usage:
https://github.com/greenplum-db/gpbackup/blob/81507afd25fff9b5af0f2de4112d7eabba1e2f48/end_to_end/end_to_end_suite_test.go#LL398C8-L398C18

PR #676, opened 2023-06-12
Modify partition filtering behavior

jmcatamney:
When including a given table in the filter, we now include that table's parent tables (and their parents, and so on), that table's children, and any other parents of those children, such that all INHERITS statements in any related tables are satisfied.

This is a tweak to the behavior introduced in https://github.com/greenplum-db/gpbackup/commit/b290ade92332fce52d3b90288d3ef554dd122c51, so that we don't follow *every* inheritance link and collect any "sibling" tables of the originally-included table that do not share common children with it.

(The second and third commits are separated out for review purposes, but will be squashed with the first one before merging.)


integration/options_integration_test.go
```diff
@@ -221,49 +212,40 @@ FORMAT 'csv';`)
 
 			tables := subject.GetIncludedTables()
 			sort.Strings(tables)
-			Expect(tables).To(HaveLen(len(expectedTableNames)))
+			Expect(tables).To(Equal(expectedTableNames))
+		})
+		It("returns only the parent table if the filter includes a leaf table and leaf-partition-data and no-inherits are set", func() {
```
khuddlefish, line 217: reading this I thought it would only include the parent table and not the specified included table. But I see it does the behavior we expect. The wording is maybe a bit confusing.

integration/options_integration_test.go
```diff
@@ -327,7 +337,89 @@ FORMAT 'csv';`)
...
+			defer testhelper.AssertQueryRuns(connectionPool, "DROP TABLE public.other CASCADE")
+			defer testhelper.AssertQueryRuns(connectionPool, "DROP TABLE public.unrelated")
+
+			subject, err := options.NewOptions(backupCmdFlags)
+			Expect(err).To(Not(HaveOccurred()))
+
+			err = subject.ExpandIncludesForPartitions(connectionPool, backupCmdFlags)
+			Expect(err).To(Not(HaveOccurred()))
+
+			expectedTableNames := []string{
```
khuddlefish, line 388: I wasn't initially expecting this to pick up grandchild. Is that because it follows the child inheritance chain all the way down to leaves?

integration/options_integration_test.go
jmcatamney: Yeah, the "only" is in reference to the previous test where it retrieves parents and external partitions, I can tweak that.

integration/options_integration_test.go
jmcatamney: Correct.  It has to traverse multiple levels of children to handle multi-level partition tables.

PR #683, opened 2023-07-12
Correctly restore tables with inherited generated columns

jmcatamney:
In GPDB 7, if a table has a generated column and another table inherits that table, the child table cannot use the GENERATED keyword to define a default for that column (whether or not it matches that of the parent's corresponding column) or GPDB will refuse to create the table.

Thus, when backing up such child tables, we need to omit the GENERATED statement from the definition of any attributes inherited from another table.

### DIFF
--- a/backup/predata_relations.go
+++ b/backup/predata_relations.go
@@ -207,7 +207,12 @@ func printColumnDefinitions(metadataFile *utils.FileWithByteCount, columnDefs []
 		}
 		if column.HasDefault {
 			if column.AttGenerated != "" {
-				line += fmt.Sprintf(" GENERATED ALWAYS AS %s %s", column.DefaultVal, column.AttGenerated)
+				// Unlike most keywords, GENERATED cannot be applied to a column that inherits from a parent table,
+				// even if the specified generation expression is identical to that of the column it inherits,
+				// so we skip printing it in that case.
+				if !column.IsInherited {
+					line += fmt.Sprintf(" GENERATED ALWAYS AS %s %s", column.DefaultVal, column.AttGenerated)
+				}
 			} else {
 				line += fmt.Sprintf(" DEFAULT %s", column.DefaultVal)
 			}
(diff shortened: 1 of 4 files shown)


backup/queries_table_defs.go
```diff
@@ -348,7 +353,15 @@ func GetColumnDefinitions(connectionPool *dbconn.DBConn) map[uint32][]ColumnDefi
 		coalesce(array_to_string(ARRAY(SELECT option_name || ' ' || quote_literal(option_value) FROM pg_options_to_table(attfdwoptions) ORDER BY option_name), ', '), '') AS fdwoptions,
 		CASE WHEN a.attcollation <> t.typcollation THEN quote_ident(cn.nspname) || '.' || quote_ident(coll.collname) ELSE '' END AS collation,
 		coalesce(sec.provider,'') AS securitylabelprovider,
-		coalesce(sec.label,'') AS securitylabel
+		coalesce(sec.label,'') AS securitylabel,
+		EXISTS (
+			SELECT 1
+			FROM pg_inherits AS i
+			JOIN pg_attribute AS a2
```
AJR-VMware, line 360: Can we explicitly specify the type of join, please?  I know INNER is default, but for clarity it'd be nice to have.

backup/predata_relations.go
```diff
@@ -207,7 +207,12 @@ func printColumnDefinitions(metadataFile *utils.FileWithByteCount, columnDefs []
 		}
 		if column.HasDefault {
 			if column.AttGenerated != "" {
-				line += fmt.Sprintf(" GENERATED ALWAYS AS %s %s", column.DefaultVal, column.AttGenerated)
+				// Unlike most keywords, GENERATED cannot be applied to a column that inherits from a parent table,
+				// even if the specified generation expression is identical to that of the column it inherits,
+				// so we skip printing it in that case.
+				if !column.IsInherited {
```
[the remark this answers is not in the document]
jmcatamney: If we did that, then if a column _is_ inherited it would `else` into printing `DEFAULT`.

backup/queries_table_defs.go
jmcatamney: Can do.

AJR-VMware (approved):
LGTM!

PR #689, opened 2023-07-20
Correct the end time printed to the report file

jmcatamney:
Due to changes made to support the new history database, gpbackup was setting the end time too early in the backup such that the reported end time did not match the actual time displayed to the user.

Now, gpbackup will update that end time when it updates the backup status so that the time is reported correctly.


AJR-VMware (approved):
LGTM

PR #694, opened 2023-07-27
Implement --report-dir option for gprestore

InnerLife0:
New option allows to create report files in directory different from --backup-dir. Main code changes were made for filepath class. Now it takes new option as deafult path for reports with fallback to content dir. Unit and end_to_end tests were updated to show mentioned behavior.

### DIFF
--- a/filepath/filepath.go
+++ b/filepath/filepath.go
@@ -83,12 +83,8 @@ func (backupFPInfo *FilePathInfo) GetDirForContent(contentID int) string {
 	return path.Join(backupFPInfo.SegDirMap[contentID], "backups", backupFPInfo.Timestamp[0:8], backupFPInfo.Timestamp)
 }
 
-func (backupFPInfo *FilePathInfo) IsUserSpecifiedReportDir() bool {
-	return backupFPInfo.UserSpecifiedReportDir != ""
-}
-
 func (backupFPInfo *FilePathInfo) GetDirForReport(contentID int) string {
-	if backupFPInfo.IsUserSpecifiedReportDir() {
+	if backupFPInfo.UserSpecifiedReportDir != "" {
 		segDir := fmt.Sprintf("%s%d", backupFPInfo.UserSpecifiedSegPrefix, contentID)
 		return path.Join(backupFPInfo.UserSpecifiedReportDir, segDir, "backups", backupFPInfo.Timestamp[0:8], backupFPInfo.Timestamp)
 	}
--- a/filepath/filepath.go
+++ b/filepath/filepath.go
@@ -27,6 +27,7 @@ type FilePathInfo struct {
 	UserSpecifiedBackupDir string
 	UserSpecifiedSegPrefix string
 	BaseDataDir            string
+	UserSpecifiedReportDir string
 }
 
 func NewFilePathInfo(c *cluster.Cluster, userSpecifiedBackupDir string, timestamp string, userSegPrefix string, useMirrors ...bool) FilePathInfo {
@@ -34,6 +35,7 @@ func NewFilePathInfo(c *cluster.Cluster, userSpecifiedBackupDir string, timestam
 	backupFPInfo.PID = os.Getpid()
 	backupFPInfo.UserSpecifiedBackupDir = userSpecifiedBackupDir
 	backupFPInfo.UserSpecifiedSegPrefix = userSegPrefix
+	backupFPInfo.UserSpecifiedReportDir = ""
 	backupFPInfo.Timestamp = timestamp
 	backupFPInfo.SegDirMap = make(map[int]string)
 	backupFPInfo.BaseDataDir = "<SEG_DATA_DIR>"
@@ -49,6 +51,14 @@ func NewFilePathInfo(c *cluster.Cluster, userSpecifiedBackupDir string, timestam
 	}
 
 	return backupFPInfo
+}
+
+/*
+ * Set user specified dir for report.
+ * Currently used for restore only.
+ */
+func (backupFPInfo *FilePathInfo) SetReportDir(userSpecifiedReportDir string) {
+	backupFPInfo.UserSpecifiedReportDir = userSpecifiedReportDir
 }
 
 /*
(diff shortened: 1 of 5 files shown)


AJR-VMware:
Hi! 
Thanks for your contribution!

In reviewing this PR, the team had a few questions:
1. What is the use case for this new flag?
2. We recently made some changes so that if the report or errors files were unable to be written, the backup itself would not fail.  Does this change the need/use for this flag?

InnerLife0:
> Hi! Thanks for your contribution!
> 
> In reviewing this PR, the team had a few questions:
> 
> 1. What is the use case for this new flag?
> 2. We recently made some changes so that if the report or errors files were unable to be written, the backup itself would not fail.  Does this change the need/use for this flag?

Hi! Thanks for answer.
1. The popular scenario: We want to restore a backup which is mounted in read-only mode due to security reasons. We still need report files to debug something, to keep a history, etc.
2. No, as we need report files to be existed.

filepath/filepath.go
```diff
@@ -73,6 +83,18 @@ func (backupFPInfo *FilePathInfo) GetDirForContent(contentID int) string {
 	return path.Join(backupFPInfo.SegDirMap[contentID], "backups", backupFPInfo.Timestamp[0:8], backupFPInfo.Timestamp)
 }
 
+func (backupFPInfo *FilePathInfo) IsUserSpecifiedReportDir() bool {
```
AJR-VMware, line 86: We do not need a dedicated checker function, given that it's one line and only called once anyway.  Let's remove it, and just check the `==""` logic directly.

filepath/filepath.go
```diff
@@ -73,6 +83,18 @@ func (backupFPInfo *FilePathInfo) GetDirForContent(contentID int) string {
...
+	return backupFPInfo.UserSpecifiedReportDir != ""
+}
+
+func (backupFPInfo *FilePathInfo) GetDirForReport(contentID int) string {
+	if backupFPInfo.IsUserSpecifiedReportDir() {
+		segDir := fmt.Sprintf("%s%d", backupFPInfo.UserSpecifiedSegPrefix, contentID)
+		return path.Join(backupFPInfo.UserSpecifiedReportDir, segDir, "backups", backupFPInfo.Timestamp[0:8], backupFPInfo.Timestamp)
```
AJR-VMware, line 93: For a flag like this, if the user has indicated a directory we should not create structure underneath it.  Let's just return `backupFPInfo.UserSpecifiedReportDir` in that case, and work directly with what the user provided.

This is especially relevant because we're going to be changing the backup-dir structure to be simpler in our next release, to simplify resize-restore use cases.

options/flag.go
```diff
@@ -120,6 +121,7 @@ func SetRestoreFlagDefaults(flagSet *pflag.FlagSet) {
 	flagSet.Bool(LEAF_PARTITION_DATA, false, "For partition tables, create one data file per leaf partition instead of one data file for the whole table")
 	flagSet.Bool(RUN_ANALYZE, false, "Run ANALYZE on restored tables")
 	flagSet.Bool(RESIZE_CLUSTER, false, "Restore a backup taken on a cluster with more or fewer segments than the cluster to which it will be restored")
+	flagSet.String(REPORT_DIR, "", "The absolute path of the directory to which all report files will be written")
```
AJR-VMware, line 124: Instead of "all report files", let's say "restore report and error tables".  It's a short list, might as well be explicit.

filepath/filepath.go
```diff
@@ -51,6 +53,14 @@ func NewFilePathInfo(c *cluster.Cluster, userSpecifiedBackupDir string, timestam
 	return backupFPInfo
 }
 
+/*
+ * Set user specified dir for report.
+ * Currently used for restore only.
+ */
+func (backupFPInfo *FilePathInfo) SetReportDir(userSpecifiedReportDir string) {
```
AJR-VMware, line 60: I'm not sure we need a dedicated setter function.  Is this just to make testing simpler?

AJR-VMware (commented):
We'd like to bring this feature in.  There are a few changes that would be good.  Please see comments, let me know if you have any thoughts.  Thanks!

filepath/filepath.go
InnerLife0: > For a flag like this, if the user has indicated a directory we should not create structure underneath it. Let's just return `backupFPInfo.UserSpecifiedReportDir` in that case, and work directly with what the user provided.

What is the point? What is the difference with backupDir scenario?

> This is especially relevant because we're going to be changing the backup-dir structure to be simpler in our next release, to simplify resize-restore use cases.

I can't predict this changes, but I see two ways.
1. Wait for the next release and try to change current implementation according to new GetDirForContent() implementation.
2. Left it as is and change both functions all together as a part of new release.

filepath/filepath.go
InnerLife0: This is done to keep interface of NewFilePathInfo() the same and avoid a lot of code changes.

options/flag.go
InnerLife0: Done

filepath/filepath.go
InnerLife0: Done

filepath/filepath.go
InnerLife0: 1. Despite the fact, I still don't know how it'll look like, I understand your idea and it looks nice. Now I understand your point.
2. Our idea is to keep compatibility and continuity with backup-dir option. We think that report-dir is a dependent option, so it should respect backup-dir behavior and act the same when some default value is given explicitly and when there is no option given.

khuddlefish: Hi @InnerLife0, I'd say the difference between this option and backupDir is that backupDir will have many more files that go under it so we needed to impose some structure there. For this option it only gets one or two files so it's a bit silly to create that extra directory structure when redirecting the report files.

That being said, I think we're fine with merging this how it currently is. We will just plan to change the directory structure for this option very soon with that new feature Andrew mentioned.

AJR-VMware (approved):
Looks good!  Thanks again for the contribution!

PR #695, opened 2023-07-27
Only include table names in error_tables_metadata

jmcatamney:
As the error_tables_metadata file created by a failed gprestore is intended to be used with --include-table and --exclude-table filters on a subsequent restore, it should not include the names of any non-table objects that failed to be restored.


end_to_end/end_to_end_suite_test.go
```diff
@@ -858,6 +858,39 @@ var _ = Describe("backup and restore end to end tests", func() {
 			Expect(tables).To(HaveLen(len(expectedErrorTablesMetadata)))
 			_ = os.Remove(files[0])
 		})
+		It(`ensure gprestore of corrupt backup with --on-error-continue only logs tables in error_tables_metadata file`, func() {
```
AJR-VMware, line 861: Nit: we uses the grammatical convention for these.  Would be "ensures".

end_to_end/end_to_end_suite_test.go
```diff
@@ -858,6 +858,39 @@ var _ = Describe("backup and restore end to end tests", func() {
...
+			if segmentCount != 3 {
+				Skip("Restoring from a tarred backup currently requires a 3-segment cluster to test.")
+			}
+			// The functionality works on 5, but it's currently difficult to make a saved backup on a 5 cluster.
+			// TODO: Remove this and re-do the saved backup once we get local 5 testing working again.
```
AJR-VMware, line 866: Are we planning to get local 5 testing working again?  This seems like it'll be a permanent TODO.

end_to_end/end_to_end_suite_test.go
jmcatamney: Blame the other tests I copy-pasted.  I'll fix all three.

end_to_end/end_to_end_suite_test.go
jmcatamney: We haven't given up on making it work, as far as I know.  We can chat with the team and make a decision on that.

AJR-VMware (approved):

PR #698, opened 2023-08-07
Backup privilege statements as separate entries

InnerLife0:
Before patch, all privilege statements placed in metadata file as one pre-data
section entry. `editStatementsRedirectSchema()`, used by schema redirection,
[replace](https://github.com/greenplum-db/gpbackup/blob/7d11d77a4e956069188340d8c5401dbfa9ec1cc9/restore/restore.go#L354) only first schema name occurrence in entry. This caused a bug - only
first privilege statement was modifed. This, in turn, caused vaious errors.
`gprestore`, used on same database, applied most of privilege statements to
original schema. Or, when used on different database, caused `schema does not
exist` error.

From now, `GetPrivilegesStatements()` returns a set of separate statements,
which later used as separate entries. `strings.TrimSpace()` was removed as all
statements are already trimmed.

Existed tests were modified as separate entries have different format. New test
shows correct privilege applying.
______
This is an example of how privilege entry looked after `editStatementsRedirectSchema()` applying.
```
REVOKE ALL ON TABLE schema_to.test_table FROM PUBLIC;
REVOKE ALL ON TABLE schema_from.test_table FROM innerlife;
GRANT ALL ON TABLE schema_from.test_table TO innerlife;
GRANT SELECT ON TABLE schema_from.test_table TO test_user;
```
Only first statement has correct schema name (`schema_to`)
After patch, all statements placed inside each own entries, so `editStatementsRedirectSchema()` modification applied to each entry separately.
```
REVOKE ALL ON TABLE schema_to.test_table FROM PUBLIC;


REVOKE ALL ON TABLE schema_to.test_table FROM innerlife;


GRANT ALL ON TABLE schema_to.test_table TO innerlife;


GRANT SELECT ON TABLE schema_to.test_table TO test_user;
```

### DIFF
--- a/backup/metadata_globals_test.go
+++ b/backup/metadata_globals_test.go
@@ -50,9 +50,9 @@ var _ = Describe("backup/metadata_globals tests", func() {
 				`CREATE DATABASE testdb TEMPLATE template0;`,
 				`COMMENT ON DATABASE testdb IS 'This is a database comment.';`,
 				`ALTER DATABASE testdb OWNER TO testrole;`,
-				`REVOKE ALL ON DATABASE testdb FROM PUBLIC;
-REVOKE ALL ON DATABASE testdb FROM testrole;
-GRANT TEMPORARY,CONNECT ON DATABASE testdb TO testrole;`,
+				`REVOKE ALL ON DATABASE testdb FROM PUBLIC;`,
+				`REVOKE ALL ON DATABASE testdb FROM testrole;`,
+				`GRANT TEMPORARY,CONNECT ON DATABASE testdb TO testrole;`,
 				`SECURITY LABEL FOR dummy ON DATABASE testdb IS 'unclassified';`}
 			testutils.AssertBufferContents(tocfile.GlobalEntries, buffer, expectedStatements...)
 		})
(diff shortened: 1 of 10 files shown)


jmcatamney:
Thanks for identifying this issue!  We're not sure whether splitting up the privileges block into separate entries is the best fix for this, as we've seen that increasing the number of statements gprestore has to execute during the restore can impact restore performance; I'm going to play around with this change at scale to see if we want to go with this PR or whether we want to instead modify the replacement logic in `editStatementsRedirectSchema` to work with the existing structure.

InnerLife0:
> Thanks for identifying this issue! We're not sure whether splitting up the privileges block into separate entries is the best fix for this, as we've seen that increasing the number of statements gprestore has to execute during the restore can impact restore performance; I'm going to play around with this change at scale to see if we want to go with this PR or whether we want to instead modify the replacement logic in `editStatementsRedirectSchema` to work with the existing structure.

Our [second](https://github.com/arenadata/gpbackup/pull/32/files) approach was to modify `editStatementsRedirectSchema`, but fixing the root cause of the problem was an unanimous decision.

I can't say much about overall performance, but our test toc file (2,6M) from a real backup became bigger by 34%, and metadata file (326M) became bigger by 0,004%.

AJR-VMware:
Due to the overhead of additional statements (especially with GPDB7 partitioning being so verbose and chewing up such overhead already) we decided to go a different route for addressing this bug.  

It has been fixed by commit: `7a2ad0ca95d9d72c3714d4bf731f097e38ed5f27`

Thank you very much for bringing this to our attention!

PR #699, opened 2023-08-08
Improve name handling for table filters

AJR-VMware:
Our previous processing could not handle dots in table names, and was not escaping double quotes if a user included them. This commit handling should achieve parity with what pg_dump is able to process.

In addition, the input filter lists were re-processed in multiple places throughout the codebase, causing some room for quote_ident to possibly be run on the same name more than once.  To clean things up, we now use a struct representation that is populated one time, and then referenced for the remainder of the run.  We now process based on OID wherever the filter lists are needed.  This required fairly wide-ranging changes to the integration test suite.

This commit leaves one TODO that should be addressed.  Our options module needs access to this struct, but cannot import the backups module due to circular import errors.  This presents a challenge, and the current solution is to have identical options.Relation and backup.Relation structs, and a routine to convert from one to the other. This is obviously inefficient, but the alternative would be to pull large chunks of logic "upstream" into options so that options can be the official home of this struct, and that seems very invasive to the codebase.  A better solution would be good, but none seems clear at the moment.


backup/validate_test.go
```diff
@@ -62,11 +62,13 @@ var _ = Describe("backup/validate tests", func() {
 	})
 	Describe("ValidateTablesExist", func() {
 		var tableRows, partitionTables, schemaAndTable, schemaAndTable2 *sqlmock.Rows
+		// var backupCmdFlags *pflag.FlagSet
```
jmcatamney, line 65: I assume this and line 71 should be removed, not just commented out?

integration/options_integration_test.go
```diff
@@ -121,44 +121,58 @@ bar";
 		AfterEach(func() {
 			testhelper.AssertQueryRuns(connectionPool, `DROP TABLE public."CAPpart"`)
 		})
-
 		It("adds parent table when child partition with special chars is included", func() {
-			err := backupCmdFlags.Set(options.INCLUDE_RELATION, `public.CAPpart_1_prt_girls`)
+			parentTable := `public.CAPpart`
+			partTable := `public.CAPpart_1_prt_girls`
```
jmcatamney, line 126: Should these be double-quoted like the changes on lines 104-106?

integration/options_integration_test.go
```diff
@@ -101,13 +101,13 @@ bar";
...
-				`CAPschema.BAR`,
-				`CAPschema.baz`,
+				`public."BAR"`,
+				`"CAPschema"."BAR"`,
+				`"CAPschema".baz`,
 				`public.foo_bar`,
-				`public.foo ~#$%^&*()_-+[]{}><\|;:/?!bar`,
-				"public.tab\tbar",     // important to use double quotes to allow \t to become tab
-				"public.newline\nbar", // important to use double quotes to allow \n to become newline
+				`public."foo ~#$%^&*()_-+[]{}><\|;:/?!bar"`,
```
jmcatamney, line 108: Since we now support periods in schema and table names, can we add a `.` to this one and add another table with a dot-containing schema to cover that?

integration/options_integration_test.go
AJR-VMware: It works either way.  We have not forced users to quote-wrap such inputs in the past, even though the server does.  For the sake of not breaking things that used to work, I feel we're obliged to handle both.  Having them both in the test suite seems good, given that.

options/options.go
```diff
@@ -178,82 +177,53 @@ func QuoteTableNames(conn *dbconn.DBConn, tableNames []string) ([]string, error)
...
-		return err
-	}
+		if schemaNameIsQuoted && tableNameIsQuoted {
+			quotesPattern = regexp.MustCompile("^\"|\"$|\"\\.\"")
+		} else if schemaNameIsQuoted {
+			quotesPattern = regexp.MustCompile("^\"|\"\\.")
+		} else if tableNameIsQuoted {
+			quotesPattern = regexp.MustCompile("\"$|\\.\"")
+		} else {
+			quotesPattern = regexp.MustCompile("\\.")
```
jmcatamney, line 196: Nit: Could you backtick-quote these for better readability? e.g. "^\\"|\\"$|\\"\\\\.\\"" -> \`^"|"$|"."\`

integration/options_integration_test.go
AJR-VMware: Done!

options/options.go
AJR-VMware: Sure!

integration/options_integration_test.go
jmcatamney: If it works without quotes, I'm fine with that.

options/options_test.go
```diff
@@ -214,11 +214,25 @@ var _ = Describe("options", func() {
...
-				{SchemaName: `FO!@#`, TableName: `BAR`},
+			tableList := []string{
+				"foo.Bar",
+				"FOO.Bar",
+				"FO!@#.BAR",
+				"ezschema.\"myt.able\"",             // dots in one part
+				"\"mys.chema\".\"othert.able\"",     // dots in both parts
+				"easy.\"tryme\"\"",                  // quotes with escaping
+				"\"mys.chema\".\"anoth\"ert.able\"", // quotes without escaping
+				"easy.no\"quotes",                   // quotes without wrapping or escaping
```
jmcatamney, line 225: Let's backtick-quote these as well.

AJR-VMware: Done!

backup/validate_test.go
AJR-VMware: Yep!  Done!

options/options.go
AJR-VMware: And done!

jmcatamney (approved):

PR #700, opened 2023-08-08
Pipeline platform cleanup

khuddlefish:
- Make resource names more consistent
- Use test images everywhere rather than build images (I don't believe we need
  the build images since they are intended for building gpdb)
- Use gpdb6 images for gpdb6 jobs and gpdb7 images for gpdb7 jobs
- Be consistent with platforms within a job
- Update jobs to run on newer platforms and gp versions
- Update default values for what we use most (rocky8, google-nvme-block-device)
- Remove several references to /opt/gcc_env.sh since that has been removed from
  newer images. It is still needed for gpdb5 images
- add back libyaml resource for build ddboost on centos6

test pipeline: https://prod.ci.gpdb.pivotal.io/teams/main/pipelines/gpbackup_platform_cleanup


AJR-VMware (approved):
This all looks great!

While we're doing all this, the only other cruft I see would be that the file is names "setup-centos-env" when we use it for every environment.  Might be worth renaming it?  I don't think it's a PR-blocker, but figured I'd mention.

PR #701, opened 2023-08-09
Enhancements to backup directory functionality

jmcatamney:
This PR slightly "flattens" the directory hierarchy created by --backup-dir and allows omitting the --timestamp flag if only one backup exists in a given directory.

See individual commit messages for further details.

### DIFF
--- a/restore/restore.go
+++ b/restore/restore.go
@@ -29,7 +29,6 @@ func DoInit(cmd *cobra.Command) {
 	CleanupGroup.Add(1)
 	gplog.InitializeLogging("gprestore", "")
 	SetCmdFlags(cmd.Flags())
-	_ = cmd.MarkFlagRequired(options.TIMESTAMP)
 	utils.InitializeSignalHandler(DoCleanup, "restore process", &wasTerminated)
 }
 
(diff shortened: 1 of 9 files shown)


restore/restore.go
```diff
@@ -28,7 +29,6 @@ func DoInit(cmd *cobra.Command) {
 	CleanupGroup.Add(1)
 	gplog.InitializeLogging("gprestore", "")
 	SetCmdFlags(cmd.Flags())
-	_ = cmd.MarkFlagRequired(options.TIMESTAMP)
```
AJR-VMware, line 31: The "must have either timestamp or backup-dir flag" logic should probably be added to `ValidateFlagCombinations`.  We catch it downstream, but earlier in the process would be safer I think.

AJR-VMware (approved):
Aside from validation comment above, LGTM.

restore/restore.go
jmcatamney: The --backup-dir flag isn't actually required in this case, as it also works if there's a single backup in the standard backup location.

AJR-VMware: Ah, good point!  Never mind, then.

PR #706, opened 2023-08-19
Point CI at DP Concourse by default

khuddlefish:
Use dp concourse locations for google project and data domain VMs so we can fly the prod pipeline in the dp concourse

Also remove the rc-build-type-gcs since it doesn't seem to be doing anything right now.

Looking for feedback on vault variable naming. Like for moving the google and datadomain credentials to use the ones that are currently in `dev` but if this becomes our new default then eventually they would actually be considered the `prod` ones. So I can move stuff around in vault if needed. Just didn't want to change that until we took a look.


AJR-VMware:
I think the only things where we would really care about `dev` vs `prod` would be the GCS buckets where we store our artifacts, and those (if I recall correctly) should both be reachable from DP concourse.  So I guess my vote would be to stick with the current dev/prod split in `gen_pipeline.py` for those, and point everything except the GCS bucket variables statically to `dev`.

khuddlefish:
@AJR-VMware So does that mean it's good the way I have it currently in the PR? I don't think I changed the bucket variables away from prod. The gcs project and datadomain cred stuff is mostly pointing at dev aside from the ones where the credential is only available in prod.

AJR-VMware:
> @AJR-VMware So does that mean it's good the way I have it currently in the PR? I don't think I changed the bucket variables away from prod. The gcs project and datadomain cred stuff is mostly pointing at dev aside from the ones where the credential is only available in prod.

@khuddlefish Yeah, I think what you have here already looks good, assuming it's running green.  Nothing about the design I'd want to change, myself.

khuddlefish:
It is running green. If you like it can I get an approval?

jmcatamney (approved):

PR #708, opened 2023-08-23
Improve GetConstraits performance

jmcatamney:
The query was originally written for GPDB 5, which lacks the coninhcount column in pg_constraint and so requires a costly subquery on pg_inherit to retrieve inheritance information.  In 6 and later, we can simply use that column to omit inherited constraints instead.


AJR-VMware (approved):
Nice!

PR #711, opened 2023-09-01
Improve schema renaming for --redirect-schema

jmcatamney:
The existing logic for renaming schemas did not properly handle schema objects, objects whose schemas contained periods or quotation marks, or permissions blocks containing multiple GRANT or REVOKE statements, and this commit fixes that.

It also adds more test cases for editStatementsRedirectSchema and makes the tests easier to debug.


AJR-VMware (approved):
LGTM!

PR #715, opened 2023-09-06
Add table filters to index rename query

AJR-VMware:
This query has been having some performance issues on customer installs. Add table filters so at least in the filtered case it will be performant.

### DIFF
--- a/backup/queries_postdata.go
+++ b/backup/queries_postdata.go
@@ -250,38 +250,43 @@ func GetIndexes(connectionPool *dbconn.DBConn) []IndexDefinition {
 
 func GetRenameExchangedPartitionQuery(connection *dbconn.DBConn) string {
 	// In the case of exchanged partition tables, restoring index constraints with system-generated
-	// will cause a name collision in GPDB7+.  Rename those constraints to match their new owning
-	// tables.  In GPDB6 and below this renaming was done automatically by server code.
+	// will cause a name collision in GPDB7+. Rename those constraints to match their new owning
+	// tables. In GPDB6 and below this renaming was done automatically by server code.
 	cteClause := ""
 	if connectionPool.Version.Before("7") {
-		cteClause = `SELECT c.relname
-             FROM pg_class c
-             INNER JOIN pg_partition_rule p
-             ON c.oid = p.parchildrelid`
+		cteClause = fmt.Sprintf(`
+            SELECT c.relname
+            FROM pg_class c
+                INNER JOIN pg_partition_rule p
+                    ON c.oid = p.parchildrelid
+                INNER JOIN pg_namespace n ON c.relnamespace = n.oid
+            WHERE %s`, relationAndSchemaFilterClause())
 	} else {
-		cteClause = `SELECT DISTINCT cl.relname
-             FROM pg_class cl
-             WHERE
-                cl.relkind IN ('r', 'f')
-                AND cl.relispartition = true
-                AND cl.relhassubclass = false`
+		cteClause = fmt.Sprintf(`
+            SELECT DISTINCT c.relname
+            FROM pg_class c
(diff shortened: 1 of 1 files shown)


backup/queries_postdata.go
```diff
@@ -250,38 +250,43 @@ func GetIndexes(connectionPool *dbconn.DBConn) []IndexDefinition {
...
+                AND c.relkind IN ('r', 'f')
+                AND c.relispartition = true
+                AND c.relhassubclass = false`, relationAndSchemaFilterClause())
 	}
 	query := fmt.Sprintf(`
         WITH table_cte AS (%s)
         SELECT
             ic.relname AS origname,
-            rc.relname || SUBSTRING(ic.relname, LENGTH(ch.relname)+1, LENGTH(ch.relname)) AS newname
+            c.relname || SUBSTRING(ic.relname, LENGTH(ch.relname)+1, LENGTH(ch.relname)) AS newname
```
khuddlefish, line 278: I notice we changed the way of referring to pg_class from distinct letters to using the same letter for all references. Will that cause any conflicts in the sub queries?

backup/queries_postdata.go
```diff
@@ -250,38 +250,43 @@ func GetIndexes(connectionPool *dbconn.DBConn) []IndexDefinition {
...
         FROM
             pg_index i
-            JOIN pg_class ic ON i.indexrelid = ic.oid
-            JOIN pg_class rc
-                ON i.indrelid = rc.oid
-                AND rc.relname != SUBSTRING(ic.relname, 1, LENGTH(rc.relname))
-            JOIN pg_namespace n ON rc.relnamespace = n.oid
+            INNER JOIN pg_class ic ON i.indexrelid = ic.oid
```
khuddlefish, line 281: I'm reading that `JOIN` does an `INNER JOIN` by default so this should end up the same as when we just said `JOIN` before right? Are we just changing it so all `JOIN`s are consistent or any other behavior reason?

khuddlefish (commented):
I think this looks ok to me though I'm not 100% confident in my query reading skills

backup/queries_postdata.go
AJR-VMware: Yep, just consistency and readability.  The behavior is identical either way.

backup/queries_postdata.go
AJR-VMware: Our `relationAndSchemaFilterClause` utility function applies the table filters, and it has the hardcoded assumption that the pg_class we're using for tables is aliased to `c`.  Because this query uses two references to pg_class to achieve its logic, I had given them both slightly longer aliases to differentiate.  

This change puts the pg_class reference that joins on the `indrelid` field (the table the index is on) to match the `c` convention our table filtering utility expects, as that's the one we want to limit to only consider indexes on tables we're filtering to.

khuddlefish:
@AJR-VMware Do we have any tests for this query? And does it make sense to?

AJR-VMware:
> @AJR-VMware Do we have any tests for this query? And does it make sense to?

We definitely do.  This logic is finnicky enough that it would be dangerous not to, imo.
https://github.com/greenplum-db/gpbackup/blob/b061a47b673238439442340e66ca57d896edacd5/integration/postdata_create_test.go#L349
https://github.com/greenplum-db/gpbackup/blob/b061a47b673238439442340e66ca57d896edacd5/backup/queries_shared_test.go#L38
https://github.com/greenplum-db/gpbackup/blob/b061a47b673238439442340e66ca57d896edacd5/backup/queries_postdata_test.go#L39C3-L39C3

khuddlefish (approved):
I'm still not 100% confident in my query reading abilities but from what I can see it doesn't seem we have changed the functionality aside from adding the filtering

PR #716, opened 2023-09-07
Fix bug in redirect-schema processing

AJR-VMware:
Recent change had a bug where CREATE TABLE statements for partition leaves were falling through the ATTACH PARTITION logic and then not getting edited properly on GPDB7 restores.

### DIFF
--- a/restore/restore.go
+++ b/restore/restore.go
@@ -414,15 +414,14 @@ func editStatementsRedirectSchema(statements []toc.StatementWithType, redirectSc
 	// This expression matches an ATTACH PARTITION statement and captures both the parent and child schema names
 	attachRE := regexp.MustCompile(fmt.Sprintf(`(ALTER TABLE(?: ONLY)?) (%[1]s)(\..+ ATTACH PARTITION) (%[1]s)(\..+)`, schemaMatch))
 	for i := range statements {
-		statementObject := statements[i]
 		oldSchema := fmt.Sprintf("%s.", statements[i].Schema)
 		newSchema := fmt.Sprintf("%s.", redirectSchema)
-		statement := statementObject.Statement
+		statement := statements[i].Statement
 		// Schemas themselves are a special case, since they lack the trailing dot.
 		statement = strings.Replace(statement, fmt.Sprintf("CREATE SCHEMA %s", statements[i].Schema), fmt.Sprintf("CREATE SCHEMA %s", redirectSchema), 1)
 
 		statements[i].Schema = redirectSchema
-		if statements[i].ObjectType == "SCHEMA" {
+		if statements[i].ObjectType == toc.OBJ_SCHEMA {
 			statements[i].Name = redirectSchema
 		}
 		replaced := false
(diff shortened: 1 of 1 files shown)


restore/restore.go
```diff
@@ -301,8 +301,10 @@ func restorePredata(metadataFilename string) {
 		schemaStatements = GetRestoreMetadataStatementsFiltered("predata", metadataFilename, []string{toc.OBJ_SCHEMA}, []string{}, filters)
 	}
 	statements := GetRestoreMetadataStatementsFiltered("predata", metadataFilename, []string{}, []string{toc.OBJ_SCHEMA}, filters)
+	gplog.Debug("AJR -- statements before redirect: %+v\n", statements)
```
khuddlefish, line 304: Make sure to remove these before merging

restore/restore.go
```diff
@@ -414,15 +416,14 @@ func editStatementsRedirectSchema(statements []toc.StatementWithType, redirectSc
 	// This expression matches an ATTACH PARTITION statement and captures both the parent and child schema names
 	attachRE := regexp.MustCompile(fmt.Sprintf(`(ALTER TABLE(?: ONLY)?) (%[1]s)(\..+ ATTACH PARTITION) (%[1]s)(\..+)`, schemaMatch))
 	for i := range statements {
-		statementObject := statements[i]
 		oldSchema := fmt.Sprintf("%s.", statements[i].Schema)
 		newSchema := fmt.Sprintf("%s.", redirectSchema)
-		statement := statementObject.Statement
```
khuddlefish, line 420: Does this have any functionality difference or is it just removing an extra line?

restore/restore.go
```diff
@@ -434,7 +435,7 @@ func editStatementsRedirectSchema(statements []toc.StatementWithType, redirectSc
 		}
 
 		// ALTER TABLE schema.root ATTACH PARTITION schema.leaf needs two schema replacements
-		if connectionPool.Version.AtLeast("7") && statements[i].ObjectType == toc.OBJ_TABLE && statements[i].ReferenceObject != "" {
+		if connectionPool.Version.AtLeast("7") && statements[i].ObjectType == toc.OBJ_TABLE && statements[i].ReferenceObject != "" && strings.Contains(statement, "ATTACH") {
```
khuddlefish, line 438: Was this processing things that didn't have ATTACH in them or missing things that did? The previous logic seems like it should have picked up ATTACH statements as long as they meet the other conditionals

khuddlefish (commented):
I'm not sure I quite understand what the issue was or how this fixes it

restore/restore.go
AJR-VMware: Yeah, just a weird creation of a variable that we then only use once but not in every other case we call `statements[i]`.  Just editing it to be consistent.

restore/restore.go
AJR-VMware: The issue was that CREATE TABLE `schema.leaftable` was falling into this section, as it met all these conditions.  However, the `attachRE` didn't match those statements, so nothing was actually getting replaced.  Then, `replaced = true` was being set, so the more general replacement on line 442 wouldn't fire.  This change adds one additional condition so that the CREATE TABLE statements (which do not have ATTACH in them) will not fall into this gate, and will instead be correctly handled on line 443.

khuddlefish (approved):

PR #719, opened 2023-09-12
Update scale ci to fix version issues

AJR-VMware:
Move everything for this pipeline to RHEL8-compatible, to deal with GLIBC mismatches.

There's still a bug in the pipeline, but it's related to the changes to directory structure so this should go in to start pinning down some of the moving parts.

### DIFF
--- a/ci/scale/scale_pipeline.yml
+++ b/ci/scale/scale_pipeline.yml
@@ -39,10 +39,10 @@ resource_types:
 
 resources:
 ##### Docker Images #####
-- name: rocky8-gpdb7-image
+- name: rocky8-gpdb6-image
   type: registry-image
   source:
-    repository: gcr.io/data-gpdb-public-images/gpdb7-rocky8-test
+    repository: gcr.io/data-gpdb-public-images/gpdb6-rocky8-test
     tag: latest
 
 ##### Source Code ####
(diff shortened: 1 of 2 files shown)


ci/scale/scale_pipeline.yml
```diff
@@ -69,11 +69,10 @@ resources:
...
 #### Binaries ####
-- name: bin_gpdb_6x_stable_centos7
+- name: bin_gpdb_6x_rhel8
   type: gcs
-  icon: google
   source:
-    bucket: ((dp/dev/gcs-bucket))
+    bucket: ((dp/prod/gcs-ci-bucket))
     json_key: ((concourse-gcs-resources-service-account-key))
     regexp: server/published/gpdb6/server-rc-(.*)-rhel7_x86_64.tar.gz
```
khuddlefish, line 77: You're still pointing at a rhel7 artifact so this is not `rhel8` like the name says

khuddlefish (changes requested):
The artifact you're pointing at is built for the wrong platform still

ci/scale/scale_pipeline.yml
AJR-VMware: I had it flown right but must have not committed that change.  Good catch, thank you.

khuddlefish (approved):

PR #720, opened 2023-09-19
Convert index name exchange to go code

AJR-VMware:
The current query is much too slow on large customer installs, so convert to a faster, simpler SQL query and handle the name processing in application code.


khuddlefish (approved):
I like this version much better. The other query feels way too complicated.

PR #721, opened 2023-09-19
WIP -- Refactor exchange index query

AJR-VMware:
The original version of this query saw very poor performance on some very large installs. This version is far more performant, at the cost of some significant code complexity.

Some WIP/TODO is left here, as we will want the team to compare this to the original approach and the application-code approach in a separate PR.  This will be edited out before merging.


AJR-VMware:
Closing this because the "do the logic in go code" approach was unanimously preferred.

PR #722, opened 2023-09-21
Update slack-alert url to work for any Concourse deployment

khuddlefish:
We had many things in this path hardcoded that should have been variables. Now it will generate the correct url in more scenarios


AJR-VMware (approved):
Love this change.  Thanks!

PR #723, opened 2023-09-22
Require bckup-dir if timestamp is not provided

AJR-VMware:
The recent change to allow finding a backup timestamp if only one is present should only work with backup-dir functionality.  Add a check that at least one of those flags is provided.


jmcatamney (approved):
In the commit message, `bckup-dir` -> `backup-dir`.

Otherwise, LGTM.

PR #729, opened 2023-10-04
Rhel9 support

khuddlefish:
Also remove zstd installation for tests since it is already installed


AJR-VMware (approved):
I don't see any issues, looks like everything is in place to me.

PR #730, opened 2023-10-05
Improve gpbackup_manager test coverage

jmcatamney:
As we have observed some issues with gpbackup_manager that were not caught by the single-node tests in the pipeline, this commit modifies the gpbackup_manager test job to have a multi-node setup.

Additionally, the recent change to gpbackup's directory hierarchy necessitates testing gpbackup_manager with both the old and new formats, so an additional task is added to run the test suite against a backup version from before the format change.

### DIFF
--- a/ci/templates/gpbackup-tpl.yml
+++ b/ci/templates/gpbackup-tpl.yml
@@ -1240,8 +1240,14 @@ jobs:
 {% else %}
       passed: [build_binaries_rhel6, build_binaries_rhel7, build_binaries_rhel8]
 {% endif %}
+    - get: gpdb_binary
+      resource: bin_gpdb_6x_rhel8
+{% if "gpbackup-release" != pipeline_name %}
+      trigger: true
+{% endif %}
     - get: gpbackup
       passed: [build_gppkgs]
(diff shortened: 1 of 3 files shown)


ci/templates/gpbackup-tpl.yml
```diff
@@ -1108,25 +1108,64 @@ jobs:
...
+      resource: bin_gpdb_6x_centos7
+{% if "gpbackup-release" != pipeline_name %}
+      trigger: true
 {% endif %}
     - get: gpbackup
       passed: [build_gppkgs]
+    - get: ccp_src
     - get: bin_gpdb
-      resource: bin_gpdb_6x_rhel8
+      resource: bin_gpdb_6x_centos7
```
AJR-VMware, line 1127: Is there a reason we rolled this back to centos7 instead of testing on 8?

jmcatamney: Because I was using the backwards_compatibility job as a template and that one is still using 7.  I can go ahead and update this back to 8.

AJR-VMware (approved):
LGTM!

PR #733, opened 2023-10-26
Add support for section-based backup and restore

jmcatamney:
To give users more granular control over which parts of a database are backed up or restored, this commit adds a new --sections flag that allows users to choose which backup "sections" (globals, predata, data, postdata, and/or statistics, the same sections currently specified in the ToC files) will be included in a specific backup or restore.

This flag covers the same functionality as --data-only, --metadata-only, --with-globals, --with-stats, and --without-globals, so those flags have been deprecated.  They are retained for backwards compatibility, but may not be used in conjunction with --sections.


history/history.go
```diff
@@ -72,6 +75,16 @@ func ReadConfigFile(filename string) *BackupConfig {
...
+	// We don't want to output DeprecatedMetadata.  The YAML "-" annotation doesn't cover this use case,
+	// because we still have to be able to *read* it when restoring older backups; using a pointer so we
+	// can set it to nil and have the marshalar ignore it causes Marshal to silently fail; and the standard
+	// methods of overriding MarshalYAML or implementing an IsZero function for DeprecatedMetadata doesn't
+	// work, for whatever reason; so we just marshal config as-is and chop off the output of those fields
+	// in the DeprecatedMetadata struct, as they'll always be at the end of the BackupConfig.
+	// TODO: Figure out a less hacky way to do this
+	configContents = []byte(strings.Split(string(configContents), "dataonly")[0])
+
 	gplog.FatalOnError(err)
```
AJR-VMware, line 88: Should the error go before the split (right after the Marshal call), to avoid panics on that array access?

restore/global_variables.go
```diff
@@ -46,6 +46,9 @@ var (
 	 * or the signal handler.
 	 */
 	CleanupGroup *sync.WaitGroup
+
+	// sections to restore
+	BackupSections options.Sections
```
AJR-VMware, line 51: Should this be `RestoreSections`?

restore/restore.go
```diff
@@ -106,15 +110,15 @@ func DoSetup() {
...
+	if BackupSections.Globals || BackupSections.Predata || BackupSections.Postdata {
 		gplog.Verbose("Metadata will be restored from %s", metadataFilename)
 	}
 	unquotedRestoreDatabase := utils.UnquoteIdent(backupConfig.DatabaseName)
 	if MustGetFlagString(options.REDIRECT_DB) != "" {
 		unquotedRestoreDatabase = MustGetFlagString(options.REDIRECT_DB)
 	}
-	ValidateDatabaseExistence(unquotedRestoreDatabase, MustGetFlagBool(options.CREATE_DB), backupConfig.IncludeTableFiltered || backupConfig.DataOnly)
-	if MustGetFlagBool(options.WITH_GLOBALS) {
+	ValidateDatabaseExistence(unquotedRestoreDatabase, MustGetFlagBool(options.CREATE_DB), backupConfig.IncludeTableFiltered && !backupConfig.Globals)
```
AJR-VMware, line 120: I don't follow how the logic to generate the `isFiltered` boolean is equivalent before/after, here.  I might be wrong on this one, but would like to chat through it if we can.

history/history.go
jmcatamney: Yes it should, good catch.

restore/global_variables.go
jmcatamney: Indeed it should.

restore/restore.go
jmcatamney: Just a typo on my part, good catch.

jmcatamney:
> Has this already passed a full prod pipeline?

It's passed a dev pipeline [here](https://dp.ci.gpdb.pivotal.io/teams/main/pipelines/sections), prod pipeline to be run shortly.

AJR-VMware (approved):

PR #736, opened 2023-11-14
Resolve CI Issues

AJR-VMware:
* We are dropping support from RHEL6-based distributions for future
releases, so remove all references to it from our CI.
    * This is due, in part, to the GCC and LIBC versions being incompatible with go 1.21.0+, as well as RHEL6 being long-since EOL.

* Recent changes to gp_toolkit on GPDB7+ break restores if the gp_toolkit
schema was not already created.  Further we currently hard-fail if an
extension is meant to be in a schema that was not prior created.
* To address the gp_toolkit issue and provide some additional stability,
explicitly add a CREATE IF NOT EXISTS statement for extension schemas in GP7 and up.


khuddlefish (approved):
LGTM. It will be so nice to get rid of this janky centos6 stuff

PR #737, opened 2023-11-14
Fix some bugs in sections code

AJR-VMware:
In initial implementation of the sections feature, there were a couple bugs to be fixed.

1) Simple typos in the CHECK constraints applied to the new fields in
   gpbackup_history.db.  Those are just fixed.

2) Instead of updating the new fields to 0 across the board, update them
   to correct values.  This fixes a broken interaction with
   gpbackup_manager, as well as just keeping more information available
   and correct.  Given that we have the information available to do so,
   there's no good reason not to.


khuddlefish:
I'm not totally sure how to read this change. Maybe we can talk it through tomorrow?

khuddlefish (approved):
LGTM

PR #739, opened 2023-11-17
Restore original backup-dir structure as default

AJR-VMware:
The recent changes to backup-dir structure had some unexpected consequences. To mitigate these, we are restoring the original behavior to be the default, and gating the new "single directory of all files" structure behind a new "--single-backup-dir" flag.


AJR-VMware:
I'm still doing some manual testing of all this, but it looks good so far and CI is green so I thought it would be best to get a PR open.

AJR-VMware:
> I'm still doing some manual testing of all this, but it looks good so far and CI is green so I thought it would be best to get a PR open.

Manual testing passes for me.  Full steam ahead, pending PR feedback.

restore/remote.go
```diff
@@ -52,6 +52,10 @@ func VerifyBackupFileCountOnSegments() {
 		// outside the destination array will be "[]", which the find command can handle safely in this context.
 		contentsList := fmt.Sprintf("(%s)", strings.Join(contentMap[contentID], "|"))
 		cmdString := fmt.Sprintf(`find %s -type f -regextype posix-extended -regex ".*gpbackup_%s_%s.*" | wc -l`, globalFPInfo.GetDirForContent(contentID), contentsList, globalFPInfo.Timestamp)
+
+		// NOTE: this is the original regex from default-dir structure. It is kept here commented,
+		// for use in possible future perf investigations if needed
+		// return fmt.Sprintf("find %s -type f | wc -l", globalFPInfo.GetDirForContent(contentID))
```
khuddlefish, line 58: What do we expect to use this for?

restore/restore.go
```diff
@@ -84,12 +84,13 @@ func DoSetup() {
 	err = opts.QuoteExcludeRelations(connectionPool)
 	gplog.FatalOnError(err)
 
-	segPrefix, err = filepath.ParseSegPrefix(MustGetFlagString(options.BACKUP_DIR), backupTimestamp)
+	segPrefix, singleBackupDir, err := filepath.ParseSegPrefix(MustGetFlagString(options.BACKUP_DIR), backupTimestamp)
 	gplog.FatalOnError(err)
 	globalFPInfo = filepath.NewFilePathInfo(globalCluster, MustGetFlagString(options.BACKUP_DIR), backupTimestamp, segPrefix)
+	globalFPInfo.SingleBackupDir = singleBackupDir
```
khuddlefish, line 90: It's kind of weird to me how we didn't include SingleBackupDir in the NewFilePathInfo function. Was there a reason for that?

ci/gpbackup-generated.yml
```diff
@@ -881,15 +881,17 @@ jobs:
   - task: setup-cluster-env
     file: gpbackup/ci/tasks/setup-cluster-env.yml
     image: rocky8-gpdb6-image
-  - task: run-tests-gpbackup-1.29.0
```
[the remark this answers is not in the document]
AJR-VMware: The way we're running these, the environment variables are set on the cluster, and never cleared.  Having the env vars set for the old version was keeping them set for the new version, and so we were running all tests in this job against gpbackup 1.29.0.  It seems to have been that way since this test was reworked to be multi-node.  Switching the order lets us run the latest-version tests correctly, then set the env var on the "old version" test, and gets us back to the intended behavior.

end_to_end/end_to_end_suite_test.go
```diff
@@ -1502,11 +1536,12 @@ var _ = Describe("backup and restore end to end tests", func() {
...
 		It("runs gpbackup and gprestore with statistics section", func() {
-			// gpbackup before version 1.18.0 does not dump pg_class statistics correctly
-			skipIfOldBackupVersionBefore("1.18.0")
+			if useOldBackupVersion {
+				Skip("This test is not needed for old backup versions")
+			}
 
 			timestamp := gpbackup(gpbackupPath, backupHelperPath,
-				"--backup-dir", backupDir, sectionBackupWithStats)
+				"--backup-dir", backupDir, sectionBackupWithStats, "--single-backup-dir")
```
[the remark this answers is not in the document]
AJR-VMware: Later parts of the test check specific file locations, and those are only present in the expected location if we use the flag.  I figured it would be good to have some of our test coverage still using the alternative structure instead of ripping it all out, so I kept some.

restore/remote.go
AJR-VMware: If the new structure ends up being truly unworkably slow, I wanted to keep the old "fast" file check around to for perf testing.  The find commands are different enough I suspect there may be some amount of improvement possible there, but there may not.

restore/restore.go
AJR-VMware: We use NewFilePathInfo to get filepaths in some places where we don't know whether we're in "singleBackupDir" mode or not.  The distinction is made downstream and the differentiated parts of the struct are populated by other functions.

It boils down to we'd have to arbitrarily pass in a dummy true/false value in a few places, without knowing if it's correct or not at runtime.  I felt that having such hardcoded, possibly incorrect, values passed into the function would be more a landmine for future us than just leaving it out, and updating the struct when we actually have the information.  I did go back and forth on this a bit, though.

end_to_end/end_to_end_suite_test.go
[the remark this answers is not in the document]
AJR-VMware: For sure, just added!

restore/remote.go
khuddlefish: Just give it a nice clear commit message so we can search for it properly

restore/restore.go
khuddlefish: Shouldn't we be able to know if we're in singleBackupDir format pretty early in the process? For backup side we should get as flag and on restore side it should be an early step in finding the backup files, right? I feel like we should be able to figure it out about as easily as we do segPrefix

restore/remote.go
AJR-VMware: Yep, will do.

backup/validate.go
```diff
@@ -187,7 +187,14 @@ func validateFlagValues() {
 }
 
 func validateFromTimestamp(fromTimestamp string) {
-	fromTimestampFPInfo := filepath.NewFilePathInfo(globalCluster, globalFPInfo.UserSpecifiedBackupDir, fromTimestamp, "")
+	var fromTimestampFPInfo filepath.FilePathInfo
+	if MustGetFlagBool(options.SINGLE_BACKUP_DIR) {
+		fromTimestampFPInfo = filepath.NewFilePathInfo(globalCluster, globalFPInfo.UserSpecifiedBackupDir, fromTimestamp, "", true)
+	} else {
+		fromTimestampFPInfo = filepath.NewFilePathInfo(globalCluster, globalFPInfo.UserSpecifiedBackupDir,
+			fromTimestamp, globalFPInfo.UserSpecifiedSegPrefix, false)
```
khuddlefish, line 195: could we use `globalFPInfo.SingleBackupDir` here?

AJR-VMware: Ah, yeah that would look cleaner. Will do.

backup/backup.go
```diff
@@ -116,7 +120,14 @@ func DoBackup() {
 	var targetBackupFPInfo filepath.FilePathInfo
 	if MustGetFlagBool(options.INCREMENTAL) {
 		targetBackupTimestamp = GetTargetBackupTimestamp()
-		targetBackupFPInfo = filepath.NewFilePathInfo(globalCluster, globalFPInfo.UserSpecifiedBackupDir, targetBackupTimestamp, "")
+
+		if MustGetFlagBool(options.SINGLE_BACKUP_DIR) {
```
khuddlefish, line 124: Do we still need this conditional? I'm not sure I get the importance of setting segPrefix to "". Is globalFPInfo.UserSpecifiedSegPrefix not already the right value?

restore/restore.go
AJR-VMware: Per conversation offline, this ended up being more workable than I thought, and is done.

backup/backup.go
AJR-VMware: Ah, good catch.  It is, and it removes this branch in two places.  Fixing now.

bmdoil:
> The recent changes to backup-dir structure had some unexpected consequences

For the uninitiated, what are the unexpected consequences?

filepath/filepath.go
```diff
@@ -208,36 +217,51 @@ func (backupFPInfo *FilePathInfo) GetHelperLogPath() string {
...
 	_, err := operating.System.Stat(fmt.Sprintf("%s/backups/%s/%s", backupDir, dateString, timestamp))
 	if err == nil {
-		// We're using the current directory format, there's no prefix to parse
-		return "", nil
+		// We're using the single-backup-dir directory format, there's no prefix to parse
+		return "", true, nil
 	}
 	if err != nil && !os.IsNotExist(err) {
-		return "", fmt.Errorf("Failure while trying to locate backup directory in %s. Error: %s", backupDir, err.Error())
+		return "", false, fmt.Errorf("Failure while trying to locate backup directory in %s. Error: %s", backupDir, err.Error())
```
bmdoil, line 247: nit: The diff here is a little funky because the added `GetSegPrefix` function is interspersed with the changes made to `ParseSegPrefix`. Any way we can clean that up?

restore/restore.go
```diff
@@ -42,6 +42,7 @@ func DoValidation(cmd *cobra.Command) {
 	gplog.FatalOnError(err)
 	err = utils.ValidateFullPath(MustGetFlagString(options.PLUGIN_CONFIG))
 	gplog.FatalOnError(err)
+
```
bmdoil, line 45: nit: Unrelated whitespace

bmdoil (approved):
LGTM just a couple nits

backup/validate.go
```diff
@@ -187,7 +190,10 @@ func validateFlagValues() {
 }
 
 func validateFromTimestamp(fromTimestamp string) {
-	fromTimestampFPInfo := filepath.NewFilePathInfo(globalCluster, globalFPInfo.UserSpecifiedBackupDir, fromTimestamp, "")
+	var fromTimestampFPInfo filepath.FilePathInfo
+	fromTimestampFPInfo = filepath.NewFilePathInfo(globalCluster, globalFPInfo.UserSpecifiedBackupDir,
+		fromTimestamp, globalFPInfo.UserSpecifiedSegPrefix, globalFPInfo.SingleBackupDir)
+
```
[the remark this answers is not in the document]
AJR-VMware: Nope, leftover from a removed `if` statement.  I'll change it.

PR #740, opened 2023-11-18
Clean pipeline trigger conditions

khuddlefish:
I noticed the gpdb7 job in gpbackup-release kept getting triggered when it wasn't supposed to. Make the trigger conditions a bit more unfiorm and ensure that jobs in release pipeline don't get triggered unless there is a new gpbackup build


AJR-VMware (approved):
LGTM

PR #743, opened 2023-12-07
Add --no-privileges flag to gpbackup

jmcatamney:
This flag causes gpbackup to omit statements related to object ownership and privileges from the metadata file, in order to (among other things) allow a predata restore to be done without needing all of the original roles to exist in the target database or needing to perform a restore of globals beforehand.

This commit does not add the same flag to gprestore to allow ignoring ownership and privilege statements for objects during a restore that does include them, but it does refactor the ToC logic to assign those statements their own types (as they are currently treated as tables) to enable that functionality to be more easily added to gprestore in a later commit.

pipeline: https://dp.ci.gpdb.pivotal.io/teams/main/pipelines/no-priv-on-1.30.x

### DIFF
--- a/backup/backup.go
+++ b/backup/backup.go
@@ -230,7 +230,9 @@ func backupGlobals(metadataFile *utils.FileWithByteCount) {
 	backupResourceQueues(metadataFile)
 	backupResourceGroups(metadataFile)
 	backupRoles(metadataFile)
-	backupRoleGrants(metadataFile)
+	if !MustGetFlagBool(options.NO_PRIVILEGES) {
+		backupRoleGrants(metadataFile)
+	}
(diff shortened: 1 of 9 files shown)


kyeap-vmware:
Unfortunately, this breaks filtered restore of partition tables on the restore side due to child ALTER statements running. Before this change they would not run since they are tied to the child partition creation. If the child partition ALTER become separate from the child partition CREATE, we get in a state where having the child ALTER reference the child CREATE in the TOC means filtered restore on the root breaks. If the child ALTER reference the root CREATE, the filter on the child statement breaks. Since we would like to keep feature parity, we'd like to give this a bit more thought on how to move forward with this.

PR #745, opened 2023-12-11
The gprestore helper skips distributed tables if there are replicated ones.

RekGRpth:
The gprestore helper skips distributed tables if there are replicated ones.

That leads to freeze of the restore process.
This occurs due to an incorrect condition when restoring to a smaller cluster.
I corrected this condition and also extended the test to restore replicated
tables on a smaller cluster so that it checks for the identified error.

The problem can be reproduced either through a test without changing the code,
or by working with a plugin, or by manually moving files, as is done in the test.


khuddlefish:
Hi @RekGRpth we reworked the logic in this part of the code with some recent changes. I think we likely fixed the issue you were seeing as a part of that rework.

PR #747, opened 2023-12-21
Support backup and restore database containing an extension with creating user mappinger

airfan1994:
gp_backup backup user mappings in pg_user_mappings catalog, and gp_restore restore user mappings in target database using create user mapping SQL. However, If user create an extension, in which code a user mapping is created. in such condition, target user mapping entry is inserted into pg_user_mappings catalog in source database.

when doing restore, the extension is recreated in target database. When creating the extension, the corresponding user mapping is created because of the redoing extension code. However, gp_restore will also restore the user mapping in pg_user_mapping catalog. So an error "user mapping xxx already exists." will be raised.

I open a MR and add IF EXISTS to CREATE USER MAPPING SQL in backup/predata_functions.go. Please have a look at this.

### DIFF
--- a/backup/predata_functions.go
+++ b/backup/predata_functions.go
@@ -484,7 +484,7 @@ func PrintCreateServerStatement(metadataFile *utils.FileWithByteCount, objToc *t
 
 func PrintCreateUserMappingStatement(metadataFile *utils.FileWithByteCount, objToc *toc.TOC, mapping UserMapping) {
 	start := metadataFile.ByteCount
-	metadataFile.MustPrintf("\n\nCREATE USER MAPPING FOR %s\n\tSERVER %s", mapping.User, mapping.Server)
+	metadataFile.MustPrintf("\n\nCREATE USER MAPPING IF NOT EXISTS FOR %s\n\tSERVER %s", mapping.User, mapping.Server)
 	if mapping.Options != "" {
 		metadataFile.MustPrintf("\n\tOPTIONS (%s)", mapping.Options)
 	}


khuddlefish:
Hi @airfan1994 thanks for your contribution. I assume you developed this against Greenplum 7? We also need to support Greenplum 5 and Greenplum 6. I think this object is mostly relevant for Greenplum 6 and Greenplum 7. Unfortunately the `IF NOT EXISTS` syntax for this object was not introduced until the postgres version included in Greenplum 7 so it fails if you try to run it on Greenplum 6.

```
postgres=# create user mapping if not exists for usr1
postgres-# server foreignserver;
ERROR:  syntax error at or near "if"
LINE 1: create user mapping if not exists for usr1
                            ^
```

Please revise your PR so we can generate correct syntax for both Greenplum 6 and Greenplum 7.

khuddlefish:
I also am not sure this is really the correct fix for this issue. I think with extensions objects there is some entry in pg_depend or some catalog table that is supposed to tell us an object is owned by an extension and so not to create it separately and let the extension handle it. That must not be working correctly here. The real fix is probably to make sure his object is correctly marked as being owned by an extension so we won't try to back it up separately.

airfan1994:
> I also am not sure this is really the correct fix for this issue. I think with extensions objects there is some entry in pg_depend or some catalog table that is supposed to tell us an object is owned by an extension and so not to create it separately and let the extension handle it. That must not be working correctly here. The real fix is probably to make sure this object is correctly marked as being owned by an extension so we won't try to back it up separately.
> 
> See here represented by deptype `e` https://www.postgresql.org/docs/9.1/catalog-pg-depend.html

Good idea! I agree with you, and I will try to solve this problem in that way.

PR #748, opened 2023-12-27
Move constraints back to postdata for data load performance

kyeap-vmware:
Originally, constraints were dumped in predata. This caused a restore
issue with tables that used constraints that used the `NOT VALID`
clause. This is because there were scenarios where there was data
grandfathered into tables that violate the constraint that has not yet
been resolved by users, but allowed because of the `NOT VALID` clause.
Because of this issue, and the assumption that nothing can depend on
constraints, constraints were moved from predata to postdata to allow
tables with data that violate `NOT VALID` constraints to restore.
https://github.com/greenplum-db/gpbackup/commit/90b3b54b83fe01f8107a329f2bd10652abd67bd6

It was then later discovered that views can depend on constraints. All
constraints, expect ones with `NOT VALID` clause, were moved back into
predata.
https://github.com/greenplum-db/gpbackup/commit/5895076a2678b00e50455e889d7df93d003fd3e8

While this fixes the dependency issues, this is a significant
performance regression because it is much more performant to restore
constraints after all data has been restored. This is because
constraints use internal indexes. If constraints are restored during
predata, data load would slow down considerably because every row insert
would trigger constraint index update.

In order resolve both the performance regression and original issue, we
can move constraint restore back into postdata and resolve the circular
dependency of a predata view depending on a postdata constraint using a
dummy view in predata. Now if a view that depends on a constraint is
detected, a dummy view that satisfies dependency conditions for other
relations gets dumped in predata. The dummy view and will eventually get
replaced by the dump of the full view definition in postdata after
constraints are dumped.

This circular dependency of predata views depending on postdata
constraints is not applicable to 5X. View dependency on constraints
exist because of the feature to allow incomplete GROUP BY lists that was
introduced in GPDB6 (postgres 9.1). 
https://github.com/greenplum-db/gpdb/commit/e49ae8d3bc588294d07ce1a1272b31718cfca5ef

gpbackup pipeline:
https://dp.ci.gpdb.pivotal.io/teams/main/pipelines/dump-constraints-postdata

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,10 +60,17 @@ To run only unit tests, use
 ```bash
 make unit
 ```
-To run only integration tests (requires a running GPDB instance), use
+To run only integration tests
 ```bash
 make integration
 ```
+Integration test requirements
+ - Running GPDB instance
+ - GPDB's gpcloud extension
+```bash
+make -C gpcontrib/gpcloud/ install
+```
+ - GPDB configured with `--with-perl`
 
 To run end to end tests (requires a running GPDB instance), use
 ```bash
--- a/backup/backup.go
+++ b/backup/backup.go
@@ -276,11 +276,15 @@ func backupPredata(metadataFile *utils.FileWithByteCount, tables []Table, tableO
 
 	retrieveViews(&objects)
 	sequences := retrieveAndBackupSequences(metadataFile, relationMetadata)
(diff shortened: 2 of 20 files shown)


khuddlefish (approved):
LGTM

PR #749, opened 2023-12-28
Use type TABLE when printing GRANT statements for views

khuddlefish:
The TABLE keyword in GRANT/REVOKE statements is optional and we omitted printing it for VIEWS, FOREIGN TABLES, and MATERIALIZED VIEWS likely because it looks a bit strange to say TABLE for a view. However, when we do `--redirect-schema` we have a regex that matches on GRANT statements and replaces the schema. Since all other objects have an OBJECT type, the regex wasn't matching properly for the above object types and ended up creating syntax errors when it tried to replace.

Now we print TABLE as the object type for TABLES, FOREIGN TABLES, VIEWS, and MATERIALIZED VIEWS so the format will match for all objects. Note that pg_dump also prints TABLE in the statements for these object types.


kyeap-vmware (approved):
Solution looks looks good to me. It's better for consistency to explicitly list out an object type, even if it may look a bit strange for objects that aren't explicitly tables, but just look like them.

PR #751, opened 2024-01-09
New sections pr

khuddlefish:
We did most of the work of implementing a new feature called `--sections` and merged it in main. This would allow users to specify which sections of the backup or restore they want to run from the list of globals, stats, predata, data, postdata. We planned to release this as part of gpbackup 1.31.0. However, the main engineers on the project are no longer with the company and we are lacking some confidence in bringing in this change now. There is some ambiguity about which objects would be included in each section (constraints and views for example). We want to pull this out for now and think more about if we want to do this and how we want to do it. The most relevant ask is the ability for uses to restore postdata separately from predata.

This also includes a commit to remove rhel6 support from the pipeline. The rhel6 gcc version was incompatible with go 1.21 and the sections implementation used some feature that required go 1.21.


bmdoil:
Closing this PR in favor of a different implementation. The requirement is to provide an option to restore excluding indexes and index-backed constraints and the scope of the sections feature is too large.

PR #753, opened 2024-01-26
Fix example plugin tests to use relative paths

kyeap-vmware:
When gpbackup repo is not in $GOPATH and integration tests are run,
tests that use the example plugin will hang because the test cannot
proceed without the example plugin reading data from test pipes. This
happens because the example plugin and it's config file locations were
hardcoded to be in $GOPATH. This is a legacy issue from when go projects
were required to be in $GOPATH. This requirement was changed to be
optional in go1.11. This is an update to look for example plugin using a
relative path. The test plugin config file is also now setup before
tests run because the executablepath is also updated to be a relative
path.

pipeline: https://dp.ci.gpdb.pivotal.io/teams/main/pipelines/fix-tests-local

### DIFF
--- a/end_to_end/end_to_end_suite_test.go
+++ b/end_to_end/end_to_end_suite_test.go
@@ -61,7 +61,10 @@ var (
 	backupHelperPath        string
 	restoreHelperPath       string
 	gprestorePath           string
-	pluginConfigPath        string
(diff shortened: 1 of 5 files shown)


bmdoil (approved):
LGTM

PR #754, opened 2024-01-30
Allow restore of tables distributed by enum

bmdoil:
Tables or materialized views with enum distribution keys have not been restorable. Issue https://github.com/greenplum-db/gpdb/issues/14198 provides further context on the reasoning. The issue has been resolved for pg_dump/pg_restore by cherry-picking https://github.com/postgres/postgres/commit/7e7c5b683985c85fb990c4d49ab960cbc83434b4. For gpbackup, the issue occurs because `COPY ON SEGMENT` loads data directly from the segment host into the segment database itself. Since enums are unstable when re-created due to their oids changing, data that was backed up on one segment may be hashed to a separate one on restore. Since the COPY is unable to redistribute tuples to the correct segments, the restore fails.

This PR introduces a workaround for restoring tables/matviews distributed by enum. The `GetDistributionPolicies` query now also returns a bool `distbyenum`. A new field has been added to the coordinator TOC corresponding to `distbyenum`. On restore, if `distbyenum=t` , `gp_enable_segment_copy_checking` is disabled, the table is copied in, reorganized, and then the GUC is reset.

Due to the introduction of a new struct and function definition changes, there are quite a few test changes.

dev pipeline: https://dp.ci.gpdb.pivotal.io/teams/main/pipelines/distby_enum


end_to_end/resources/gpdb6_objects.sql
```diff
@@ -65,9 +65,16 @@ CREATE TYPE public.colors AS ENUM (
...
+
+CREATE TABLE public.legacy_enum (
+    color public.colors
+) DISTRIBUTED BY (color cdbhash_enum_ops);
+
+CREATE TABLE public.multi_key_enum (
+    a public.colors,
+    b public.colors
+) DISTRIBUTED BY (a, b)
+
```
kyeap-vmware, line 77: We'll also need to put data into these tables for a chance to trigger wrong distribution key error when loading a row distributed by an enum. Even then, with only 3 segments we may still get lucky and restore may still pass. Any ideas on how wrong distribution on restore can be forced?

putting the test under end_to_end's "edge case tests" could make it easier to enable for 7x when the time comes.

kyeap-vmware (changes requested):

kyeap-vmware (approved):
Pretty cool workaround with the reuse of the redistribution code from -resize-cluster.

PR #755, opened 2024-02-01
Remerge 1.30.x branch back into main 

kyeap-vmware:
Karen did a lot of work to keep the 1.30.x and the main branch in sync as much as possible with made this easier.

I did a revert to remove the "sections" feature. Then cherry-picked commits from 1.30.x that did not get merged into main.

Recent feature releases like the 7x granular progress bars and the
parallel metdata restore have made big changes that introduced a number
of regressions to running gpbackup in the background and parallelism in
general. In order to bring gpbackup into a more stable state and bring
the main branch into a releasable state we are reverting features that
make big code changes across the repo so we can vet the changes more
carefully before merging them in.

We are also reverting the "sections" feature so it can be vetted in
depth.

sister gp-backup-manager PR: https://github.com/pivotal/gp-backup-manager/pull/64

pipeline: https://dp.ci.gpdb.pivotal.io/teams/main/pipelines/remove-sections


khuddlefish:
I think instead of reverting this PR we should just hard reset main to match 1.30.x. I think the code will end up cleaner working from a branch that never had this in it

PR #766, opened 2024-02-26
CI: test s3 plugin using minio

kyeap-vmware:
Test the s3 plugin against s3-compliant object storage. Minio is a
popular lightweight s3 compliant storage solution. While we have a
makefile rule to test gpbackup/gprestore and the s3 plugin against a
minio instance, this does not run in the CI. Setup a local minio
instance in the CI and test basic functionality of the s3 plugin.

pipeline: https://dp.ci.gpdb.pivotal.io/teams/main/pipelines/fix-no-bucket


bmdoil (approved):
LGTM!

PR #768, opened 2024-03-14
Properly dump GRANT statements for aggregate functions

bmdoil:
Aggregates don't have a GRANT ON AGGREGATE syntax, instead GRANT ON FUNCTION is used. However, aggregates are still distinct object types and createPrivilegeStrings did not consider this. Properly handle aggregates when creating privilege strings.
dev pipeline: https://dp.ci.gpdb.pivotal.io/teams/main/pipelines/grant_on_agg

### DIFF
--- a/backup/predata_acl.go
+++ b/backup/predata_acl.go
@@ -381,6 +381,9 @@ func createPrivilegeStrings(acl ACL, objectType string) (string, string) {
 	case toc.OBJ_TYPE:
 		hasAllPrivileges = acl.Usage
 		hasAllPrivilegesWithGrant = acl.UsageWithGrant
+	case toc.OBJ_AGGREGATE:
+		hasAllPrivileges = acl.Execute
+		hasAllPrivilegesWithGrant = acl.ExecuteWithGrant
 	}
 	if hasAllPrivileges {
 		privStr = "ALL"
--- a/backup/predata_acl_test.go
+++ b/backup/predata_acl_test.go
@@ -149,7 +149,7 @@ REVOKE ALL ON FOREIGN SERVER foreignserver FROM PUBLIC;
 REVOKE ALL ON FOREIGN SERVER foreignserver FROM testrole;
 GRANT ALL ON FOREIGN SERVER foreignserver TO testrole;`)
 		})
-		It("prints FUNCTION for REVOKE and AGGREGATE for ALTER for an aggregate function", func() {
+		It("prints FUNCTION for GRANT/REVOKE and AGGREGATE for ALTER for an aggregate function", func() {
 			aggregate := backup.Aggregate{Schema: "public", Name: "testagg"}
 			aggregatePrivileges := testutils.DefaultACLForType("testrole", toc.OBJ_AGGREGATE)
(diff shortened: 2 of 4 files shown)


kyeap-vmware (changes requested):
The test still passes without the fix. The test is making sure that the aggregate ACL is being queried from the database, but we need to be testing if it gets printed.
I found an existing unit test that only tests REVOKE. I think expanding it to also test GRANT would be good enough.
```
		It("prints FUNCTION for REVOKE and AGGREGATE for ALTER for an aggregate function", func() {
			aggregate := backup.Aggregate{Schema: "public", Name: "testagg"}
			aggregatePrivileges := testutils.DefaultACLForType("testrole", toc.OBJ_AGGREGATE)
			aggregateMetadata := backup.ObjectMetadata{Privileges: []backup.ACL{aggregatePrivileges}, Owner: "testrole"}
			backup.PrintObjectMetadata(backupfile, tocfile, aggregateMetadata, aggregate, "", []uint32{0, 0})
			testhelper.ExpectRegexp(buffer, `

ALTER AGGREGATE public.testagg(*) OWNER TO testrole;


REVOKE ALL ON FUNCTION public.testagg(*) FROM PUBLIC;
REVOKE ALL ON FUNCTION public.testagg(*) FROM testrole;`)
		})
```

bmdoil:
> The test still passes without the fix. The test is making sure that the aggregate ACL is being queried from the database, but we need to be testing if it gets printed. I found an existing unit test that only tests REVOKE. I think expanding it to also test GRANT would be good enough.
> 
> ```
> 		It("prints FUNCTION for REVOKE and AGGREGATE for ALTER for an aggregate function", func() {
> 			aggregate := backup.Aggregate{Schema: "public", Name: "testagg"}
> 			aggregatePrivileges := testutils.DefaultACLForType("testrole", toc.OBJ_AGGREGATE)
> 			aggregateMetadata := backup.ObjectMetadata{Privileges: []backup.ACL{aggregatePrivileges}, Owner: "testrole"}
> 			backup.PrintObjectMetadata(backupfile, tocfile, aggregateMetadata, aggregate, "", []uint32{0, 0})
> 			testhelper.ExpectRegexp(buffer, `
> 
> ALTER AGGREGATE public.testagg(*) OWNER TO testrole;
> 
> 
> REVOKE ALL ON FUNCTION public.testagg(*) FROM PUBLIC;
> REVOKE ALL ON FUNCTION public.testagg(*) FROM testrole;`)
> 		})
> ```

Hmm, `createPrivilegeStrings` is called by `GetPrivilegesStatements` and `PrintDefaultPrivilegesStatements` so the test didn't exercise it. It's probably worth keeping around for coverage to avoid future regressions.

I updated the unit test to also test for GRANT. I think we can do better by testing privileges in `predata_functions_test.go`.  Currently we create the test metadata without privileges.
```
func DefaultMetadata(objType string, hasPrivileges bool, hasOwner bool, hasComment bool, hasSecurityLabel bool) backup.ObjectMetadata


aggMetadata = testutils.DefaultMetadata(toc.OBJ_AGGREGATE, false, true, true, true)
```
I updated the `prints an aggregate with owner` tests to check for ACLs too.

kyeap-vmware (approved):
LGTM

PR #770, opened 2024-03-19
Restore helper fixes

bmdoil:
This PR aims to improve cleanup of gpbackup_helper specific objects. It revamps the functions called by DoCleanup tasked
with ensuring we don't leave any files, pipes, processes, or COPY sessions lingering. The functions will attempt cleanup
repeatedly over a timeout period via check->cleanup loop. These functions are primarily exercised by the signal handler
e2e tests. This is intended to be a stop-gap improvement while the helper agent architecture is redesigned.

Also, it updates the permissions for created pipes to `0700`, so they are no longer accessible to other users or groups (although privileged users can still mess with them).

While an improvement, this approach doesn't handle situations where the backup/restore process exits such that DoCleanup isn't executed, or timing issues with the ssh commands dispatched to the segments such that we receive stale information.

dev pipeline: https://dp.ci.gpdb.pivotal.io/teams/main/pipelines/restore_helper_fixes


utils/agent_remote.go
```diff
@@ -184,38 +186,178 @@ HEREDOC
...
+		}
+	}
+	return totalPipes
+}
+
+// Removes all gpbackup_helper pipes from the segment data directories.
+// It's expected that gpbackup_helper cleans up the pipes on exit, but that is not guaranteed,
+// so this function is used to clean up any leftover pipes.
+func RemoveHelperPipes(c *cluster.Cluster, fpInfo filepath.FilePathInfo) {
+	errMsg := fmt.Sprintf("Unable to remove gpbackup_helper pipe(s). See %s for a complete list of segments with errors and remove manually.",
```
khuddlefish, line 270: Can this be moved down to right above where it is used? Just reads a little weird to me like the error is actually happening at the beginning rather than just creating the string

end_to_end/end_to_end_suite_test.go
```diff
@@ -295,15 +293,28 @@ func assertArtifactsCleaned(conn *dbconn.DBConn, timestamp string) {
 		oidFile := fpInfo.GetSegmentHelperFilePath(contentID, "oid")
 		scriptFile := fpInfo.GetSegmentHelperFilePath(contentID, "script")
 		pipeFile := fpInfo.GetSegmentPipeFilePath(contentID)
+		replicatedOidFile := fpInfo.GetSegmentHelperFilePath(contentID, "replicated_oid")
```
khuddlefish, line 296: With the change Kevin and I did we won't have this file anymore

end_to_end/end_to_end_suite_test.go
```diff
@@ -683,7 +686,8 @@ var _ = Describe("backup and restore end to end tests", func() {
 			skipIfOldBackupVersionBefore("1.8.2")
 			createGlobalObjects(backupConn)
 
-			timestamp := gpbackup(gpbackupPath, backupHelperPath)
+			output := gpbackup(gpbackupPath, backupHelperPath)
```
khuddlefish, line 689: Is this change of returning "output" for backup mostly to unify with the return format with use for restore? Do we have places where we're checking the backup output?

end_to_end/signal_handler_test.go
```diff
@@ -123,12 +118,11 @@ var _ = Describe("Signal handler tests", FlakeAttempts(5), func() {
 			backupConn.MustExec("ROLLBACK")
 
 			stdout := string(output)
-			timestamp, err := getBackupTimestamp(stdout)
-			Expect(err).ToNot(HaveOccurred())
-			assertArtifactsCleaned(backupConn, timestamp)
+			timestamp := getBackupTimestamp(stdout)
+			assertArtifactsCleaned(timestamp)
 			Expect(stdout).To(ContainSubstring("Received an interrupt signal, aborting backup process"))
+			Expect(stdout).To(ContainSubstring("Interrupt received while acquiring ACCESS SHARE locks on tables"))
```
khuddlefish, line 124: Do we know it will happen during that phase of the process?

khuddlefish (approved):
LGTM. I'm a little scared on the merge conflicts between this and the thing @kyeap-vmware and I are doing, but we'll work it out

end_to_end/end_to_end_suite_test.go
bmdoil: >Is this change of returning "output" for backup mostly to unify with the return format with use for restore?

Yes, not checking the backup output in any tests currently but now we can if we want to.

end_to_end/signal_handler_test.go
bmdoil: Yes it will, the test uses `BEGIN; LOCK TABLE schema2.foo2 IN ACCESS EXCLUSIVE MODE` to block and doesn't proceed until gpbackup is blocked, so we'll always hit that log line.

PR #775, opened 2024-04-06
Fix getBackupTimestamp in e2e test

bmdoil:


khuddlefish (approved):

khuddlefish (changes requested):
This would need accompanying changes everywhere we call it to include the returned error

PR #776, opened 2024-04-10
Use different find command syntax for MacOS vs Linux

khuddlefish:
The -regextype option is not present in find on mac so this caused every gprestore to fail locally on my laptop. -E is the equivalent

Authored-by: Karen Huddleston <khuddleston@vmware.com>


kyeap-vmware (approved):
LGTM

PR #777, opened 2024-04-12
Stabilize signal handler tests

khuddlefish:
Sometimes these tests will fail if gpbackup is killed before a timestamp is generated

- Allow getBackupTimestamp to handle a missing timestamp
- Check if timestamp is empty before using it for test assertions. Test should not fail if it was killed before timestamp as long as we still see the `Interrupt` log message
- Modify signal timing and comments to have a little more randomness. Will cancel between 0.5 and 2.0 seconds
- Reduce size of bigtable slightly. Will hopefully make setup a bit quicker and size was likely not helping the timing that much.


kyeap-vmware (approved):
LGTM

PR #778, opened 2024-04-17
Set log file verbosity to match shell verbosity

khuddlefish:
This will be INFO level by default and file verbosity can be changed along with shell verbosity by passing `--verbose` or `--debug` flags. Some logs moved up to `INFO` level since they are important for debugging and we want to see them for most gpbackup/gprestore runs without making the user re-run.

This also uses a new gplog option called Progress to go along with our use of progress bars. It will log to log file at INFO level but only stdout at VERBOSE level to avoid messing up progress bars.
gp-common-go-libs PR is here: https://github.com/greenplum-db/gp-common-go-libs/pull/87


kyeap-vmware:
For those running into 
```
CGO_ENABLED=1 go build -mod=readonly -tags 'gpbackup' -o /home/kyeap/go/bin/gpbackup --ldflags '-X github.com/greenplum-db/gpbackup/backup.version=1.30.4+dev.5.g0de3867'
# github.com/greenplum-db/gpbackup/backup
backup/data.go:96:9: undefined: gplog.Progress
backup/data.go:98:9: undefined: gplog.Progress
backup/data.go:117:8: undefined: gplog.Progress
make: *** [Makefile:82: build] Error 1
```
To use the correct common-go-lib
`go get -u github.com/greenplum-db/gp-common-go-libs@loglevel-spike`

kyeap-vmware (approved):
LGTM

Could we also capture the context on why we are making this change. I assumed that it was to reduce the amount of logging that gets written to our logs or is it just to correct the log levels?

bmdoil (approved):

khuddlefish:
Yeah it's mostly to reduce the amount of logging. I can add a bit more context to the commit message

PR #779, opened 2024-04-26
Add test for parallel index restore on AO table

bmdoil:
Postdata metadata restore is batched due to a deadlock issue when restoring indexes to AO tables in parallel. Previously, this has not been explicitly tested. A regression in 7X server vs. 5/6 results in deadlock even after the first index is created on each table. This test exercises that code path to catch any future regressions.


kyeap-vmware (approved):
LGTM!

Thanks for writing this test.

khuddlefish (approved):
LGTM. Let's wait to merge this until the server fix goes in so we don't get red tests

PR #780, opened 2024-04-26
Speed up cleanup commands and fix for macos

khuddlefish:
- Have cleanup commands run once immediately instead of waiting one second for the ticket to start
    - find command for locating backup_helper files did not work on MacOS. Make it work for both linux and mac
    - Remove separate command for cleaning pipes and just use command to clean files for all
    - Check again for was terminated at the end of DoTeardown to ensure cleanup finishes in the case of a race condition


khuddlefish:
Closing in favor of https://github.com/greenplum-db/gpbackup/pull/783. This one seems to have introduced some instability

PR #782, opened 2024-04-30
Remove extra cat

khuddlefish:
We use this in the case of no compression and for helpers, but it adds an extra
pipe when we don't need to.

Also fixed a bug on the gprestore side with replicated tables and larger-to-smaller resize restores


kyeap-vmware (approved):
LGTM

PR #783, opened 2024-05-01
Fix cleanup commands to run on mac

khuddlefish:
- Find command was not the right syntax for macos so have it adjust for linux
  or darwin
- Make sure to TrimSpace on result so it can be parsed properly
- Pipe cleanup function was actually calling the file cleanup function again so
  pipes may have been left behind

Also refactor code to combine helper file and pipe cleanup into a single function


khuddlefish:
Concourse pipeline: https://dp.ci.gpdb.pivotal.io/teams/main/pipelines/cleanup_improvements2

kyeap-vmware (approved):
LGTM!

PR #784, opened 2024-05-10
Start using channels and events to coordinate helper reading and writing

khuddlefish:
Previously we had been polling when trying to open named pipes for writing. The
poll would sometimes sleep for 1 minute at a time which introduced performance
issues especially when restoring with many tables.

Now instead of polling we use goroutines and channels to attempt to open the
pipe for writing in blocking mode and alerting the main process via channel
when it works. Also use fsnotify events to watch for skip file creation in the
case of COPY errors.

Also use --verbose flag for gprestore commands in end to end tests to get more logging


helper/restore_helper.go
```diff
@@ -229,58 +242,48 @@ func doRestoreAgent() error {
...
+				// ability to open pipes normally, so hard quit even if
+				// --on-error-continue is given
+				logError(fmt.Sprintf("Oid %d, Batch %d: Cannot open pipe %s for writing. Exiting with error: %s", tableOid, batchNum, currentPipe, err))
+				return err
+			case event := <-watcher.Events:
+				if event.Has(fsnotify.Create) && event.Name == fmt.Sprintf("%s_skip_%d", *pipeFile, tableOid) && *onErrorContinue {
+					logWarn(fmt.Sprintf("Oid %d, Batch %d: Skip file discovered, skipping this relation.", tableOid, batchNum))
+					err = nil
+					goto LoopEnd
+				}
```
kyeap-vmware, line 276: We get a go routine leak for every skip file. Deleting the pipe does not unblock OpenFile.

khuddlefish: Yeah I was concerned about that too. Working on a fix

khuddlefish: @kyeap-vmware PR updated with the fix. Check it out

PR #785, opened 2024-05-10
Check for "Cleanup complete" in tests

khuddlefish:
If the process hasn't gotten far enough before termination, gpbackup or gprestore might skip calling helper cleanup functions. Our tests may flake if they don't expect that. We should trust the cleanup function to do the cleanup it needs so checking for "Cleanup complete" is enough. We still check that the artifacts are actually cleaned, just not the log messages.


kyeap-vmware (approved):
I am fine with this change. We haven't figured out a good pattern for fault injection in golang yet.

PR #786, opened 2024-05-16
Replace backoff while waiting for pipe with short static sleep

khuddlefish:
We added this backoff wait when we were having test hangs caused by other errors. The backoff didn't end up helping and introduced a potentially serious performance regression when running restore with many tables. Replace the backoff with a static sleep of 50ms while we wait for COPY to connect to the pipe. This returns to similar behavior from older gpbackup versions. The 50ms sleep should also prevent the CPU from being too busy while we wait.

Also enable --verbose logging mode for gprestore end_to_end tests


bmdoil (approved):
LGTM. Do we want to include the updated scale test to exercise this change in this PR?

khuddlefish:
@bmdoil I have this PR for the scale test https://github.com/greenplum-db/gpbackup/pull/787

Though I did find that the existing scale tests seem to somewhat exercise it too since the tests with my fix ran almost 1 hour shorter than normal.

kyeap-vmware (approved):

PR #787, opened 2024-05-16
Modify scale test data to better exercise discovered slowdown

khuddlefish:
We found a slowdown with a customer caused by us polling on ENXIO errors while waiting for a previous COPY command to finish and connect with the next one. We might wait up to a minute. To see this we need a good number of tables and we need them to be large enough to trigger the backoff. Testing by creating a partition table with 100 partitions and loading around 14GB of data. We make it uneven across partitions so some are larger than others.

If we have this table we can remove the smaller 100 tables we had before and also reduce the number of tiny tables we had.


kyeap-vmware (approved):
LGTM.

PR #788, opened 2024-05-20
Move Progress log to gpbackup

khuddlefish:
We had previously added a Progress command to gplog that would let us log at INFO level to the log file and VERBOSE level to stdout. Then we found other places where controlling different levels to shell and log file would be useful to we added a new Custom function to be more general. Since gpbackup is the only project that uses this Progress command, we can move the function here and use the new Custom function to implement it.


bmdoil (approved):
LGTM

